PR #5, opened 2021-04-14
Initial ci framework pass. Triggers when  ci:run tag added to a pull request.

jwithers:
This is a starting point. Triggers when ci:run tag added to a pull request. The dockerfile being used isn't the one in ./ci, but a slightly modified version living in ./github/actions/docker_action. It still runs whatever is in test_all_new.sh. The entire dockerfile build and run strategy will be changing to use a checked in container which will make this a lot cleaner.


jwithers:
@googlebot I signed it!

advaitjain (approved):

PR #10, opened 2021-04-19
Adding ci variable checkin, running Dockerfile.micro from registry and having `ci-run` tag removed at end of run.

jwithers:
This is now running the full test suite from an image checked into dockerhub. It also commits a ci variables file that right now is just a placeholder until we have the real tests and directory set up. 

It also removes the 'ci-run' tag after the workflow run is started.


jwithers:
Bah, This was a typo. Closing and will redo from proper branch.

Issue #11, opened 2021-04-19
Update docker image used as part of github actions to be hosted on github packages.


advaitjain:
https://github.com/tensorflow/tflite-micro/pull/10/ makes use of the docker image jpwithers/tflite-micro-tests ([here](https://github.com/tensorflow/tflite-micro/pull/10/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R14)).

Creating the current issue to track the work needed to have the docker image be hosted on GitHub packages.

jwithers:
I have spent some time with the Github Container Registry (ghcr) and it is fairly straightforward to use from actions. The trick is getting the container in there.

The easy way to do this is to create a manually run action that uploads the image, this is fairly simple and uses the GITHUB_TOKEN secret to access the tensorflow organization's account on the ghcr. The only problem here dealing with forked repos, which won't be able to use the action without it failing because their GITHUB_TOKEN isn't going to have access since it isn't coming from the tensorflow organization. 

The only other way to do this would be to have someone with organizational access manually upload the container on any changes to the container. 

Note that this workflow action wouldn't run automatically at any point. You have to go into actions and manually trigger it. And the only downside to someone without access running it is the job would fail. 

I would do the manual workflow action but need your call on it @advaitjain.

jwithers:
I have this working in the latest PR #30 . Using PULL_REQUEST_TARGET didn't work as expected with the github token and the container repository using private images, using WORKFLOW_RUN does. As outlined in the PR comments, we aren't able to use the github token to access private images outside the organization, there is just no option to allow that kind of access from the container permissions side. Even if you are an admin of an org, a container put in your account's registry space has no option to allow repositories outside your account repos to access the container from automation using the github token. But you can get around this with a PAL token, which is what we are doing right now.

When/if ghcr functionality is activated in the tensorflow org this will stop being a thing and we can use an internal access only container in the tensorflow registry space and the github token which is better practice for a lot of reasons. 

[...]

PR #12, opened 2021-04-19
Run tests when ci:run label is added to a pull request.

jwithers:


.github/workflows/main.yml
```diff
@@ -3,18 +3,32 @@ on:
...
-      - name: Show Workspace Context
-        run: echo 'Workspace is ${{ github.workspace }}'
-      - name: Nasty Hack To Get Right Docker Build Context
-        run: cp ./.github/actions/docker_action/Dockerfile . && cp ./.github/actions/docker_action/action.yml .
-      - name: Run Docker Tests
-        uses: ./
-      - name: Show commit
-        run: git show -q
+      - name: test_all.sh running from Dockerfile.micro
+        uses: docker://jpwithers/tflite-micro-tests
```
advaitjain, line 14: I'm assuming that we will be moving this to a docker image hosted with GitHub packages in a follow-up change?

I have created https://github.com/tensorflow/tflite-micro/issues/11 to start having open issues for work that we are punting for later. It would be nice to add a TODO(#11) comment in here as well.

.github/workflows/main.yml
```diff
@@ -3,18 +3,32 @@ on:
...
+#       uncomment below to run only the five minute test bit.
+#        with:
+#          args: /bin/sh -c "/opt/tflm/tensorflow/lite/micro/tools/ci_build/test_all.sh"
```
advaitjain, line 17: is comment still valid? I would expect that directly running test_all.sh would be the same as using Dockerfile.micro.

.github/workflows/main.yml
```diff
@@ -3,18 +3,32 @@ on:
...
+      - name: add a file for testing
+        run: git show -q > .commitinfo
+      - name: Commiting CI Variables Files
+        uses: EndBug/add-and-commit@3276c159cee23546e7fef9c8f9997e923587b543
```
advaitjain, line 21: this is ok to use but let's start adding some notes on what 3rd party github actions we are using (in addition to the non-trivial things that the CI system is doing). I think adding a top-level continuous_integration.md and putting notes in there would be very useful. We can clean it up later but documenting as we go along would be really nice.

main.yml
```diff
@@ -0,0 +1,34 @@
+on:
```
advaitjain, line 1: is this file needed? looks like a copy of .github/workflows/main.yml

advaitjain (changes requested):

.github/workflows/main.yml
jwithers: It is valid. Those lines override the CMD in the container which removes the GITHUB_PRESUBMIT argument, thus cutting out the bazel stuff.

.github/workflows/main.yml
jwithers: Yeah, I am still sorting through that now and getting it set up on my fork.

main.yml
jwithers: Yeah, that's an error. Thanks.

.github/workflows/main.yml
advaitjain: ah, yes.

I'm going to be soon enabling many more tests from test_all.sh so its time will also increase.

If you need something quick for your work, I might suggest changing this to be  /opt/tflm/tensorflow/lite/micro/tools/ci_build/test_x86.sh (with a comment that says, uncomment to run only the x86 tests for faster development of the CI work)

advaitjain (dismissed):

advaitjain (approved):

PR #25, opened 2021-04-26
Allow CI workflow to be triggered in different ways. 

advaitjain:
* from pull request, if the appropriate label is applied
 * manual triggering via workflow_dispatch
 * part of a scheduled build

Prior to this change, manually triggering the CI workflow, or a scheduled run would be skipped because the `ci:run` label was not found.


advaitjain:
closing this PR in favor of https://github.com/tensorflow/tflite-micro/pull/26

PR #30, opened 2021-04-28
test files for container permissions issues

jwithers:
Documenting this set of tests in a shared doc now, will backport it here if needed.


jwithers:
Have changed the ci-test setup. Now it runs in two stages, the ci-test.yml workflow runs at lower privilege when coming from a pull request. A second workflow, ci-test-privileged.yml, runs at a higher privilege level. 
Unfortunately, there is no way using the GITHUB token to actually pull off giving the automation here access to a private container in another account no matter what you do (which is not how it appeared, but 8 hours later am quite sure about it). It works fine with a PAT token as a secret, tho. @advaitjain you will have to add a secret named CONTAINER_PAT to the repository secrets for this. I will send you the payload for that by another channel. Note this PAT thing is just a demo workaround until the container machinery is activated here, and then we can use a container inside the organization repo and the GITHUB token. 
This setup seems a little clunky and has a couple of annoying bits that require passing things between workflows, but after banging my head on it for a while can it is quite a bit smarter for security.

jwithers:
Okay, changed ci-test.yml to use pull_request_target trigger and point at the container that @advaitjain should be uploading. Removed ci-test-privileged.yml since we are going to dump the workflow_run trigger strategy.

advaitjain:
I have taken the content of this PR and merged it here: https://github.com/tensorflow/tflite-micro/blob/78c805c624f887de9a93a69b25124ca6dbe3a546/.github/workflows/ghcr_test.yml

advaitjain:
And tested that the workflow is working as expected with https://github.com/tensorflow/tflite-micro/pull/84

PR #104, opened 2021-05-27
Updated sync and use of shared TFL/TFLM code.

advaitjain:
* The TFLM Makefile globs all the shared TFL/TFLM code.
 * This allows us to move the explicit list of sources and headers to the sync script since that is where we determine what code needs to be sync'd from upstream TF.

With this change, we are ready to have the tflite_micro repository be the source of truth for all TFLM code and sync only the shared TFL/TFLM code from the tensorflow repo.

The last update of the TFLM code from upstream Tensorflow was with https://github.com/tensorflow/tflite-micro/pull/110. We will now move all the TFLM development to the current repository.

Bug: http://b/182914089


advaitjain:
I would like to merge this PR first thing on June 1, 2021 and then the transition of the TFLM code into the current repository will be complete.

njeffrie (approved):

Issue #108, opened 2021-06-01
Remove uint8 support in CMSIS-NN kernels


mansnils:
Uint8 support is being removed in TFLM. CMSIS-NN kernels still have some uint8 support and it should be removed.

PR #109, opened 2021-06-01
Revert "Revert: PR #48758: TFLM: Remove uint8 support for CMSIS-NN ke…

mansnils:
Some CMSIS-NN kernels are still having uint8 support.

    Remove uint8 support for remaining CMSIS-NN kernels.
    Remove uint8 support for equivalent reference kernels.
    Remove uint8 unit tests for equivalent kernels.
    Replace uint8 default model in network tester to a int8 model.

This is a fix for: https://github.com/tensorflow/tflite-micro/issues/108

This was previously merged and reverted in the old repo because of some unknown issue. There was also a revert of the reverted PR in the old repo, which this is based on. Here is that PR for more background: https://github.com/tensorflow/tensorflow/pull/49485


advaitjain:
network_tester_test is now failing for bluepill: https://github.com/tensorflow/tflite-micro/pull/109/checks?check_run_id=2721060110#step:3:5872

Would you be able to debug why that is the case?

mansnils:
@advaitjain I cannot see that network tester failed in the link you posted. The log window is just black and in the link below it looks like network_tester is passing. I am not able to reproduce it either.

I cannot reproduce the PPD crash directly in tensorflow. I can just reproduce it using another local project using FVP (if it is the same issue) that is using tensorflow and from there running the unit tests.

The fact that it happens in Github but not in my local environment is strange, perhaps something is overwritten and in my case it happens to overwrite something that does not matter. So since I am not able to reproduce it with Renode I will just increase the bluepill stack to see if it matters.

Is it possible to just retrigger the CI / Cortex-M (presubmit)?

mansnils:
I also wonder if you know approximately how long it takes for CI to return the results?

advaitjain:
All tests are now passing on CI with the larger stack size but the detection_postprocess_test is failing consistently for me locally:

```
make -f tensorflow/lite/micro/tools/make/Makefile TARGET=bluepill test_kernel_detection_postprocess_test -j8
```

Removing and re-applying the ci:run label will let you trigger the CI.

I'd like to figure out what is happening before we merge since we have already had one round of merge and revert because there is something subtle going on.

> @advaitjain I cannot see that network tester failed in the link you posted. The log window is just black and in the link below it looks like network_tester is passing. I am not able to reproduce it either.
> 

My bad. I meant to say the detection_postprocess_test is failing. network_tester is fine.

> I cannot reproduce the PPD crash directly in tensorflow. I can just reproduce it using another local project using FVP (if it is the same issue) that is using tensorflow and from there running the unit tests.
> 

Can you give more details on what you are seeing? I'm hoping that it is the same issue.

advaitjain:
> I also wonder if you know approximately how long it takes for CI to return the results?

CI will complete in around 10 mins. But you have to remove and reapply the label. If you want, you can also make a PR in your fork of tflite-micro and change the worflow file to remove everything but https://github.com/tensorflow/tflite-micro/blob/28322741351947fc94de75abb4f0be8448e2bdd4/.github/workflows/ci.yml#L57

mansnils:
Yes I believe there is some subtle issue as you say.
I tried to create an identical PR in my local fork but could not reproduce it: https://github.com/mansnils/tflite-micro/pull/2
About the issue I see when running PPD unit test from another project .. the crash seems to happen in flatbuffers (although that may just be a symptom). Changing something in flexbuffers.h can make it work.
For example this: https://github.com/mansnils/tflite-micro/pull/2/commits/94e6fef63aabd2740e00369c8f2e29b5b438159e
If that would work for you locally (or something similar) it may be the same issue..

advaitjain:
I just tried out mansnils@94e6fef and detection_postprocess_test is now passing for me locally with bluepill. I'd be ok with merging with that fix though its hard for me to understand why that fixes things. Any context from your debugging for why it is needed would be very helpful.

mansnils:
I don't understand the problem although I suspect flexbuffers. I first thought it was related to some breaking of the strict-aliasing rule in flexbuffers and maybe it still is but I am not sure.
I found some potential problem in the detection_postprocess_test.cc although I don't think it solves the underlying problem. However if that would make bluepill work it would be better to merge this, https://github.com/tensorflow/tflite-micro/pull/109/commits/6e39d12d7688143bd17acb2c5d422385baef9e0f, besides it should probably be merged anyway.

mansnils:
I was able to get the "indirect PPD unit test" that I mentioned working by removing -fno-strict-aliasing and increasing optimization level. -fstrict-aliasing is implicitly enabled at -O1 or higher.
@advaitjain this is perhaps similar what you have seen with bluepill, meaning it works with?
make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=bluepill OPTIMIZATION_LEVEL=-Os test
but not?
make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=bluepill test
On the other hand it seems -Wno-strict-aliasing is default for bluepill. So if you still see the issue locally you could try to remove -Wno-strict-aliasing?

advaitjain:
The fix to the PPD test with 6e39d12 was sufficient to get it to pass for me locally. I have reverted the stack size increase for bluepill and will merge this PR.

Yes, I was only seeing the bluepill failure with -O3. The test was passing with -Os. We have seen this previously as well (https://github.com/tensorflow/tensorflow/issues/48516#issuecomment-846303105)

The -Wno-strict-aliasing is only needed because of some (not great) code in stm32_bare_lib. I'm going to make a separate PR to fix that.

advaitjain (approved):

PR #111, opened 2021-06-01
Updated sync and use of shared TFL/TFLM code.

advaitjain:
* The TFLM Makefile globs all the shared TFL/TFLM code.
 * This allows us to move the explicit list of sources and headers to the sync script since that is where we determine what code needs to be sync'd from upstream TF.

With this change, we are ready to have the tflite_micro repository be the source of truth for all TFLM code and sync only the shared TFL/TFLM code from the tensorflow repo.

Bug: http://b/182914089

### DIFF
--- a/ci/sync_from_upstream_tf.sh
+++ b/ci/sync_from_upstream_tf.sh
@@ -21,7 +21,6 @@
 # of the TFLM code via this script.
 #
 
-set -x
 set -e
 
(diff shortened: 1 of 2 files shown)


advaitjain:
This was a test PR to investigate long queue times for github actions for https://github.com/tensorflow/tflite-micro/pull/104.

The issue appears to have resolved itself.

PR #118, opened 2021-06-03
RFC for memory allocation updates in TFLM.

bhanuprakashbv:
Signed-off-by: Vijay Pawar <vpawar@cadence.com>

### DIFF
--- a/tensorflow/lite/micro/docs/rfc/003_mem_alloc_updates.md
+++ b/tensorflow/lite/micro/docs/rfc/003_mem_alloc_updates.md
@@ -1,0 +1,52 @@
+# TensorFlow Lite for Microcontrollers Memory Allocations Update
+
+| Status        | Proposed                                                    |
+:-------------- |:----------------------------------------------------------- |
+| **RFC #3**    |                                                             |
+| **Author(s)** | Bhanu Prakash Bandaru Venkata (bhanup@cadence.com)          |
+|               | Mayur Jagtap (mayurj@cadence.com)                           |
+|               | Niranjan Yadla (nyadla@cadence.com)                         |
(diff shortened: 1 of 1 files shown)


freddan80:
Thanks for the RFC @bhanuprakashbv! In addition the @petewarden's [comments](https://github.com/tensorflow/tensorflow/pull/49134#discussion_r631509202), here's the link to the [RFC for offline planned activation buffers](https://docs.google.com/document/d/16aTSHL5wxsq99t6adVbBz1U3K8Y5tBDAvs16iroZDEU/edit) to add some more context to the flatbuffer metadata "offline plan". It is also described [here](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/memory_management.md#offline-planned-tensor-allocations).

There are currently two ways tensors can be allocated:
1. Offline, via flatbuffer metadata. Only for activation buffers, since scratch buffers aren't known until runtime.
2. Online (runtime), via the e.g. greedy memory planner. Both for activation and scratch buffers.

A hybrid of of the two are allowed, but the offline plan takes prio. 

This RFC adds a third way of allocating tensors, from the application/client. Hence, a prio mechanism needs to be determined between the three - or some other way of avoiding conflicts between the different mechanism.

As a side note, we had a stab at it in [this RFC](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/rfc/001_preallocated_tensors.md#proposed-implementation), but the solution leads to increased complexity and code size in the micro allocator. If the complexity could reside at the application/client level, as you propose, it would be a more attractive solution.

Cheers!

petewarden:
Closing this, since I think we've discussed alternative approaches.

PR #136, opened 2021-06-07
micro:BATCH_MATMUL PR3-5

ddavis-2015:
micro: port operator BATCH_MATMUL kernel from lite with test

Complete implementation of TFLM operator BATCH_MATMUL and associated TFLM test code.

PR step 5 of the work to port operator BATCH_MATMUL as tracked in Issue #153 (BUG=153)


tensorflow/lite/micro/kernels/batch_matmul_test_util.h
```diff
@@ -0,0 +1,345 @@
...
+constexpr int kOutputTensor = 2;
+constexpr float kTolerance = 1e-5;
+
+enum TestInputIndex {
+  kInputIndex0,
+  kInputIndex1,
+};
+constexpr size_t kMaxInputs = TestInputIndex::kInputIndex1 + 1;
+
+struct TestTensorData {
```
petewarden, line 45: I'm guessing that the code here is a prototype of a more general set of testing utilities, and will be moved into a less op-specific header in the future if we like this direction? Maybe a comment to that effect, and a link to docs or bugs (if any) about this work would be useful?

ddavis-2015: This is generic code.  Should the TFLM team decide to use this style of test code, I would expect the file to be renamed.

advaitjain:
Closing in favor of other higher priority PRs.

dansitu:
Hi @advaitjain, I hope you are well! We're interested in doing some work with self-attention and this PR would unblock it. Has there been any consideration given to reprioritizing this since self-attention has become so important?

mansnils:
Hi @advaitjain, BATCH_MATMUL is getting more common in transformer networks. How about reopening this PR?

PR #141, opened 2021-06-08
Change special case logic for music-break model

mike-dooley:
Change model from 2 second input 4 second input.


advaitjain (approved):

mike-dooley:
abandoning this pull request, replaced by new one.

PR #147, opened 2021-06-08
Kernel variants for keyword benchmark operators

njeffrie:
Add reference kernel variants for operators used in the keyword benchmark: FULLY_CONNECTED, QUANTIZE, SOFTMAX, SVDF.

keyword benchmark before:
```
   text    data     bss     dec
  57436   37663   46612  141711

Total: Insns=2450168 Pcycles=1761566
```

keyword benchmark after:
```
   text    data     bss     dec     hex filename
  49692   37487   46612  133791

Total: Insns=2438595 Pcycles=1751834
```


advaitjain:
Closing this PR since the decision was that since kernel variants is an optimization, they should go in the optimized kernel directories.

PR #150, opened 2021-06-09
remove ci runner labels from PRs

jwithers:
Does what it says on the tin. Removes ci:run and ci:test tags from PRs to make the interface prettier.

Uses the github-script action, which is a github maintained action and is MIT licensed. Since the maintainers and most contributors on it are github staff the threat level doesn't seem high to me, but we can also pull it into our tree if we want.

Also, the continue-on-error bit seems hinkey but is just there to ignore the error from whichever tag isn't present to be removed and stops us from having to write even more boilerplate here than we already are.


.github/workflows/ci.yml
```diff
@@ -104,3 +104,32 @@ jobs:
...
+    if: |
+      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ci:run')) ||
+      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ci:test'))
+
+    name: remove CI runner labels
+    steps:
+      - name: remove tags
+        uses: actions/github-script@a3e7071a34d7e1f219a8a4de9a5e0a34d1ee1293
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
```
advaitjain, line 120: will GITHUB_TOKEN work if the PR is coming from a fork?

use the tflm bot repo token instead?
https://github.com/tensorflow/tflite-micro/blob/0afb62f9d1074df11c92f259a2eb1fc936009b96/.github/workflows/sync.yml#L51

.github/workflows/ci.yml
```diff
@@ -104,3 +104,32 @@ jobs:
         run: |
           tensorflow/lite/micro/tools/ci_build/test_makefile.sh
           tensorflow/lite/micro/tools/ci_build/test_x86.sh
+
+  label_cleanup:
```
advaitjain, line 108: we are also using the label trigger for the xtensa workflow (and potentially others as well).

https://github.com/tensorflow/tflite-micro/blob/0afb62f9d1074df11c92f259a2eb1fc936009b96/.github/workflows/xtensa.yml

Should we move this label removing step to its own workflow file? Can we get into a race condition where the label removal happens before all the workflow that rely on ci:run are triggered?

advaitjain (changes requested):
some clarifying questions.

.github/workflows/ci.yml
jwithers: Bah, yeah, the trigger is going to have to be changed to pull_request_target. Pretty sure that is necessary for either one to work, since we haven't been able to get at secrets from fork PRs without that.

.github/workflows/ci.yml
jwithers: I did a little bit of testing and it appears that the triggering happens and queues before execution. Let me try to figure out how to do a few more tests to fully confirm that.

advaitjain: ok, that would make sense. I'm happy to try it out and if we notice that things are not working as intended we can re-evaluate.

I would say let's still move this to its own workflow.yml

.github/workflows/ci.yml
advaitjain: ah yes, you're right.

.github/workflows/ci.yml
jwithers: Yeah, so found this [pull request breakdown](https://frontside.com/blog/2020-05-26-github-actions-pull_request/) and the workflows are triggered from an event message json payload (which you can see in the variable we are using in these if conditionals). We are changing the issue itself, not the event json message, so we are good.

.github/workflows/remove-labels.yml
```diff
@@ -0,0 +1,34 @@
+name: Remove Labels
+
+on:
+  pull_request_branch:
```
advaitjain, line 4: what is pull_request_branch?

.github/workflows/remove-labels.yml
```diff
@@ -0,0 +1,34 @@
...
+    types: [labeled]
+
+jobs:
+  label_cleanup:
+    runs-on: ubuntu-latest
+
+    if: |
+      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ci:run')) ||
+      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ci:test'))
```
advaitjain, line 13: remove `gihub.event_name == pull_request` since we only trigger this workflow on a pull_request?

.github/workflows/remove-labels.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    name: remove CI runner labels
+    steps:
+      - name: remove tags
+        uses: actions/github-script@a3e7071a34d7e1f219a8a4de9a5e0a34d1ee1293
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
```
advaitjain, line 20: will GITHUB_TOKEN work for a PR coming from a fork?

.github/workflows/remove-labels.yml
```diff
@@ -0,0 +1,34 @@
...
+              repo: context.repo.repo,
+              name: 'ci:run'
+            })
+            github.issues.removeLabel({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              name: 'ci:test'
+            })
+        continue-on-error: true
```
advaitjain, line 34: Is it possible to add a comment in the yml for why `continue-on-error: true` is needed?

advaitjain (changes requested):

advaitjain (approved):

PR #151, opened 2021-06-09
Fixing style check container and calls to have it actually work corre…

jwithers:
Created a new container and modified the script call to work correctly against the current pull instead of running against a static copy. 

Tried to move this out of docker entirely but it turns out that getting the llvm buster toolchain running on the github action virtual environment descends into dependency hell pretty quickly. Dropped back to an ubuntu-18.04 runner, which is an ubuntu version based off Buster, but no joy. Given enough time I could probably sort it, but there are a crapload of packages that appear to already be installed and configured in the python:3.9.0 image, including libtinfo6 which after multiple rounds of having to install prereqs seems to be unavailable to the ubuntu-18.04 distro and that's when I decided way too much time had gone into pulling this out of a container. 

If we stick with this, it should be moved into the tflm-bot's ghcr space and the call changed to match.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,10 +69,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Check
-        # TODO(#11): Update the docker image to be hosted via github packages.
-        uses: docker://jpwithers/tflite-micro-tests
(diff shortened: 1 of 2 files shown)


advaitjain:
Thanks for looking into this. Agreed that we should keep the docker container for now and that we should have it be hosted on the tflm-bot account. And I'll close this PR and make a new one following the pattern from this PR.

PR #160, opened 2021-06-10
Add Hexagon optimized kernels.

advaitjain:
* Hexagon optimized kernels copied from https://source.codeaurora.org/quic/embedded_ai/tensorflow at 2d052806c211144875c89315a4fc6f1393064cf6
 * Changed the include paths and directory structure a bit.
 * Modified Makefile to allow optimized kernels to be in a separate directory
 * Path to the Hexagon lib is now specified on the command line.

Verified that the optimized kernels are properly built and linked with:

```
make -f tensorflow/lite/micro/tools/make/Makefile TARGET=hexagon OPTIMIZED_KERNEL_DIR=hexagon OPTIMIZED_KERNEL_DIR_PREFIX=third_party HEXAGON_TFLM_LIB=~/Qualcomm/tflm_google/hexagon_tflm_core.a  -j8 run_keyword_benchmark
```

Gives:
```
KeywordRunNIerations(1) took 52608 ticks (52 ms)
```

Whereas reference kernels with:
```
make -f tensorflow/lite/micro/tools/make/Makefile TARGET=hexagon  -j8 run_keyword_benchmark
```

Gives:
```
KeywordRunNIerations(1) took 110248 ticks (110 ms)
```

BUG=http://b/190754463


tensorflow/lite/micro/tools/make/makefile_helper_functions.py
```diff
@@ -0,0 +1,53 @@
...
+  for fpath in base_file_list:
+    fname = os.path.basename(fpath)
+    if fname in specialize_files:
+      specialized_list.append(os.path.join(directory, fname))
+    else:
+      specialized_list.append(fpath)
+  return specialized_list
+
+
+if __name__ == "__main__":
```
njeffrie, line 35: Nitpick: should this be called substitute_specialized_implementation.py instead of makefile_helper_functions.py since we cannot add other functions due to having a single entrypoint via __main__?

njeffrie (changes requested):
One naming nit. Curious why we are removing the PW download in the Makefile. Perhaps this belongs in a separate change.

tensorflow/lite/micro/tools/ci_build/test_code_style.sh
```diff
@@ -31,24 +31,15 @@ make -f tensorflow/lite/micro/tools/make/Makefile third_party_downloads
 # scripts fail with an error code.
 set +e
 
-# The pigweed scripts only work from a git repository and the Tensorflow CI
```
njeffrie, line 34: Should this be part of the same change?

advaitjain: created https://github.com/tensorflow/tflite-micro/pull/173. I'll fix up the current PR once #173 is merged.

tensorflow/lite/micro/tools/make/makefile_helper_functions.py
advaitjain: Yeah, makes sense. Renamed.

njeffrie (approved):

PR #168, opened 2021-06-11
Separate core, kernel and third party objects.

njeffrie:
Add a second optimization level in the Makefile to enable different
levels between kernels and the rest of the TFLM code. This results in a
smaller binary with minimal performance impact compared with using a
single optimization level.

Allows the use of implicit patterns to compile all sources, choosing
different flags for core, kernel and third party sources.

The following measurements are taken using the hexagon toolchain +
hexagon-size and hexagon-sim.

For the keyword benchmark using -O2:
   text    data     bss     dec
  58140   37639   46612  142391

  Cycles: 1700364

For the keyword benchmark using -O2 for kernels and -Oz for framework:
   text    data     bss     dec
  52796   37623   46612  137031

  Cycles: 1759664

### DIFF
--- a/tensorflow/lite/micro/tools/ci_build/test_bluepill.sh
+++ b/tensorflow/lite/micro/tools/ci_build/test_bluepill.sh
@@ -41,13 +41,6 @@ readable_run make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=${TARG
 readable_run make -f tensorflow/lite/micro/tools/make/Makefile clean
 readable_run make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=${TARGET} test
 
-# We have had examples where tests pass with -Os but fail without it so we run
-# the unit tests with and without -Os. See
-# https://github.com/tensorflow/tensorflow/issues/48516 for one such issue.
-
-readable_run make -f tensorflow/lite/micro/tools/make/Makefile clean
-readable_run make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=${TARGET} test
-
 # We use Renode differently when running the full test suite (make test) vs an
 # individual test. So, we test only of the kernels individually as well to have
 # both of the Renode variations be part of the CI.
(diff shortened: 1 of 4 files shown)


tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -113,7 +113,15 @@ ifneq ($(CO_PROCESSOR),)
...
 endif
 
-OPTIMIZATION_LEVEL := -O3
+CORE_OPTIMIZATION_LEVEL := -Os
+KERNEL_OPTIMIZATION_LEVEL := -O2
+
+# Warn if deprecated optimization level is set.
+OPTIMIZATION_LEVEL :=
+ifneq ($(OPTIMIZATION_LEVEL),)
+echo "OPTIMIZATION_LEVEL is no longer used."
```
advaitjain, line 122: Let's make this an error. Warnings will get lost in the output

advaitjain (changes requested):
One nit. And that will also force your to update the test_ scripts.

tensorflow/lite/micro/tools/make/Makefile
njeffrie: Done.

advaitjain (approved):

tensorflow/lite/micro/tools/ci_build/test_bluepill.sh
```diff
@@ -34,12 +34,12 @@ readable_run make -f tensorflow/lite/micro/tools/make/Makefile TARGET=${TARGET}
 
 # check that the release build is ok.
 readable_run make -f tensorflow/lite/micro/tools/make/Makefile clean
-readable_run make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=${TARGET} OPTIMIZATION_LEVEL=-O3 BUILD_TYPE=release build
+readable_run make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=${TARGET} BUILD_TYPE=release build
 
 # Next, build w/o release so that we can run the tests and get additional
 # debugging info on failures.
 readable_run make -f tensorflow/lite/micro/tools/make/Makefile clean
-readable_run make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=${TARGET} OPTIMIZATION_LEVEL=-Os test
```
advaitjain, line 42: This line is now doing the same as then next one. Let's remove one of them. I'm ok switching bluepill to only use the default optimization levels.

njeffrie: Done.

advaitjain (dismissed):

advaitjain (dismissed):

advaitjain (approved):

advaitjain (approved):

PR #169, opened 2021-06-12
Add int8 conv kernel variant for xtensa

njeffrie:
Person detection benchmark binary size using xtensa optimized kernels:
```
   text	   data	    bss	    dec
 138288	 337488	 141752	 617528
```
Reference kernels without kernel variants:
```
   text	   data	    bss	    dec
 101824	 336240	 141752	 579816
```
Reference kernels with kernel variants:
```
   text	   data	    bss	    dec
 100040	 336160	 141752	 577952
```

### DIFF
--- a/tensorflow/lite/micro/kernels/conv.h
+++ b/tensorflow/lite/micro/kernels/conv.h
@@ -72,6 +72,8 @@ TfLiteStatus CalculateOpDataConv(TfLiteContext* context, TfLiteNode* node,
 
 TfLiteStatus ConvPrepare(TfLiteContext* context, TfLiteNode* node);
 
+TfLiteRegistration Register_CONV_2D();
+
 #if defined(XTENSA)
 // Returns a TfLiteRegistration struct for kernel variant that only supports
 // int8 inputs and outputs.
--- a/tensorflow/lite/micro/kernels/conv.h
+++ b/tensorflow/lite/micro/kernels/conv.h
@@ -72,6 +72,9 @@ TfLiteStatus CalculateOpDataConv(TfLiteContext* context, TfLiteNode* node,
 
 TfLiteStatus ConvPrepare(TfLiteContext* context, TfLiteNode* node);
 
+// This is the most generic TfLiteRegistration. The actual supported types may
+// still be target dependent. The only requirement is that every implementation
+// (reference or optimized) must define this function.
 TfLiteRegistration Register_CONV_2D();
 
 #if defined(XTENSA)
--- a/tensorflow/lite/micro/kernels/conv.h
+++ b/tensorflow/lite/micro/kernels/conv.h
@@ -78,7 +78,7 @@ TfLiteRegistration Register_CONV_2D();
 // Returns a TfLiteRegistration struct for kernel variant that only supports
 // int8 inputs and outputs.
 TfLiteRegistration Register_CONV_2D_INT8REF();
-#else 
+#else
 inline TfLiteRegistration Register_CONV_2D_INT8REF() {
   return Register_CONV_2D();
 }
--- a/tensorflow/lite/micro/kernels/conv.h
+++ b/tensorflow/lite/micro/kernels/conv.h
@@ -75,9 +75,9 @@ TfLiteStatus ConvPrepare(TfLiteContext* context, TfLiteNode* node);
 #if defined(XTENSA)
 // Returns a TfLiteRegistration struct for kernel variant that only supports
 // int8 inputs and outputs.
-TfLiteRegistration Register_CONV_2D_INT8();
+TfLiteRegistration Register_CONV_2D_INT8REF();
 #else 
-inline TfLiteRegistration Register_CONV_2D_INT8() {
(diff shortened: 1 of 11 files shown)


tensorflow/lite/micro/kernels/xtensa/conv_int8.cc
```diff
@@ -0,0 +1,140 @@
...
+namespace {
+
+void* Init(TfLiteContext* context, const char* buffer, size_t length) {
+  TFLITE_DCHECK(context->AllocatePersistentBuffer != nullptr);
+  return context->AllocatePersistentBuffer(context, sizeof(OpDataConv));
+}
+
+}  // namespace.
+
+TfLiteStatus ConvReferencePrepareInt8(TfLiteContext* context, TfLiteNode* node) {
```
advaitjain, line 38: Can we call the function from conv_common.cc instead of duplicating the code?

https://github.com/tensorflow/tflite-micro/blob/a0db1a6d01b9a20bbce1ffa0b07c9101bcc441b2/tensorflow/lite/micro/kernels/conv_common.cc#L126-L130

tensorflow/lite/micro/kernels/xtensa/conv_int8.cc
```diff
@@ -0,0 +1,140 @@
...
+      tflite::micro::GetTensorShape(filter),
+      tflite::micro::GetTensorData<int8_t>(filter),
+      tflite::micro::GetTensorShape(bias),
+      tflite::micro::GetTensorData<int32_t>(bias),
+      tflite::micro::GetTensorShape(output),
+      tflite::micro::GetTensorData<int8_t>(output));
+  return kTfLiteOk;
+}
+
+TfLiteRegistration Register_CONV_2D_INT8() {
```
advaitjain, line 129: let's call this `Register_CONV_2D_INT8REF`? to be explicit that we are giving the option to register the int8 variant of the reference kernel (since we may very well be adding an optimized int8 variant as well.

And add a comment:
```cc
// TODO(b/189981943): This variant can be used for a smaller binary
// since the optimized conv implementation currently adds a lot to
// the binary size (~30KB to text section).
```

tensorflow/lite/micro/tools/make/ext_libs/xtensa.inc
```diff
@@ -1,9 +1,10 @@
 # Explicitly add kernel sources specific to the Xtensa optimized
 # implementations.
 MICROLITE_CC_KERNEL_SRCS += \
-  tensorflow/lite/micro/kernels/xtensa/softmax_int8_int16.cc \
   tensorflow/lite/micro/kernels/xtensa/conv_hifi.cc \
-  tensorflow/lite/micro/kernels/xtensa/conv_hifimini.cc
+  tensorflow/lite/micro/kernels/xtensa/conv_hifimini.cc \
+  tensorflow/lite/micro/kernels/xtensa/conv_int8.cc \
```
advaitjain, line 6: reanme to conv_int8_reference.cc?

advaitjain (changes requested):
Can you have three separate xt-size outputs in the PR description? One for current tip of tree, one with the optimized conv kernel but the MicroMutableOpsResolver and the third with the int8_ref kernel variant.

That will help document what the difference between opt and ref for int8 was at the time of this PR.

tensorflow/lite/micro/kernels/xtensa/conv_int8.cc
njeffrie: Done. The savings from removing a few "if" statements in prepare seems to be minimal.

tensorflow/lite/micro/kernels/xtensa/conv_int8.cc
njeffrie: Done.

tensorflow/lite/micro/tools/make/ext_libs/xtensa.inc
njeffrie: Done.

advaitjain (approved):

tensorflow/lite/micro/kernels/conv.h
```diff
@@ -72,6 +72,16 @@ TfLiteStatus CalculateOpDataConv(TfLiteContext* context, TfLiteNode* node,
 
 TfLiteStatus ConvPrepare(TfLiteContext* context, TfLiteNode* node);
 
+#if defined(XTENSA)
+// Returns a TfLiteRegistration struct for kernel variant that only supports
+// int8 inputs and outputs.
+TfLiteRegistration Register_CONV_2D_INT8REF();
+#else 
+inline TfLiteRegistration Register_CONV_2D_INT8REF() {
+  return Register_CONV_2D();
```
advaitjain, line 81: `Register_CONV_2D` declaration needs to be moved from micro_ops.h to the current header.

njeffrie: Good call. Added conv.h to micro_mutable_op_resolver.h as well.

advaitjain (approved):

advaitjain (approved):

PR #178, opened 2021-06-16
Explicitly use python3 from the Makefile.

advaitjain:
For the Xtensa docker container, we were getting an error message along the lines of `python not found`. And the result was that the specialization was happening incorrectly (i.e. the specialization for the kernels was failing since it relies on python since #160).

https://github.com/tensorflow/tflite-micro/issues/182 is likely the reason why this error passed the CI for #160 but started failing after.


advaitjain:
I have instead updated the Xtensa docker image to install python.

rockyrhodes (approved):

PR #181, opened 2021-06-16
Add to include path for hexagon builds with the downloaded hexagon kernels.

njeffrie:


advaitjain (approved):
why is this needed?

advaitjain (changes requested):
why is this needed?

njeffrie:
Within the hexagon kernels themselves, there is:
```
#include "hexagon_tflm_translation_fully_connected.h"
```
which assumes we have third_party/hexagon in the include path.

PR #198, opened 2021-06-17
Remove hifimini support from the bazel build.

advaitjain:
* Consolidate all kernel BUILD targets into micro_ops
 * remove select statements for xtensa_hifimini

BUG=http://b/191406930


njeffrie (approved):
Seems reasonable. I assume this means we can no long build with cstubs (I assume this is OK, since we don't do so nowodays anyway).

advaitjain:
> Seems reasonable. I assume this means we can no long build with cstubs (I assume this is OK, since we don't do so nowodays anyway).

Yes. Since this was a google-only feature, I have documented the reasoning behind removing it at http://b/191406930

Issue #205, opened 2021-06-21
Xtensa optimized kernels for AveragePool and MaxPool


advaitjain:
https://github.com/tensorflow/tflite-micro/pull/139 is adding Xtensa optimized implementations for AveragePool and MaxPool.

Prior to being able to merge the Xtensa implementations, we will need to refactor the reference implementation (and the CMSIS optimized implementation) to better share code and reduce the maintenance overhead.

This issu will track the work needed for such a refactoring as well as getting the Xtensa kernels merged.

advaitjain:
Tagging @bhanuprakashbv @nyadla-sys @vp-cad @sjoshi768

petewarden:
Closing, since the refactoring appears to be done.

PR #206, opened 2021-06-21
Refactor reference implementation of pooling.cc

advaitjain:
* share code with the cmsis implementation
 * groundwork needed to add in Xtensa implementation

BUG=https://github.com/tensorflow/tflite-micro/issues/205


advaitjain:
Tagging @bhanuprakashbv @nyadla-sys @vp-cad @SJOSHI768

tensorflow/lite/micro/kernels/cmsis_nn/pooling.cc
```diff
@@ -15,99 +15,35 @@ limitations under the License.
...
-}
-
 void AverageEvalQuantized(TfLiteContext* context, const TfLiteNode* node,
                           const TfLitePoolParams* params, const OpData& data,
                           const TfLiteEvalTensor* input,
                           TfLiteEvalTensor* output) {
   TFLITE_DCHECK(input->type == kTfLiteInt8);
 
-  RuntimeShape input_shape = tflite::micro::GetTensorShape(input);
+  RuntimeShape input_shape = micro::GetTensorShape(input);
```
njeffrie, line 43: Somewhat unrelated to this change, do we plan to move away from RuntimeShape and MatchingDim for the sake of code size?

njeffrie (approved):

tensorflow/lite/micro/kernels/cmsis_nn/pooling.cc
advaitjain: I think we should, but we don't have a concrete plan for that right now.

PR #207, opened 2021-06-21
Add support for 8x16 kernels in TFLM.

njeffrie:
Add support for 8x16 kernels in TFLM.

Support 16-bit activations and 8-bit weights for a subset of the TFLM kernels.

BUG=http://b/184839019

### DIFF
--- a/tensorflow/lite/micro/kernels/sub.cc
+++ b/tensorflow/lite/micro/kernels/sub.cc
@@ -19,6 +19,7 @@ limitations under the License.
 #include "tensorflow/lite/c/common.h"
 #include "tensorflow/lite/kernels/internal/common.h"
 #include "tensorflow/lite/kernels/internal/quantization_util.h"
+#include "tensorflow/lite/kernels/internal/reference/add.h"
 #include "tensorflow/lite/kernels/internal/reference/process_broadcast_shapes.h"
 #include "tensorflow/lite/kernels/internal/tensor_ctypes.h"
 #include "tensorflow/lite/kernels/internal/types.h"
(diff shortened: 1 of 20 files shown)


tensorflow/lite/micro/kernels/sub.cc
```diff
@@ -187,7 +196,29 @@ TfLiteStatus EvalSubQuantized(TfLiteContext* context, TfLiteNode* node,
...
+            op_params, tflite::micro::GetTensorShape(input1),
+            tflite::micro::GetTensorData<int16_t>(input1),
+            tflite::micro::GetTensorShape(input2),
+            tflite::micro::GetTensorData<int16_t>(input2),
+            tflite::micro::GetTensorShape(output),
+            tflite::micro::GetTensorData<int16_t>(output), false);
+      }
+      break;
+    }
+    case kTfLiteUInt8: {
```
advaitjain, line 221: can we do a follow-on PR to remove uint8 support?

tensorflow/lite/micro/micro_utils.h
```diff
@@ -44,11 +44,11 @@ T FloatToQuantizedType(const float value, const float scale, int zero_point) {
 
 template <typename T>
 T FloatToSymmetricQuantizedType(const float value, const float scale) {
-  int32_t result = round(value / scale);
-  result =
-      std::max(static_cast<int32_t>(std::numeric_limits<T>::min() + 1), result);
-  result =
-      std::min(static_cast<int32_t>(std::numeric_limits<T>::max()), result);
+  std::int64_t result = round(value / scale);
```
advaitjain, line 47: why do we need to change to 64 bits? might be worth a comment.

advaitjain (changes requested):
nit comments only

tensorflow/lite/micro/kernels/sub.cc
njeffrie: Ack, will do.

tensorflow/lite/micro/micro_utils.h
njeffrie: Added a comment explaining that 8x16 conv requires a 64-bit bias, so numeric_limits<int64_t> must be an int64.

advaitjain (approved):

Issue #201, opened 2021-06-18
MicroProfiler difficult to subclass


alanvgreen:
I would like to make a subclass of [tflite::MicroProfiler](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/micro_profiler.h#L29) which formats data differently. In particular, I would like CSV output to make it easy to use performance data in a spreadsheet. 

However, since the data members of MicroProfiler are all private, and there are no accessors, it is difficult to build this as a subclass.

Here are several alternatives:

1. Make the data members protected instead of private.
2. Provide accessor methods so that the data collected by the profiler can be used by subclasses (and possibly by other classes too)
3. Split Split MicroProfiler into an interface and a separate implementation. 
4. Make a custom version of MicroProfiler.

What is the best option here? I'm happy to send a PR if that would be helpful.

advaitjain:
Having CSV output sounds generally useful so for this particular case I would say

  5. Add CSV output to tflite::MicroProfiler.

I'm not too concerned about the size increase since the ScopedMicroProfiler is a noop builds where we strip all the strings to reduce the size https://github.com/tensorflow/tflite-micro/blob/56e785e78810291625697dcdaffc66c4dbe32a44/tensorflow/lite/micro/micro_graph.cc#L164-L171

I can probably go with either 2, 3, or 5. And I'm possibly leaning towards 5 since csv output could be generally useful.

What do you think?

alanvgreen:
5 sounds good to me too. One less thing to maintain downstream :)

I'll send a change.

petewarden:
Closing this, since the pull request has been merged.

PR #208, opened 2021-06-22
Refactored MicroProfiler to use an Event struct for tracking events.

jackagoogle:
Using `Event` paired with `MicroProfiler::GetActiveEvents()` allows developers to access all of the event data for post-processing, such as exporting to a csv, or performing statistical analysis of latency across multiple runs of the same model.


advaitjain:
I haven't had a chance to look at the code, but tagging @alanvgreen since I'm guessing there is overlap between this PR and #201

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -56,15 +64,19 @@ class MicroProfiler {
   // Prints the profiling information of each of the events.
   void Log() const;
 
+  // Returns the number of active events that are being profiled.
+  uint32_t GetTotalEvents() const;
```
advaitjain, line 68: nit: have this return an int instead of uint32_t?

advaitjain (changes requested):
It seems useful to have some of the post-processing (such as exporting to csv) to be part of the tflite-micro codebase as well (e.g. #201).

What do you think about export CSV directly to MicroProfiler instead of adding an API to access the MicroProfiler::Event struct? Will that work for you?

tensorflow/lite/micro/micro_profiler.h
jackagoogle: Done.

jackagoogle:
> It seems useful to have some of the post-processing (such as exporting to csv) to be part of the tflite-micro codebase as well (e.g. #201).
> 
> What do you think about export CSV directly to MicroProfiler instead of adding an API to access the MicroProfiler::Event struct? Will that work for you?

I think there's benefits to both. The CSV will help with exporting raw data out for postprocessing in other languages, but some of what we've built just pretty prints the events in a nicer format for when we don't necessarily want to export to python.

We're also planning on exporting the allocation statistics as well, so maybe it makes sense to make the csv exporting more general purpose? Below's an example of what I've done with the `MicroProfiler::Event` struct.

```
Average Ticks 2014888
Tag                  Ticks       % CPU      stddev      
-------------------------------------------------------
CONV               | 478647    | 23.75    | 83        | 
FULLY_CONNECTED    | 86576     | 4.29     | 37        | 
CONV               | 201757    | 10.1     | 79        | 
FULLY_CONNECTED    | 86595     | 4.29     | 56        | 
CONV               | 201716    | 10.1     | 94        | 
FULLY_CONNECTED    | 86562     | 4.29     | 44        | 
CONV               | 201752    | 10.1     | 44        | 
FULLY_CONNECTED    | 86574     | 4.29     | 43        | 
CONV               | 156309    | 7.75     | 26        | 
CONV               | 183916    | 9.12     | 25        | 
CONV               | 183888    | 9.12     | 44        | 
FULLY_CONNECTED    | 8413      | 0.41     | 10        | 
SOFTMAX            | 52178     | 2.58     | 30        | 
-------------------------------------------------------
 
Unique Events
Average Ticks 2014888
Tag                  Ticks        % CPU      stddev      
--------------------------------------------------------
CONV               | 1607988    | 79.80    | 134       | 
FULLY_CONNECTED    | 354722     | 17.60    | 83        | 
SOFTMAX            | 52178      | 2.58     | 30        | 
--------------------------------------------------------
```

advaitjain:
Yeah, makes sense. My biggest concern is that the MicroProfiler::Event struct might change and break downstream tools.

If you are ok with either having your tools break every now and then, or putting the more general tools directly in tflite-micro, then we can go with the public Event struct.

One possible next step is:
 1. add csv export to MicroProfiler
 1. replacing the current printing in MicroProfiler with the pretty-print that you have (which looks really nice!).
 1. still keep the public Event struct if there are additional tools that you have that would not work with the csv export (with the understanding that the Event struct will be less stable than the csv format).

If there are no tools that will need the Event struct once we have 1. and 2. in the MicroProfiler class then maybe we can keep it private for now.

Thoughts?

jackagoogle:
Is there a particular reason why we couldn't make the actual backing data stores `protected`? This way i could subclass `MicroProfiler` and keep the `Event` struct internally to my needs as well.

This sounds fine to me, but I would also need to make sure that we can save a CSV for the `RecordingMicroAllocator` as well.

advaitjain:
Closing this PR since we have CSV output with https://github.com/tensorflow/tflite-micro/pull/218. We can revisit this again if needed.

PR #218, opened 2021-06-23
micro_profiler.cc: add CSV output

alanvgreen:
Adds a method to log output in CSV form, suitable for input to a
spreadsheet.

BUG=#201


advaitjain:
tagging @jackagoogle to avoid duplicated work.

tensorflow/lite/micro/micro_profiler.cc
```diff
@@ -55,4 +55,14 @@ void MicroProfiler::Log() const {
 #endif
 }
 
+void MicroProfiler::LogCsv() const {
+#if !defined(TF_LITE_STRIP_ERROR_STRINGS)
+  MicroPrintf("\"Event\",\"Tag\",\"Ticks\",\"ms\"");
+  for (int i = 0; i < num_events_; ++i) {
+    int32_t ticks = end_ticks_[i] - start_ticks_[i];
+    MicroPrintf("%d,%s,%d,%d", i, tags_[i], ticks, TicksToMs(ticks));
```
jackagoogle, line 63: I've found the ms to be pretty useless since it doesn't take into effect the actual clock freq. I'd be open to adding the freq to the `MicroProfiler` constructor to get a better estimation of duration.

jackagoogle (approved):

tensorflow/lite/micro/micro_profiler.cc
advaitjain: Unless @alanvgreen needs this, I'm ok skipping the ms as well for the csv output. We really should just go with ticks for all analysis.

alanvgreen: I don't feel strongly about having ms in the CSV output. "Ticks" are the important metric for our work - ms are secondary.

However, @jackagoogle I'd like to understand why you are finding that the ms conversion doesn't take into account the actual clock frequency. Are you working in a codebase where you are able to implement GetCurrentTicks() satisfactorily but not ticks_to_ms()?

alanvgreen: I'm unclear on next steps. Is ms a blocker? I'd prefer to keep it, but don't feel strongly.

To my mind, if ticks_to_ms is not able to be implemented properly on some platforms, then a decision needs to be made to remove that mechanism or clearly document where it doesn't work.

The current ms calculation, using ticks_to_ms(), works fine for any system that can count ticks based on wall time. This includes a number of systems from small MCUs (which typically have specialized timer peripherals) to x86 Linux (which has millisecond accurate timers synchronized to a real time clock). I'd like to hear about which systems this doesn't work for, because I'm sure they're interesting :)

advaitjain: Apologies for the delay here, just got back from leave. For this PR, let's remove ms from the CSV output.

I would like to have most analysis use ticks, MCPS etc. and it would be nice to have the CSV also steer users in that direction.

I would guess that @jackagoogle is encountering an issue with the fact that TFLM currently does not allow for multiple implementations of [ticks_per_second](https://github.com/tensorflow/tflite-micro/blob/a7f2a887bd947c7378d8eada2d0b82dd5a8c1ca9/tensorflow/lite/micro/micro_time.h#L22-L24) for the same `TARGET=<blah>`.

Regardless, let's have that conversation separate from merging this PR.

alanvgreen:
Thanks for the clear direction. I have removed milliseconds from this PR.

advaitjain (approved):

PR #220, opened 2021-06-24
Support 16-bit add kernel

njeffrie:
Add support for 16-bit add, 8bit-16bit quantize, transpose_conv 16-bit bias and add relevant tests.

BUG=http://b/184839019

### DIFF
--- a/tensorflow/lite/micro/kernels/add.cc
+++ b/tensorflow/lite/micro/kernels/add.cc
@@ -66,12 +66,12 @@ TfLiteStatus CalculateOpData(TfLiteContext* context, TfLiteAddParams* params,
                              OpData* data) {
   data->requires_broadcast = !HaveSameShapes(input1, input2);
 
-  if (output->type == kTfLiteInt8) {
+  if (output->type == kTfLiteInt8 || output->type == kTfLiteInt16) {
     // 8bit -> 8bit general quantized path, with general rescalings
     data->input1_offset = -input1->params.zero_point;
     data->input2_offset = -input2->params.zero_point;
     data->output_offset = output->params.zero_point;
-    data->left_shift = 20;
+    data->left_shift = (output->type == kTfLiteInt16) ? 15 : 20;
     const double twice_max_input_scale =
         2 * static_cast<double>(
                 std::max(input1->params.scale, input2->params.scale));
@@ -150,21 +150,51 @@ TfLiteStatus EvalAddQuantized(TfLiteContext* context, TfLiteNode* node,
       tflite::micro::GetTensorShape(input1),
       tflite::micro::GetTensorShape(input2), &op_params);
 
-  if (need_broadcast) {
-    reference_integer_ops::BroadcastAdd4DSlow(
-        op_params, tflite::micro::GetTensorShape(input1),
-        tflite::micro::GetTensorData<int8_t>(input1),
-        tflite::micro::GetTensorShape(input2),
-        tflite::micro::GetTensorData<int8_t>(input2),
-        tflite::micro::GetTensorShape(output),
-        tflite::micro::GetTensorData<int8_t>(output));
-  } else {
-    reference_integer_ops::Add(op_params, tflite::micro::GetTensorShape(input1),
-                               tflite::micro::GetTensorData<int8_t>(input1),
-                               tflite::micro::GetTensorShape(input2),
-                               tflite::micro::GetTensorData<int8_t>(input2),
-                               tflite::micro::GetTensorShape(output),
-                               tflite::micro::GetTensorData<int8_t>(output));
+  switch (output->type) {
+    case kTfLiteInt8: {
+      if (need_broadcast) {
+        reference_integer_ops::BroadcastAdd4DSlow(
+            op_params, tflite::micro::GetTensorShape(input1),
+            tflite::micro::GetTensorData<int8_t>(input1),
+            tflite::micro::GetTensorShape(input2),
+            tflite::micro::GetTensorData<int8_t>(input2),
+            tflite::micro::GetTensorShape(output),
+            tflite::micro::GetTensorData<int8_t>(output));
+      } else {
+        reference_integer_ops::Add(
+            op_params, tflite::micro::GetTensorShape(input1),
+            tflite::micro::GetTensorData<int8_t>(input1),
+            tflite::micro::GetTensorShape(input2),
+            tflite::micro::GetTensorData<int8_t>(input2),
+            tflite::micro::GetTensorShape(output),
+            tflite::micro::GetTensorData<int8_t>(output));
+      }
+      break;
+    }
+    case kTfLiteInt16: {
+      if (need_broadcast) {
+        reference_ops::BroadcastAdd4DSlow(
+            op_params, tflite::micro::GetTensorShape(input1),
+            tflite::micro::GetTensorData<int16_t>(input1),
+            tflite::micro::GetTensorShape(input2),
+            tflite::micro::GetTensorData<int16_t>(input2),
+            tflite::micro::GetTensorShape(output),
+            tflite::micro::GetTensorData<int16_t>(output));
+      } else {
+        reference_ops::Add(op_params, tflite::micro::GetTensorShape(input1),
+                           tflite::micro::GetTensorData<int16_t>(input1),
+                           tflite::micro::GetTensorShape(input2),
+                           tflite::micro::GetTensorData<int16_t>(input2),
+                           tflite::micro::GetTensorShape(output),
+                           tflite::micro::GetTensorData<int16_t>(output),
+                           false);
+      }
+      break;
+    }
+    default:
+      TF_LITE_KERNEL_LOG(context, "Type %s (%d) not supported.",
+                         TfLiteTypeGetName(output->type), output->type);
+      return kTfLiteError;
   }
 
   return kTfLiteOk;
@@ -210,7 +240,7 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
 
   if (output->type == kTfLiteFloat32) {
     EvalAdd(context, node, params, data, input1, input2, output);
-  } else if (output->type == kTfLiteInt8) {
+  } else if (output->type == kTfLiteInt8 || output->type == kTfLiteInt16) {
     TF_LITE_ENSURE_OK(context, EvalAddQuantized(context, node, params, data,
                                                 input1, input2, output));
   } else {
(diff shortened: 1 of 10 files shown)


tensorflow/lite/micro/kernels/add.cc
```diff
@@ -150,21 +150,51 @@ TfLiteStatus EvalAddQuantized(TfLiteContext* context, TfLiteNode* node,
...
+                           tflite::micro::GetTensorData<int16_t>(input2),
+                           tflite::micro::GetTensorShape(output),
+                           tflite::micro::GetTensorData<int16_t>(output),
+                           false);
+      }
+      break;
+    }
+    default:
+      TF_LITE_KERNEL_LOG(context, "Type %s (%d) not supported.",
+                         TfLiteTypeGetName(output->type), output->type);
```
advaitjain, line 196: let's switch this to MicroPrintf. I'd like to incrementally move away from the KERNEL_LOG macros.

tensorflow/lite/micro/kernels/add_test.cc
```diff
@@ -362,4 +362,132 @@ TF_LITE_MICRO_TEST(QuantizedAddWithMixedBroadcastInt8) {
   }
 }
 
+TF_LITE_MICRO_TEST(QuantizedAddNoActivationInt16) {
+  const float scales[] = {0.25, 0.5, 1.0};
+  const int zero_points[] = {0, 0, 0};
+  const int output_dims_count = 4;
```
advaitjain, line 368: constexpr, and move closer to first use?

tensorflow/lite/micro/kernels/add_test.cc
```diff
@@ -362,4 +362,132 @@ TF_LITE_MICRO_TEST(QuantizedAddWithMixedBroadcastInt8) {
...
+      inout_shape, input1_values, input1_quantized, scales[0], zero_points[0],
+      inout_shape, input2_values, input2_quantized, scales[1], zero_points[1],
+      inout_shape, golden_values, golden_quantized, scales[2], zero_points[2],
+      kTfLiteActNone, output);
+}
+
+TF_LITE_MICRO_TEST(QuantizedAddActivationRelu1Int16) {
+  const float scales[] = {0.25, 0.5, 1.0};
+  const int zero_points[] = {0, 0, 0};
+  const int output_dims_count = 4;
```
advaitjain, line 389: constexpr, and move closer to first use?

tensorflow/lite/micro/kernels/add_test.cc
```diff
@@ -362,4 +362,132 @@ TF_LITE_MICRO_TEST(QuantizedAddWithMixedBroadcastInt8) {
...
+      inout_shape, input1_values, input1_quantized, scales[0], zero_points[0],
+      inout_shape, input2_values, input2_quantized, scales[1], zero_points[1],
+      inout_shape, golden_values, golden_quantized, scales[2], zero_points[2],
+      kTfLiteActReluN1To1, output);
+}
+
+TF_LITE_MICRO_TEST(QuantizedAddVariousInputShapesInt16) {
+  const float scales[] = {0.1, 0.05, 0.1};
+  const int zero_points[] = {0, 0, 0};
+  const int output_dims_count = 6;
```
advaitjain, line 410: constexpr, and move closer to first use?

tensorflow/lite/micro/kernels/add_test.cc
```diff
@@ -362,4 +362,132 @@ TF_LITE_MICRO_TEST(QuantizedAddWithMixedBroadcastInt8) {
...
+    tflite::testing::TestAddQuantized(
+        test_shapes[i], input1_values, input1_quantized, scales[0],
+        zero_points[0], test_shapes[i], input2_values, input2_quantized,
+        scales[1], zero_points[1], test_shapes[i], golden_values,
+        golden_quantized, scales[2], zero_points[2], kTfLiteActNone, output);
+  }
+}
+
+TF_LITE_MICRO_TEST(QuantizedAddWithScalarBroadcastInt16) {
+  const int output_dims_count = 6;
```
advaitjain, line 440: constexpr, and move closer to first use?

tensorflow/lite/micro/kernels/add_test.cc
advaitjain: And this (mis)use of a const instead of constexpr for array sizes seems to be common across this entire file. Maybe fix all the occurrences in a follow-up PR?

tensorflow/lite/micro/kernels/transpose_conv.cc
```diff
@@ -105,6 +106,19 @@ TfLiteStatus CalculateOpData(TfLiteContext* context, TfLiteNode* node,
         &data->params.quantized_activation_max,
         data->per_channel_output_multiplier, data->per_channel_output_shift,
         output_channels));
+
+    // Some models have int16 biases which must be converted to int64 for
+    // inference.
```
advaitjain, line 111: can we have a bug referenced here? It would be good to have additional context for why this is the case and is this a short term workaround or something that we have to live with.

tensorflow/lite/micro/kernels/transpose_conv.cc
```diff
@@ -105,6 +106,19 @@ TfLiteStatus CalculateOpData(TfLiteContext* context, TfLiteNode* node,
...
+    if (input->type == kTfLiteInt16) {
+      TFLITE_DCHECK(filter->type == kTfLiteInt8);
+      TFLITE_DCHECK(output->type == kTfLiteInt16);
+      if (bias->type == kTfLiteInt16) {
+        TFLITE_DCHECK(
+            context->RequestScratchBufferInArena(
+                context, GetTensorShape(bias).FlatSize() * sizeof(std::int64_t),
+                &(data->bias_converted_buffer_index)) == kTfLiteOk);
```
advaitjain, line 119: seems wasteful from the standpoint of memory footprint. Is it possibly better to have a flatbuffer modification so that int64 biases can be stored directly in the flatbuffer? Again, we don't need it fixed in this PR, but would be good to have a bug and circle back.

tensorflow/lite/micro/kernels/transpose_conv.cc
```diff
@@ -259,17 +273,38 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
-          tflite::micro::GetTensorShape(nullptr), nullptr, scratch_buffer);
+      // Handle models with int16 biases by converting to int64 on the fly.
+      if (bias->type == kTfLiteInt16) {
+        std::int64_t* bias_converted_buffer =
+            static_cast<int64_t*>(context->GetScratchBuffer(
+                context, data.bias_converted_buffer_index));
+        for (int i = 0; i < tflite::micro::GetTensorShape(bias).FlatSize();
+             i++) {
+          bias_converted_buffer[i] = bias->data.i16[i];
+        }
```
advaitjain, line 284: the complexity in the inference code makes me wonder if we should do something else.

tensorflow/lite/micro/test_helpers.cc
```diff
@@ -1198,6 +1198,20 @@ TfLiteFloatArray* FloatArrayFromFloats(const float* floats) {
...
 }
 
+TfLiteTensor CreateQuantizedBiasTensor(const float* data, int16_t* quantized,
+                                       TfLiteIntArray* dims, float input_scale,
+                                       float weights_scale, bool is_variable) {
+  float bias_scale = input_scale * weights_scale;
+  tflite::SymmetricQuantize(data, quantized, ElementCount(*dims), bias_scale);
+
+  // Quantized int32_t tensors always have a zero point of 0, since the range of
+  // int32_t values is large, and because zero point costs extra cycles during
```
advaitjain, line 1208: this comment seems a bit out of place?

advaitjain (changes requested):
Biggest question is about transpose_conv. Maybe have that be a separate PR and we can discuss in more detail?

tensorflow/lite/micro/kernels/transpose_conv.cc
njeffrie: Fortunately it seems to have little effect on overall memory, since it's a scratch buffer rather than a persistent buffer. Still, I filed b/192090531 and referenced it here to track this issue.

tensorflow/lite/micro/kernels/transpose_conv.cc
njeffrie: Done. Created b/192090531

tensorflow/lite/micro/kernels/transpose_conv.cc
njeffrie: As this is a temporary workaround, I would prefer to keep it as-is. I cannot think of a simpler immediate solution. Model reworking would be both more complex and likely throwaway work. I'm open to suggestions, however.

tensorflow/lite/micro/test_helpers.cc
njeffrie: Ack. Changed to int16 since it still applies, just had the wrong type.

tensorflow/lite/micro/kernels/add.cc
njeffrie: Filed b/192091017 to replace TF_LITE_KERNEL_LOG with MicroPrintf across all kernels. I would prefer to do this all in one go. Currently we use TF_LITE_KERNEL_LOG everywhere else in our [kernels](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/tensorflow/tflite-micro%24+file:tensorflow/lite/micro/kernels+MicroPrintf&patternType=literal)

tensorflow/lite/micro/kernels/add_test.cc
njeffrie: I find-and-replaced all "const int output_dims_count" with "constexpr int output_dims_count" here since it's pretty straightforward.

tensorflow/lite/micro/kernels/add_test.cc
njeffrie: Done.

tensorflow/lite/micro/kernels/add_test.cc
njeffrie: Done.

advaitjain (dismissed):

advaitjain (approved):

advaitjain:
We really should refactor the CMSIS kernels to call into the reference code for all types other than int8. Maybe follow-on PR?

njeffrie:
Agreed. Filed b/192114220.

PR #222, opened 2021-06-24
Add a way to get the interpreter from a benchmark runner.

njeffrie:
Add API to support multiple inputs from a benchmark runner.

BUG=http://b/184839019

### DIFF
--- a/tensorflow/lite/micro/benchmarks/micro_benchmark.h
+++ b/tensorflow/lite/micro/benchmarks/micro_benchmark.h
@@ -48,10 +48,10 @@ class MicroBenchmarkRunner {
     }
   }
 
-  void SetRandomInput(const int random_seed) {
+  void SetRandomInput(const int random_seed, int input_index = 0) {
     // The pseudo-random number generator is initialized to a constant seed
     std::srand(random_seed);
-    TfLiteTensor* input = interpreter_.input(0);
+    TfLiteTensor* input = interpreter_.input(input_index);
 
     // Pre-populate input tensor with random values.
     int input_length = input->bytes / sizeof(inputT);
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/benchmarks/micro_benchmark.h
```diff
@@ -77,6 +77,10 @@ class MicroBenchmarkRunner {
     interpreter_.GetMicroAllocator().PrintAllocations();
   }
 
+  tflite::RecordingMicroInterpreter* GetInterpreter() {
```
advaitjain, line 80: nit: let's return a mutable reference. That way we know it can not be nullptr.

http://go/totw/labs/return-by-pointer-or-reference#discouraged-alternative

advaitjain (changes requested):
I'm not a huge fan of exposing the interpreter since anyone using the benchmark should not be using most of the interpreter API.

Is adding an input index param (default to 0) to https://github.com/tensorflow/tflite-micro/blob/784e08be0131db591cdc9e3300eaa1027870a1b8/tensorflow/lite/micro/benchmarks/micro_benchmark.h#L51

and https://github.com/tensorflow/tflite-micro/blob/784e08be0131db591cdc9e3300eaa1027870a1b8/tensorflow/lite/micro/benchmarks/micro_benchmark.h#L67

sufficient?

That way we get to reuse the SetRandomInput code for multiple inputs as well.

tensorflow/lite/micro/benchmarks/micro_benchmark.h
njeffrie: Done.

njeffrie:
Good call. I've updated the change to use input index as a new argument to SetInput and SetRandomInput, defaulting the index to 0.

advaitjain (approved):

advaitjain (approved):

advaitjain (approved):

PR #243, opened 2021-07-02
Provide a lightweight version of flexbuffers::Vector

shlmregev:
- Save ~4KB of RAM by implementing a lightweight version of the vector, which only accesses integer values, without implicit conversion.
- Change Micro kernels to access flexbuffers as vectors instead of maps, which is recommended by the flexbuffers doc for efficiency. Since the values in the vector are ordered alphabetically by their keys, the kernels can access them by index instead.
- Revert detection_postprocess to the native flexbuffers API. The
LiteVector API doesn't support IsNull() and I prefer not to support two
custom flexbuffer APIs.

BUG=b/192311485

### DIFF
--- a/tensorflow/lite/micro/kernels/circular_buffer.cc
+++ b/tensorflow/lite/micro/kernels/circular_buffer.cc
@@ -54,6 +54,11 @@ namespace {
 constexpr int kInputTensor = 0;
 constexpr int kOutputTensor = 0;
 
+// Indices into the init flexbuffer's vector.
+// The parameter's name is in the comment that follows.
+// Elements in the vectors are ordered alphabetically by parameter name.
+constexpr int kCyclesMaxIndex = 0;  // 'cycles_max'
+
 // TODO(b/149795762): Add this to TfLiteStatus enum.
 constexpr TfLiteStatus kTfLiteAbort = static_cast<TfLiteStatus>(-9);
 
@@ -74,10 +79,9 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
 
   if (buffer != nullptr && length > 0) {
     const uint8_t* buffer_t = reinterpret_cast<const uint8_t*>(buffer);
-    const flexbuffers::Map& m =
-        tflite::micro::FlexbuffersWrapperGetRootAsMap(buffer_t, length);
-    op_data->cycles_max =
-        tflite::micro::FlexbuffersWrapperAsInt32(m, "cycles_max");
+    const tflite::micro::LiteVector& lv =
+        tflite::micro::GetFlexbufferRootAsLiteVector(buffer_t, length);
+    op_data->cycles_max = lv.ElementAsInt32(kCyclesMaxIndex);
   } else {
     op_data->cycles_max = 0;
   }
--- a/tensorflow/lite/micro/kernels/circular_buffer.cc
+++ b/tensorflow/lite/micro/kernels/circular_buffer.cc
@@ -20,6 +20,7 @@ limitations under the License.
 #include "tensorflow/lite/kernels/internal/tensor_ctypes.h"
 #include "tensorflow/lite/kernels/kernel_util.h"
 #include "tensorflow/lite/kernels/op_macros.h"
+#include "tensorflow/lite/micro/flatbuffer_utils.h"
 #include "tensorflow/lite/micro/kernels/kernel_util.h"
 
 /*
@@ -79,9 +80,8 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
 
   if (buffer != nullptr && length > 0) {
     const uint8_t* buffer_t = reinterpret_cast<const uint8_t*>(buffer);
-    const tflite::micro::LiteVector& lv =
-        tflite::micro::GetFlexbufferRootAsLiteVector(buffer_t, length);
-    op_data->cycles_max = lv.ElementAsInt32(kCyclesMaxIndex);
+    tflite::FlexbufferWrapper wrapper(buffer_t, length);
+    op_data->cycles_max = wrapper.ElementAsInt32(kCyclesMaxIndex);
   } else {
     op_data->cycles_max = 0;
   }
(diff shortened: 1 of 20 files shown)


njeffrie (dismissed):

njeffrie (approved):

tensorflow/lite/micro/kernels/kernel_util.h
```diff
@@ -31,16 +31,17 @@ namespace micro {
...
-                                                      size_t size);
-
-int32_t FlexbuffersWrapperAsInt32(const flexbuffers::Map& m, const char* key);
-
-bool FlexbuffersWrapperAsBool(const flexbuffers::Map& m, const char* key);
-
-float FlexbuffersWrapperAsFloat(const flexbuffers::Map& m, const char* key);
-
-bool FlexbuffersWrapperIsNull(const flexbuffers::Map& m, const char* key);
+class LiteVector : public flexbuffers::Vector {
```
advaitjain, line 34: nit: can we use a more descriptive name than LiteVector?

Maybe FlexbufferVectorWrapper?

tensorflow/lite/micro/kernels/kernel_util.h
```diff
@@ -31,16 +31,17 @@ namespace micro {
...
+ public:
+  LiteVector(const flexbuffers::Vector& v);
+  int64_t ElementAsInt64(size_t i) const;
+  uint64_t ElementAsUInt64(size_t i) const;
+  int32_t ElementAsInt32(size_t i) const;
+  bool ElementAsBool(size_t i) const;
+};
+
+const LiteVector GetFlexbufferRootAsLiteVector(const uint8_t* buffer,
```
advaitjain, line 43: nit: document the parameters.

and remove the const from the type that is returned by value: http://go/totw/labs/const#when-should-i-use-const-in-return-types

Actually, it might be more readable to make this a constructor for the LiteVector class:
```cc
class FlexbufferVectorWrapper {

public:
  explicit FlexbufferVectorWrapper(const uint8_t* buffer, size_t size);

...
}

```

And the usage would become:
```
op_data->cycles_max = tflite::FlexbufferVectorWrapper(buffer_t, length).ElementAsInt32(kCyclesMaxIndex);
```

tensorflow/lite/micro/kernels/kernel_util.cc
```diff
@@ -20,25 +20,29 @@ limitations under the License.
...
-  return m[key].AsFloat();
+bool LiteVector::ElementAsBool(size_t i) const {
+  return static_cast<bool>(LiteVector::ElementAsUInt64(i));
 }
 
-bool FlexbuffersWrapperIsNull(const flexbuffers::Map& m, const char* key) {
-  return m[key].IsNull();
+const LiteVector GetFlexbufferRootAsLiteVector(const uint8_t* buffer,
+                                               size_t size) {
+  return static_cast<LiteVector>(flexbuffers::GetRoot(buffer, size).AsVector());
```
advaitjain, line 45: I'm a bit confused by the static_cast and I'm not confident that casting is correct.

Might be better to explicitly create a LiteVector object?

```cc
return LiteVector(*flexbuffers::GetRoot(buffer, size).AsVector());
```

tensorflow/lite/micro/kernels/kernel_util.h
```diff
@@ -31,16 +31,17 @@ namespace micro {
...
-int32_t FlexbuffersWrapperAsInt32(const flexbuffers::Map& m, const char* key);
-
-bool FlexbuffersWrapperAsBool(const flexbuffers::Map& m, const char* key);
-
-float FlexbuffersWrapperAsFloat(const flexbuffers::Map& m, const char* key);
-
-bool FlexbuffersWrapperIsNull(const flexbuffers::Map& m, const char* key);
+class LiteVector : public flexbuffers::Vector {
+ public:
+  LiteVector(const flexbuffers::Vector& v);
```
advaitjain, line 36: nit: mark constructor as explicit (http://go/totw/14)

tensorflow/lite/micro/kernels/circular_buffer.cc
```diff
@@ -54,6 +54,11 @@ namespace {
 constexpr int kInputTensor = 0;
 constexpr int kOutputTensor = 0;
 
+// Indices into the init flexbuffer's vector.
+// The parameter's name is in the comment that follows.
+// Elements in the vectors are ordered alphabetically by parameter name.
```
advaitjain, line 59: can you expand this comment to describe how this ordering is achieved? Does the flexbuffer implementation enforce this or something else?

tensorflow/lite/micro/kernels/kernel_util.h
advaitjain: why do we need to inherit from flexbuffers::Vector?

advaitjain (changes requested):
Similar to https://github.com/tensorflow/tflite-micro/pull/245#discussion_r665603110, it might make sense to create a flatbuffer_utils.h/flatbuffer_utils.cc module.

tensorflow/lite/micro/kernels/kernel_util.h
shlmregev: We need access to   
const uint8_t *data_;
uint8_t byte_width_;
Both are protected members in Object, which is Vector's ancestor class.

shlmregev: sgtm re. naming.

tensorflow/lite/micro/kernels/kernel_util.cc
shlmregev: Done

tensorflow/lite/micro/kernels/kernel_util.h
shlmregev: Done

tensorflow/lite/micro/kernels/circular_buffer.cc
shlmregev: Done in flatbuffer_utils.h

advaitjain (approved):

advaitjain (dismissed):

advaitjain (approved):

advaitjain (approved):

advaitjain (approved):

PR #245, opened 2021-07-02
Check for the case where the Operators vector in a tflite model is missing

shlmregev:
Some of our tools remove the vector if it's empty (i.e. the subgraph doesn't contain ops). The check is added until the tools are fixed.

Note that to properly build the code, we needed to add `#define FLATBUFFERS_LOCALE_INDEPENDENT 0` to a number of sources and headers in the TFLM code. http://b/193264978 has more context on why that was needed and what options we might have to get the same outcome in a more maintainable manner.

BUG=http://b/192589496

### DIFF
--- a/tensorflow/lite/micro/micro_graph.cc
+++ b/tensorflow/lite/micro/micro_graph.cc
@@ -18,6 +18,7 @@ limitations under the License.
 #include "flatbuffers/flatbuffers.h"  // from @flatbuffers
 #include "tensorflow/lite/c/common.h"
 #include "tensorflow/lite/kernels/internal/compatibility.h"
+#include "tensorflow/lite/micro/flatbuffer_utils.h"
 #include "tensorflow/lite/micro/memory_helpers.h"
 #include "tensorflow/lite/micro/micro_error_reporter.h"
 #include "tensorflow/lite/micro/micro_profiler.h"
@@ -57,7 +58,7 @@ TfLiteStatus MicroGraph::InitSubgraphs() {
   for (size_t subgraph_idx = 0; subgraph_idx < subgraphs_->size();
        subgraph_idx++) {
     current_subgraph_index_ = subgraph_idx;
-    int operators_size = NumSubgraphOperators(subgraph_idx);
+    uint32_t operators_size = NumSubgraphOperators(model_, subgraph_idx);
     for (size_t i = 0; i < operators_size; ++i) {
       TfLiteNode* node =
           &(subgraph_allocations_[subgraph_idx].node_and_registrations[i].node);
@@ -91,7 +92,7 @@ TfLiteStatus MicroGraph::PrepareSubgraphs() {
   for (size_t subgraph_idx = 0; subgraph_idx < subgraphs_->size();
        subgraph_idx++) {
     current_subgraph_index_ = subgraph_idx;
-    int operators_size = NumSubgraphOperators(subgraph_idx);
+    uint32_t operators_size = NumSubgraphOperators(model_, subgraph_idx);
     for (size_t i = 0; i < operators_size; ++i) {
       TfLiteNode* node =
           &(subgraph_allocations_[subgraph_idx].node_and_registrations[i].node);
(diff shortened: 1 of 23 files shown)


njeffrie (approved):
Looks good. Hopefully we can revert this once the flatbuffer API respects empty subgraphs. Perhaps reference a bug indicating this in the PR description.

tensorflow/lite/micro/micro_graph.cc
```diff
@@ -94,7 +101,14 @@ TfLiteStatus MicroGraph::PrepareSubgraphs() {
     current_subgraph_index_ = subgraph_idx;
 
     const SubGraph* subgraph = (*subgraphs_)[subgraph_idx];
-    for (size_t i = 0; i < subgraph->operators()->size(); ++i) {
```
njeffrie, line 97: Do you think we can split this repeated code into a helper method?

tensorflow/lite/micro/micro_interpreter.cc
```diff
@@ -96,7 +96,15 @@ TfLiteStatus MicroInterpreter::PrepareNodeAndRegistrationDataFromFlatbuffer() {
     auto* opcodes = model_->operator_codes();
     BuiltinDataAllocator* builtin_data_allocator =
         allocator_.GetBuiltinDataAllocator();
-    for (size_t i = 0; i < subgraph->operators()->size(); ++i) {
+
+    // TODO(b/192589496): Ops must always be there. Remove this check when fixed
+    int operators_size;
```
njeffrie, line 101: Maybe even this can call the helper method, if it's a public method in MicroGraph.

njeffrie (changes requested):

tensorflow/lite/micro/micro_graph.cc
shlmregev: Done

njeffrie (dismissed):

tensorflow/lite/micro/micro_graph.cc
```diff
@@ -244,4 +241,14 @@ TfLiteEvalTensor* MicroGraph::GetSubgraphOutput(int subgraph_idx,
   return &subgraph_allocations_[subgraph_idx].tensors[tensor_idx];
 }
 
+int MicroGraph::NumSubgraphOperators(int subgraph_idx) {
```
advaitjain, line 244: there isn't any reason for this function to be a public member of the MicroGraph class.

Can we make it a free function:
```cc
int NumSubgraphOperators(const Subgraph& subgraph);
```

And have this function be used across micro_graph.cc and micro_allocator.cc?

I'd be ok with putting this and the code from https://github.com/tensorflow/tflite-micro/pull/243 into a flatbuffer_utils.h/flatbuffer_utils.cc module.

advaitjain (changes requested):

tensorflow/lite/micro/micro_graph.cc
shlmregev: Is this the path you had in mind:
tensorflow/lite/micro/flatbuffer_utils.*

advaitjain: yup

tensorflow/lite/micro/micro_graph.h
```diff
@@ -60,12 +60,15 @@ class MicroGraph {
...
 
-  // Number of tensor outputs to a specified subgraph in the model.
+  // Number of tensor outputs from a specified subgraph in the model.
   virtual size_t NumSubgraphOutputs(int subgraph_idx);
 
   // Get the specified output tensor of a specified subgraph in the model.
   virtual TfLiteEvalTensor* GetSubgraphOutput(int subgraph_idx, int output_idx);
 
+  // Number of operators in the specified subgraph in the model.
+  virtual int NumSubgraphOperators(int subgraph_idx);
```
advaitjain, line 70: I think I got confused in the internal review of this change. If this is a new function added in the current change, I think it is not needed anymore (now that we have the free functions in flatbuffer_utils.cc).

tensorflow/lite/micro/micro_graph.cc
advaitjain: I think this function is now unused and can be removed?

advaitjain (changes requested):
One clarification about whether we still need `MicroGraph::NumSubgraphOperators`.

Other than that, lgtm.

tensorflow/lite/micro/micro_graph.h
shlmregev: I made a mess by merging the origin branch before pushing my update. It overwrote my changes.

shlmregev: OK, I think I fixed it.

advaitjain (dismissed):

advaitjain (approved):

advaitjain (approved):

advaitjain (dismissed):

advaitjain (approved):

Issue #242, opened 2021-07-01
The kernel sources are compiled into both MICROLITE_LIB_OBJS and MICROLITE_KERNELS_OBJS


jenselofsson:
**System information**
- Host OS Platform and Distribution: Ubuntu 18.04
- TensorFlow installed from: source
- Tensorflow version (commit SHA if source): e172d86

**Describe the problem**
We've noticed that the kernel sources seems to be compiled into both the MICROLITE_LIB_OBJS and MICROLITE_KERNEL_OBJS. This leads to the kernels sources being compiled with both CORE_OPTIMIZATION_LEVEL (-Os) and KERNEL_OPTIMIZATION_LEVEL (-O2) depending on which object file they're found in.

This might lead to the final binary using kernel source being built with -Os, instead of -O2 which means that the execution of the kernels will be slower. armclang will also display a warning saying `L6876W: Minor variants of archive member '<member>' include multiple base variants`

Removing line 612 from [tensorflow/lite/micro/tools/make/Makefile](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/tools/make/Makefile#L612) seems to solve the issue at first glance.

**Please provide the exact sequence of commands/steps when you ran into the problem**
`$ make -f tensorflow/lite/micro/tools/make/Makefile TARGET=cortex_m_corstone_300 TARGET_ARCH=cortex-m55 BUILD_TYPE=release_with_logs TOOLCHAIN=armclang test_hello_world_test`

jenselofsson:
The problem seems to be a bit more complex than the [proposed solution](https://github.com/jenselofsson/tflite-micro/commit/36f13902052d5c7b6a8ac8503301aee2b2f3598c). If I run the script tensorflow/lite/micro/tools/ci_build/test_all.sh, the Arduino build fails with the following errors:
```
/tmp/tflite-arduino-build/libraries/tensorflow_lite/tensorflow/lite/micro/all_ops_resolver.cpp.o: In function `tflite::AllOpsResolver::AllOpsResolver()':
/home/username/Arduino/libraries/tensorflow_lite/src/tensorflow/lite/micro/micro_mutable_op_resolver.h:118: undefined reference to `tflite::ops::micro::Register_ABS()'
/home/username/Arduino/libraries/tensorflow_lite/src/tensorflow/lite/micro/micro_mutable_op_resolver.h:123: undefined reference to `tflite::ops::micro::Register_ADD()'
[...]

PR #258, opened 2021-07-08
Stop the kernel sources from being compiled into both the core objects and kernel objects.

jenselofsson:
Solves #242. The kernel sources are being compiled into both the core objects and kernels objects, which are built with different optimization flags. If the kernels are linked from the core objects, this leads to slower execution of the kernels.

BUG=#242


njeffrie (commented):
Try adding changing line 835 from:

list_library_sources:
	@echo $(MICROLITE_CC_SRCS)

to:

list_library_sources:
	@echo $(MICROLITE_CC_SRCS) \
    @echo $(MICROLITE_CC_KERNEL_SRCS)

This should fix the project generation + arduino errors.

advaitjain (approved):

PR #275, opened 2021-07-13
Have a TFLM-specific version of lite/kernels/op_macros.h

advaitjain:
* After this change, we can remove the TFLM-specific code from upstream Tensorflow.
 * Both the TfLite and TFLM implementations of op_macros.h will be simplified.
 * Corresponding change in Tensorflow: http://cl/384575900

BUG=http://b/187728891


njeffrie (dismissed):
Looks good. I assume we will clean up the TF version of this file separately?

njeffrie (approved):

PR #227, opened 2021-06-29
Updated convolution kernel to work with embARC MLI Library 2.0 for ARC

gerbauz:
[1/11]

This PR brings updates for ARC convolution kernel to support [embARC MLI Library 2.0](https://github.com/foss-for-synopsys-dwc-arc-processors/embarc_mli/tree/Release_2.0_EA) as an experimental feature.

BUG=#86

PR #303, opened 2021-07-20
Fix incorrect values in recorded_allocation on BE machines

### LINKED ISSUE #297: recorded_allocation parameters for quantized tensors are incorrect for BE machines
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow): No
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Ubuntu 18.04
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device: N/A
- TensorFlow installed from (source or binary): source
- TensorFlow version (use command below): 2.3.1
- Python version: 3.6.9
- Bazel version (if compiling from source): 3.4.1
- GCC/Compiler version (if compiling from source): Ubuntu 7.5.0-3ubuntu1~18.04
- CUDA/cuDNN version: N/A
- GPU model and memory: N/A

**Describe the current behavior**
The quantized tensor count and bytes are not recorded properly on s390x. 
While running `//tensorflow/lite/micro:recording_micro_allocator_test` on s390x,  test case fails with following error:
```
tensorflow/lite/micro/recording_micro_allocator_test_binary: FAIL - '~~~ALL TESTS PASSED~~~' not found in logs.
Testing TestRecordsTfLiteTensorArrayData
Testing TestRecordsTensorArrayQuantizationData
recorded_allocation.count == quantized_tensor_count * 2 failed at tensorflow/lite/micro/recording_micro_allocator_test.cc:110 (51 vs 24)
recorded_allocation.requested_bytes == expected_requested_bytes failed at tensorflow/lite/micro/recording_micro_allocator_test.cc:112 (1436 vs 752)
Testing TestRecordsNodeAndRegistrationArrayData
2/3 tests passed
```
On debugging, it seems that `RecordingSimpleMemoryAllocator::AllocateFromTail` function gets called twice the times it's being called on x86. `AllocateFromTail` function updates `used_bytes_`, `requested_bytes_` and `alloc_count_` which are used to update `recorded_allocation` parameters in `RecordingMicroAllocator::RecordAllocationUsage` function.
For x86,  `AllocateFromTail` is only called when extracting tensor data from serialized tensor buffers but on s390x, `FlatBufferVectorToTfLiteTypeArray` also calls `AllocateFromTail` to convert endianness of flatbuffer array in case of BE systems. This causes the data in `requested_bytes_` and `alloc_count_`  private variables of class `RecordingSimpleMemoryAllocator` to be incorrect.

One way to fix this could to be take a global variable, we can set it when calling `AllocatefromTail` to convert endianness. If this variable is set, then we do not update `recorded_allocation` parameters. Can you suggest a better way to approach this problem?

**Describe the expected behavior**
The `RecordingSimpleMemoryAllocator` should record `recorded_allocation` parameters correctly and the test case should pass.

**Standalone code to reproduce the issue**
```
bazel --host_jvm_args="-Xms1024m" --host_jvm_args="-Xmx2048m" test --host_javabase="@local_jdk//:jdk" --test_tag_filters=-gpu,-benchmark-test,-v1only,-no_oss,-oss_serial -k --test_timeout 300,450,1200,3600 --build_tests_only --test_output=errors --verbose_failures  -- //tensorflow/lite/micro:recording_micro_allocator_test
```

**Other info / logs** 
Code in `AllocateFromTail` that is recording the count:
```
uint8_t* RecordingSimpleMemoryAllocator::AllocateFromTail(size_t size,
                                                          size_t alignment) {
  const uint8_t* previous_tail = GetTail();
  uint8_t* result = SimpleMemoryAllocator::AllocateFromTail(size, alignment);
  if (result != nullptr) {
    used_bytes_ += previous_tail - GetTail();
    requested_bytes_ += size;
    alloc_count_++;
  }
  return result;
}
```
The code that is calling `AllocateFromTail` more than it should is in `FlatBufferVectorToTfLiteTypeArray` which being called from `FlatBufferVectorToTfLiteTypeArray`:
```
 if (FLATBUFFERS_LITTLEENDIAN) {
    // On little-endian machines, TfLite*Array happens to have the same memory
    // layout as flatbuffers:Vector<kFlatBufferVectorType>, so we can
    // reinterpret_cast the flatbuffer vector and avoid a copy and malloc.
    *result = const_cast<kTfLiteArrayType*>(
        reinterpret_cast<const kTfLiteArrayType*>(flatbuffer_array));
  } else {
    // Big-endian architecture can not use the same memory layout as
    // flatbuffers::Vector<kFlatBufferVectorType>. Allocate from the tail and
    // copy values from the flatbuffer into the newly allocated chunk.
    kTfLiteArrayType* array =
        reinterpret_cast<kTfLiteArrayType*>(allocator->AllocateFromTail(
            TfLiteIntArrayGetSizeInBytes(flatbuffer_array->Length()),
            alignof(kTfLiteArrayType)));
```

kun-lu20:
This PR addresses incorrect `count `and `requested_bytes` values in `recorded_allocation` when testing model allocations.

Test case `//tensorflow/lite/micro:recording_micro_allocator_test` fails on Big-Endian systems because `RecordingSimpleMemoryAllocator::AllocateFromTail()` would be called one more time and the `alloc_count_` and `requested_tail_bytes_` fields would be added repeatedly.

`FlatBufferVectorToTfLiteTypeArray()` function in `micro_allocated.cc` has been modified to fix this by invoking the base class version of `AllocateFromTail()` function (`allocator->SimpleMemoryAllocator::AllocateFromTail()`) on BE machines, rather than the derived class version (`allocator->AllocateFromTail()`) which might update the value of `alloc_count_` and `requested_tail_bytes_` fields besides allocating the space.

This PR will fix #297.

BUG=#297

### DIFF
--- a/tensorflow/lite/micro/micro_allocator.cc
+++ b/tensorflow/lite/micro/micro_allocator.cc
@@ -370,8 +370,8 @@ TfLiteStatus FlatBufferVectorToTfLiteTypeArray(
     // Big-endian architecture can not use the same memory layout as
     // flatbuffers::Vector<kFlatBufferVectorType>. Allocate from the tail and
     // copy values from the flatbuffer into the newly allocated chunk.
-    kTfLiteArrayType* array =
-        reinterpret_cast<kTfLiteArrayType*>(allocator->AllocateFromTail(
+    kTfLiteArrayType* array = reinterpret_cast<kTfLiteArrayType*>(
+        allocator->SimpleMemoryAllocator::AllocateFromTail(
             TfLiteIntArrayGetSizeInBytes(flatbuffer_array->size()),
             alignof(kTfLiteArrayType)));
     if (array == nullptr) {


kun-lu20:
@advaitjain Thanks for letting me move code changes in [tensorflow/tensorflow#50283](https://github.com/tensorflow/tensorflow/pull/50283) to here.

advaitjain:
@kun-lu20 please be sure to check the box that allows maintainers to make edits to the PR. The PR needs to be up to date with the main branch prior to being merged and I currently do not have the ability to do that.

advaitjain (approved):

advaitjain:
> @kun-lu20 please be sure to check the box that allows maintainers to make edits to the PR. The PR needs to be up to date with the main branch prior to being merged and I currently do not have the ability to do that.

I think you are encountering the same issue as https://github.com/tensorflow/tflite-micro/pull/227#issuecomment-880757254

If at all possible, we would recommend sending PRs from a user github account (instead of an organization account). Giving the TFLM maintainers the ability to update PRs and make small fixes will make the process easier for everyone.

kun-lu20:
@advaitjain Yes, it's the same issue. I've clicked "update branch" button and I think you could merge it now. Thanks!

advaitjain (approved):

kun-lu20:
Thanks @advaitjain !

PR #310, opened 2021-07-22
Fix esp32 ci build

### LINKED ISSUE #294: Fix esp32 ci build
I filed #292 to add a community build for esp32.  

When the ci script runs it fails.  Currently the error output can be seen here:
https://github.com/mocleiri/tflite-micro/actions/runs/1040366564

@vikramdattu you fixed it last time (https://github.com/tensorflow/tensorflow/pull/45334) could you take a look?

mocleiri:
BUG=Fixes #294

I was able to run this with some minor alterations on when the job ran on my fork:
https://github.com/mocleiri/tflite-micro/actions/runs/1054903715

Copy from xtensa ci scripts to run the test script inside of a docker
container.

We use the espressif/idf:release-v4.3 image.

Add a build that builds the full tensorflow-microlite.a library.

The person detection example is more complicated and not fixed here.

### DIFF
--- a/.github/workflows/esp32.yml
+++ b/.github/workflows/esp32.yml
@@ -23,4 +23,4 @@ jobs:
       - uses: actions/checkout@v2
       - name: Test
         run: |
-          tensorflow/lite/micro/tools/ci_build/test_esp32.sh
+          docker run --rm -v `pwd`:/opt/tflite-micro espressif/idf:release-v4.3 /opt/tflite-micro/tensorflow/lite/micro/tools/ci_build/test_esp32.sh
--- a/tensorflow/lite/micro/tools/ci_build/test_esp32.sh
+++ b/tensorflow/lite/micro/tools/ci_build/test_esp32.sh
@@ -26,34 +26,34 @@ pwd
 source tensorflow/lite/micro/tools/ci_build/helper_functions.sh
 
 TARGET=esp
+TARGET_ARCH=xtensa-esp32
 
-# setup esp-idf and toolchains
-echo "Checking out esp-idf..."
-readable_run git clone --recursive --single-branch --branch release/v4.2 https://github.com/espressif/esp-idf.git
-export IDF_PATH="${ROOT_DIR}"/esp-idf
-cd $IDF_PATH
-readable_run ./install.sh
-readable_run . ./export.sh
-cd "${ROOT_DIR}"
+readable_run make -f tensorflow/lite/micro/tools/make/Makefile TARGET=${TARGET} TARGET_ARCH=${TARGET_ARCH} third_party_downloads
 
 # clean all
 readable_run make -f tensorflow/lite/micro/tools/make/Makefile clean
 
+# validate esp32 build for libtensorflow-microlite.a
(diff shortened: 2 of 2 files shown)


.github/workflows/esp32.yml
```diff
@@ -23,4 +23,4 @@ jobs:
       - uses: actions/checkout@v2
       - name: Test
         run: |
-          tensorflow/lite/micro/tools/ci_build/test_esp32.sh
+          docker run --rm -v `pwd`:/opt/tflite-micro espressif/idf:release-v4.3 /opt/tflite-micro/tensorflow/lite/micro/tools/ci_build/test_esp32.sh
```
mocleiri, line 26: I copied this from the xtensa f1 workflow which seems to use a custom image assocated with the tflm bot.  Here because the espressif image is public it just works.  

The -v part is a volume mount that brings the current working directory (the root of the tflite-micro repo) into the image in the /opt/tflite-micro directory.

Then the arg is the specific file from within the image to run.

tensorflow/lite/micro/tools/ci_build/test_esp32.sh
```diff
@@ -26,34 +26,34 @@ pwd
 source tensorflow/lite/micro/tools/ci_build/helper_functions.sh
 
 TARGET=esp
+TARGET_ARCH=xtensa-esp32
```
mocleiri, line 29: Without specifying this the resultant generated example directory has xthe wrong name.  It doesn't detect the arch right and has x86 in it instead of what we want.

tensorflow/lite/micro/tools/ci_build/test_esp32.sh
```diff
@@ -26,34 +26,34 @@ pwd
...
 
-# setup esp-idf and toolchains
-echo "Checking out esp-idf..."
-readable_run git clone --recursive --single-branch --branch release/v4.2 https://github.com/espressif/esp-idf.git
-export IDF_PATH="${ROOT_DIR}"/esp-idf
-cd $IDF_PATH
-readable_run ./install.sh
-readable_run . ./export.sh
-cd "${ROOT_DIR}"
```
mocleiri, line 37: While this is removed out there is another one above on line 23 that puts us into the ROOT_DIR (/opt/tflite-micro) before the commands are run.

tensorflow/lite/micro/tools/ci_build/test_esp32.sh
```diff
@@ -26,34 +26,34 @@ pwd
...
+readable_run make -f tensorflow/lite/micro/tools/make/Makefile TARGET=${TARGET} TARGET_ARCH=${TARGET_ARCH} third_party_downloads
 
 # clean all
 readable_run make -f tensorflow/lite/micro/tools/make/Makefile clean
 
+# validate esp32 build for libtensorflow-microlite.a
+readable_run make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=${TARGET} TARGET_ARCH=${TARGET_ARCH} TARGET_TOOLCHAIN_PREFIX=xtensa-esp32-elf-
```
mocleiri, line 37: For esp32 a no argument invocation of the makefile results in a clean build of the static library.  There are build errors if you choose the 'build' target liike some of the other ci scripts.

advaitjain (approved):

PR #321, opened 2021-07-26
Fixes for RISC V build

petewarden:
It looks like the RISC V build process has been broken by some changes over the las few months. This fix at least gets binaries to the point of compiling and building, though the runtime tests haven't yet been tried.  Addresses issue #203.

BUG=#203.

### DIFF
--- a/tensorflow/lite/micro/tools/make/targets/mcu_riscv_makefile.inc
+++ b/tensorflow/lite/micro/tools/make/targets/mcu_riscv_makefile.inc
@@ -16,24 +16,9 @@ PLATFORM_FLAGS = \
   -fno-exceptions \
   -DTF_LITE_MCU_DEBUG_LOG \
   -DTF_LITE_USE_GLOBAL_CMATH_FUNCTIONS \
-  -fno-unwind-tables \
-  -ffunction-sections \
-  -fdata-sections \
   -funsigned-char \
-  -Wvla \
-  -Wall \
-  -Wextra \
-  -Wsign-compare \
-  -Wdouble-promotion \
-  -Wshadow \
-  -Wunused-variable \
-  -Wmissing-field-initializers \
-  -Wno-unused-parameter \
-  -Wno-write-strings \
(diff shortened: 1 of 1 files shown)


petewarden:
The diff is messy, but mostly because I removed the nested whitespace when the ifeq guard was deleted.

tensorflow/lite/micro/tools/make/targets/mcu_riscv_makefile.inc
```diff
@@ -1,69 +1,70 @@
...
-    -fno-use-cxa-atexit \
-    -Os
+$(eval $(call add_third_party_download,$(RISCV_TOOLCHAIN_URL),$(RISCV_TOOLCHAIN_MD5),riscv_toolchain,))
+$(eval $(call add_third_party_download,$(SIFIVE_FE310_LIB_URL),$(SIFIVE_FE310_LIB_MD5),sifive_fe310_lib,))
 
-  CXXFLAGS += $(PLATFORM_FLAGS) \
-    -fpermissive \
-    -fno-rtti \
-    --std=gnu++11
+PLATFORM_FLAGS = \
```
advaitjain, line 10: Many of the flags are now consolidated into the top-level makefile:
https://github.com/tensorflow/tflite-micro/blob/bfd9f923c7d6c1a998c6077cf46bdece46ddaf65/tensorflow/lite/micro/tools/make/Makefile#L130-L153

It might be worth shortening the riscv list to only be target-specific, for example here is the bluepill PLATFORM_FLAGS:
https://github.com/tensorflow/tflite-micro/blob/bfd9f923c7d6c1a998c6077cf46bdece46ddaf65/tensorflow/lite/micro/tools/make/targets/bluepill_makefile.inc#L25-L32

petewarden: Good idea, I've shrunk the number of flags down.

advaitjain (approved):

Issue #329, opened 2021-07-27
Incorrect convolution computation in a specific case on Xtensa HiFi4


lvadke:
Hi, I'm not sure if TFLM supports Xtensa HiFi4 cores, because the optimized kernels have preprocessing conditions like:
```c++
#if defined(HIFIMINI)
      ConvEvalHifiMini(ConvParamsQuantized(params, op_data.reference_op_data),
                       op_data.reference_op_data.per_channel_output_multiplier,
                       op_data.reference_op_data.per_channel_output_shift,
                       tflite::micro::GetTensorShape(input),
                       tflite::micro::GetTensorData<int8_t>(input),
                       tflite::micro::GetTensorShape(filter),
                       tflite::micro::GetTensorData<int8_t>(filter),
                       tflite::micro::GetTensorShape(bias),
                       tflite::micro::GetTensorData<int32_t>(bias),
                       tflite::micro::GetTensorShape(output),
                       tflite::micro::GetTensorData<int8_t>(output));
#elif defined(FUSION_F1) || defined(HIFI5)
      ConvEvalHifi(context, node, params, op_data, input, filter, bias, output);
#else
      return ConvReferenceEvalInt8(context, node);
#endif
```
But if it does (I used a small workaround), then it computes convolution incorrectly in one specific case. When I compare the outputs from HiFi4 and reference implementations, I see some differences:

![output](https://user-images.githubusercontent.com/52713197/127188787-39861d75-287c-4c74-a892-6688cce76538.png)

I created a unit test which reproduces the issue.

[conv_test.txt](https://github.com/tensorflow/tflite-micro/files/6887125/conv_test.txt)

On Xtensa Fusion F1 core the computation works fine.

advaitjain:
The hifimini implementation is not intended for use with Hifi4.

The naming was unclear previously but that has recently been improved. Please see this comment for the current way to build TFLM with Hifi4 optimized kernels:
https://github.com/tensorflow/tflite-micro/issues/296#issuecomment-886868989

lvadke:
Sorry, I was not clear which configuration I used, but I used the same for HiFi4 as for Fusion F1. HiFi Mini would not compile with the HiFi4 compiler configuration (different inline assembler). I stepped through the code and it uses the HiFi4 NN library.

nyadla-sys:
@cad-audio

advaitjain:
Thanks for the clarification. We'll look into this along with the nnlib authors (@cad-audio)

[...]

PR #330, opened 2021-07-27
Cadence HiFi 4 Neural Network Library v2.4.1:

cad-audio:
Updated the script to download HiFi 4 Neural Network Library v2.4.1.
Removed the hifi4 patch
Updates in v2.4.1 library:
+ Fixed the issue related to special case handling in xa_nn_dot_prod_16x16_asym8s.
  Moved the special cases handling for vec_length=8/32 to the start of the function.
+ Removed legacy condition checks for depthwise_conv kernels.
+ Fixed the non-bit-exact issue related to xa_nn_conv2d_std_per_chan_sym8sxasym8s.
  updated to select correct vec values in case of vec_align_val is 1 and

BUG=#329


cad-audio:
@advaitjain

advaitjain:
Does this update also address https://github.com/tensorflow/tflite-micro/issues/329?

advaitjain (approved):

cad-audio:
@advaitjain 
Yes, I have verified with the unit test case as well (which is attached in #329 )

advaitjain:
> @advaitjain
> Yes, I have verified with the unit test case as well (which is attached in #329 )

Thanks!

PR #337, opened 2021-07-29
Add support for c array generation during make and bazel builds

njeffrie:
Add a python script to generate cc and header files from tflite and bmp sources during the build. This allows tflite files and test inputs to be used directly and included in the repo, rather than using human-generated cc and header files.

BUG=http://b/196065539


petewarden (approved):

tensorflow/lite/micro/tools/make/make_generate_cc_arrays.py
```diff
@@ -0,0 +1,55 @@
...
+    args = parser.parse_args()
+    for input_file in args.input_files:
+        output_base_fname = os.path.join(args.output_dir,
+                                         input_file.split('.')[0])
+        if input_file.endswith('.tflite'):
+            output_base_fname = output_base_fname + '_model_data'
+        elif input_file.endswith('.bmp'):
+            output_base_fname = output_base_fname + '_image_data'
+        else:
+            raise ValueError('input file must be .tflite, .bmp')
```
advaitjain, line 40: It would be better if we can have a single python binary that can be used with bazel, make and as a standalone command line tool too.

If possible, we might just put everything into generate_cc_array.py

Else, we can have a convention like:
 * generate_cc_array.py
 * generate_cc_array_lib.py

http://yaqs/2372556976740630528

tensorflow/lite/micro/BUILD
```diff
@@ -477,6 +477,17 @@ cc_test(
...
+py_library(
+    name = "generate_cc_arrays",
+    srcs = ["tools/make/generate_cc_arrays.py"],
+)
+
+py_binary(
+    name = "bzl_generate_cc_arrays",
+    srcs = ["tools/make/bzl_generate_cc_arrays.py"],
+    deps = [":generate_cc_arrays"],
+)
```
advaitjain, line 489: should we put the BUILD file in tools/make? That's a more common convention.

Or possibly a new folder (tools/data_generation)? I'm wondering if we can have a directory that can serve as a place for additional tools as well (similar to the lite/tools/ directory in the TF repo

tensorflow/lite/micro/examples/person_detection/BUILD
```diff
@@ -16,13 +18,49 @@ cc_library(
...
+    name="generated_person_detect_model_cc",
+    src="person_detect.tflite",
+    out="person_detect_model_data.cc",
+)
+
+generate_cc_arrays(
+    name="generated_person_detect_model_hdr",
+    src="person_detect.tflite",
+    out="person_detect_model_data.h",
+)
```
advaitjain, line 55: since we use these in both benchmarks and examples, would this PR be a good point in time to centralize such artificats? maybe a top-level data directory?

tensorflow/lite/micro/tools/make/helper_functions.inc
```diff
@@ -475,13 +475,14 @@ endef
...
 #
 # Arguments are:
 # 1 - Name of target.
 # 2 - C/C++ source files
 # 3 - C/C++ header files
-# 4 - if "exclude", then the non-test target will be excluded from
+# 4 - Model sources and model test inputs in .tflite or .bmp format.
+# 5 - if "exclude", then the non-test target will be excluded from
 #     MICROLITE_BUILD_TARGETS. This exception is needed because not all the
 #     microlite_test targets (e.g. the examples) are buildable on all platforms.
```
advaitjain, line 487: I don't see anywhere in the code that we are actually using "exclude". Might be worth removing it completely. Can be a separate PR.

tensorflow/lite/micro/tools/make/helper_functions.inc
```diff
@@ -505,6 +506,8 @@ $$($(1)_BINARY): $$($(1)_LOCAL_OBJS) $$(MICROLITE_LIB_PATH)
 $(1): $$($(1)_BINARY)
 $(1)_bin: $$($(1)_BINARY).bin
 
+GENERATOR_INPUTS += $(4)
```
advaitjain, line 509: might be worth a comment that we only add to the list here and the actual generation happens later to prevent repeating the same generation multiple times for different build targets.

tensorflow/lite/micro/tools/make/make_generate_cc_arrays.py
```diff
@@ -0,0 +1,55 @@
...
+
+def main():
+    """Create cc sources with c arrays with data from each .tflite or .bmp."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument('output_dir', help='base directory for all outputs')
+    parser.add_argument('input_files',
+                        nargs='+',
+                        help='input bmp or tflite files to convert')
+    args = parser.parse_args()
+    for input_file in args.input_files:
```
advaitjain, line 32: if the same file is listed multiple times in args.input_list it will be generated multiple times as well, right? So MICROLITE_CC_SRCS can have the same path repeated multiple times?

tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -635,6 +636,13 @@ include $(MICRO_LITE_BENCHMARKS)
 # Load custom kernel tests.
 include $(MAKEFILE_DIR)/additional_tests.inc
 
+# Generate cc files and headers for all models and bitmaps in project Makefiles.
+# Add all generated cc sources to the build.
+MICROLITE_CC_SRCS += $(shell python3 $(MAKEFILE_DIR)/make_generate_cc_arrays.py $(GENERATED_SRCS_DIR) $(GENERATOR_INPUTS))
```
advaitjain, line 641: I think we want to avoid putting the generated files into MICROLITE_CC_SRCS.

I'd like to avoid having these be part of libtensorflow-microlite.a but rather as a separate libtflm-data.a or only be compiled with the individual binaries.

advaitjain (changes requested):

tensorflow/lite/micro/tools/make/helper_functions.inc
njeffrie: Ack - added generation here.

tensorflow/lite/micro/tools/make/helper_functions.inc
njeffrie: Removed it in this PR since we are already messing with arguments.

tensorflow/lite/micro/examples/person_detection/BUILD
njeffrie: Done.

tensorflow/lite/micro/BUILD
njeffrie: Moved build file and scripts to tensorflow/lite/micro/tools.

tensorflow/lite/micro/tools/make/Makefile
njeffrie: Done. Moved to microlite_test.

tensorflow/lite/micro/tools/make/make_generate_cc_arrays.py
njeffrie: Moved to single python file.

tensorflow/lite/micro/tools/make/make_generate_cc_arrays.py
njeffrie: Added dedup in this loop.

tensorflow/lite/micro/tools/make/make_generate_cc_arrays.py
njeffrie: Merged into single python file.

tensorflow/lite/micro/tools/BUILD
```diff
@@ -0,0 +1,10 @@
+package(
+    default_visibility = ["//visibility:public"],
+    features = ["-layering_check"],
```
advaitjain, line 3: nit: let's remove these two

advaitjain (approved):
one small nit comment.

tensorflow/lite/micro/tools/BUILD
njeffrie: Done.

advaitjain (approved):

advaitjain (approved):

PR #354, opened 2021-08-05
Add a new test case for conv operator (#329)

deqiangc:
This new test case is based on issue #329 to increase
coverage on optimized kernel's data precision for conv operator.
In this test, input, output and filter are all 8 bits and filter
tensor is of dimension 8x3x3x3 with different scales per output channel.

TESTED= local test with x86 and HiFi4.


tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -794,9 +794,11 @@ ifneq ($(findstring $(EXPLICITLY_SPECIFIED_TEST),$(MICROLITE_TEST_SRCS)),)
   MICROLITE_TEST_SRCS := $(filter-out $(EXPLICITLY_SPECIFIED_TEST), $(MICROLITE_TEST_SRCS))
   EXPLICITLY_SPECIFIED_TEST_SRCS := \
   $(EXPLICITLY_SPECIFIED_TEST) \
-  tensorflow/lite/micro/kernels/conv_test_common.cc
+  tensorflow/lite/micro/kernels/conv_test_common.cc \
```
njeffrie, line 797: Thoughts about putting all `*_test_data*` under lite/micro/kernels/testdata? I'm wondering if that might help avoid polluting the kernels/ directory once we add more separate test data files.

deqiangc: Sounds like a good suggestion to me.  We have a few *_generated_data* that seems to be for unit test as well and also under kernel folder.  If we decide have a lite/micro/kernels/testdata, we should probably move those as well? 

Another point that suggestion is welcome is the use of namespace to separate test data: this CL puts test data into tflite::testing::conv to avoid contaminate of common global test variables.  The way that our test is constructed seems to be built and run one by one with separate binary and there is no real danger of pollution; but I feel explicit different namespace may reduce mental load in code reading.

tensorflow/lite/micro/kernels/conv_test.cc
```diff
@@ -791,4 +792,114 @@ TF_LITE_MICRO_TEST(Int8Input32x1Filter32x32ShouldMatchGolden) {
                      kQuantizationTolerance));
 }
 
+// This test is created based on
+// https://github.com/tensorflow/tflite-micro/issues/329.
```
advaitjain, line 796: nit: remove the period to allow for the link to be clickable.

tensorflow/lite/micro/kernels/conv_test_data.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include "tensorflow/lite/c/common.h"
+
+namespace tflite {
+namespace testing {
+namespace conv {
+extern const int8_t kInput1x32x32x3[];
+extern const int8_t kFilter8x3x3x3[];
+extern const int32_t kBiasQuantized8[];
+extern const int8_t kGoldenOutput1x16x16x8[];
```
advaitjain, line 27: Since we will likely have even more such arrays, I think we will want to come up with some naming convention to prevent ODR violations.

For example, kBiasQuantized8 is likely too generic a name.

Any ideas?

tensorflow/lite/micro/tools/make/Makefile
advaitjain: +1 to a common micro/kernels/testdata and to moving the existing code around to match that.

On the namespace: TfLite and TFLM has historically used nested namespaces but for tflm we are moving towards a flat namespace: http://b/170230530

This does mean that we will need some convention for naming the arrays.

advaitjain (approved):

tensorflow/lite/micro/tools/make/ext_libs/arc_mli.inc
```diff
@@ -24,7 +24,7 @@ ifeq ($(filter no_arc_mli,$(ARC_TAGS)),)
 ALL_TAGS += arc_mli
 
 ifeq ($(BUILD_ARC_MLI),true)
-  MLI_LIB_DIR ?= arc_mli_$(basename $(TCF_FILE_NAME))
```
advaitjain, line 27: is this a bad merge?

tensorflow/lite/micro/tools/make/targets/arc/arc_common.inc
```diff
@@ -97,9 +97,13 @@ ifeq ($(ARC_TOOLCHAIN), mwdt)
...
 # the configuration bundled with MWDT (usually without .tcf extension) and that doesn't require copying.
 ifneq (,$(findstring .tcf,$(TCF_FILE)))
+ifneq ($(findstring microlite, $(MAKECMDGOALS)),)
+  TCF_FILE_NAME = $(TCF_FILE)
+else
   TCF_FILE_NAME = $(notdir $(TCF_FILE))
+endif
   ARC_TARGET_COPY_FILES += $(notdir $(TCF_FILE))!$(TCF_FILE)
-  MAKE_PROJECT_FILES += $(TCF_FILE_NAME)
+  MAKE_PROJECT_FILES += $(notdir $(TCF_FILE))
```
advaitjain, line 106: bad merge?

advaitjain (changes requested):

deqiangc:
Replace this PR request with PR #377  due to this PR request does not show diff properly.

PR #360, opened 2021-08-05
Fix example generation in wrong directory

vikramdattu:
BUG=wrong example generation directory

`GENDIR` variable specifically need to be defined after correct $TARGET_ARCH is defined.
Otherwise, generation directory gets changed to default TARGET_ARCH i.e., HOST_ARCH

Fixed this by moving these defaults to later stage.

### DIFF
--- a/tensorflow/lite/micro/tools/make/Makefile
+++ b/tensorflow/lite/micro/tools/make/Makefile
@@ -233,19 +233,6 @@ else ifeq ($(BUILD_TYPE), no_tf_lite_static_memory)
 	CCFLAGS := $(filter-out -DTF_LITE_STATIC_MEMORY, $(CCFLAGS))
 endif
 
-# This library is the main target for this makefile. It will contain a minimal
-# runtime that can be linked in to other programs.
-MICROLITE_LIB_NAME := libtensorflow-microlite.a
-
-# Where compiled objects are stored.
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -620,6 +607,19 @@ ALL_SRCS := \
...
+
+# Where compiled objects are stored.
+GENDIR := $(MAKEFILE_DIR)/gen/$(TARGET)_$(TARGET_ARCH)_$(BUILD_TYPE)/
+CORE_OBJDIR := $(GENDIR)obj/core/
+KERNEL_OBJDIR := $(GENDIR)obj/kernels/
+THIRD_PARTY_OBJDIR := $(GENDIR)obj/third_party/
+BINDIR := $(GENDIR)bin/
+LIBDIR := $(GENDIR)lib/
+PRJDIR := $(GENDIR)prj/
+
```
vikramdattu, line 622: @advaitjain moving this makes sure that the `$(TARGET)_makefile.inc` included and TARGET_ARCH variable is set before this.

advaitjain:
I think https://github.com/tensorflow/tflite-micro/pull/371 has the same fix. I'm going to close this PR since #371 is merged.

Issue #332, opened 2021-07-27
Zephyr sample build broken


laurenmurphyx64:
**System information**

- Have I written custom code (as opposed to using a stock example script provided in TensorFlow): No
- OS Platform and Distribution: Linux Ubuntu 18.04
- TensorFlow installed from (source or binary): source
- GCC/Compiler version: Ubuntu 7.5.0-3ubuntu1~18.04

**Describe the current behavior**

After the Zephyr build system compiles all of the TFLM library files itself based on the `CMakeLists.txt` generated by TF's build system, it breaks during linking with undefined references:

```
...
[ 94%] Linking CXX executable zephyr_prebuilt.elf
...
~/zephyr-sdk-0.12.3/riscv64-zephyr-elf/bin/../lib/gcc/riscv64-zephyr-elf/10.2.0/../../../../riscv64-zephyr-elf/bin/ld.bfd: ../app/libapp.a(all_ops_resolver.cc.obj): in function `tflite::MicroMutableOpResolver<128u>::AddArgMin()':
~/tflite-micro/tensorflow/lite/micro/tools/make/gen/zephyr_vexriscv_x86_64_default/prj/hello_world/cmake/./tensorflow/lite/micro/micro_mutable_op_resolver.h:138: undefined reference to `tflite::Register_AVERAGE_POOL_2D()'
...
(74 other undefined references)
```

`libtensorflow-microlite.a`, which TF's build system builds but the Zephyr build system does not use, compiles and links successfully. 

Comparing the files included in the `libtensorflow-microlite.a` (from the log) to the files listed in the generated `tensorflow/lite/micro/tools/make/gen/zephyr_vexriscv_x86_64_default/prj/hello_world/cmake/third_party/CMakeLists.txt`, my conclusion is that Zephyr is missing the files of `MICROLITE_CC_KERNEL_SRCS` (listed at `tensorflow/lite/micro/tools/make/Makefile:349`):

```
MICROLITE_CC_KERNEL_SRCS := \
tensorflow/lite/micro/kernels/activations.cc \
tensorflow/lite/micro/kernels/activations_common.cc \
tensorflow/lite/micro/kernels/add.cc \
...
```

**Describe the expected behavior**

The Zephyr build of the samples should be able to link the executable without undefined references.

**[Contributing](https://www.tensorflow.org/community/contribute)**

- Do you want to contribute a PR? (yes/no): Yes
- Briefly describe your candidate solution(if contributing):

[...]

PR #363, opened 2021-08-05
Fix wrong tflm sources

### LINKED ISSUE #361: Makefile is broken?
The new Makefile for generating example now picks all the sources instead of nit-picking.

Following lines seem to be culprit:

```
MICROLITE_CC_BASE_SRCS := \
$(wildcard tensorflow/lite/micro/*.cc) \
$(wildcard tensorflow/lite/micro/memory_planner/*.cc) \
$(TFL_CC_SRCS)

...

MICROLITE_CC_SRCS := $(filter-out $(MICROLITE_TEST_SRCS), $(MICROLITE_CC_BASE_SRCS))
MICROLITE_CC_SRCS := $(filter-out $(MICROLITE_BENCHMARK_SRCS), $(MICROLITE_CC_SRCS))
```

This generates wrong examples, e.g., in case of `esp`, all of these files get added to CMakeList and hence generated example fails to compile.

vikramdattu:
Fixes #361 

#360

### DIFF
--- a/tensorflow/lite/micro/tools/make/Makefile
+++ b/tensorflow/lite/micro/tools/make/Makefile
@@ -426,7 +426,14 @@ $(shell find tensorflow/lite/ -type d \( -path tensorflow/lite/experimental -o -
 MICROLITE_CC_BASE_SRCS := \
 $(wildcard tensorflow/lite/micro/*.cc) \
 $(wildcard tensorflow/lite/micro/memory_planner/*.cc) \
-$(TFL_CC_SRCS)
+tensorflow/lite/c/common.c \
+tensorflow/lite/core/api/error_reporter.cc \
+tensorflow/lite/core/api/flatbuffer_conversions.cc \
+tensorflow/lite/core/api/op_resolver.cc \
+tensorflow/lite/core/api/tensor_utils.cc \
+tensorflow/lite/kernels/internal/quantization_util.cc \
+tensorflow/lite/kernels/kernel_util.cc \
+tensorflow/lite/schema/schema_utils.cc
 
 MICROLITE_CC_HDRS := \
 $(wildcard tensorflow/lite/micro/*.h) \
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -439,7 +426,14 @@ $(shell find tensorflow/lite/ -type d \( -path tensorflow/lite/experimental -o -
 MICROLITE_CC_BASE_SRCS := \
 $(wildcard tensorflow/lite/micro/*.cc) \
 $(wildcard tensorflow/lite/micro/memory_planner/*.cc) \
-$(TFL_CC_SRCS)
```
advaitjain, line 442: we need to keep this variable around. Its unclear to me why this would cause trouble.

tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -620,6 +614,19 @@ ALL_SRCS := \
...
+MICROLITE_LIB_NAME := libtensorflow-microlite.a
+
+# Where compiled objects are stored.
+GENDIR := $(MAKEFILE_DIR)/gen/$(TARGET)_$(TARGET_ARCH)_$(BUILD_TYPE)/
+CORE_OBJDIR := $(GENDIR)obj/core/
+KERNEL_OBJDIR := $(GENDIR)obj/kernels/
+THIRD_PARTY_OBJDIR := $(GENDIR)obj/third_party/
+BINDIR := $(GENDIR)bin/
+LIBDIR := $(GENDIR)lib/
+PRJDIR := $(GENDIR)prj/
```
advaitjain, line 628: why did this need to be moved?

advaitjain (commented):
I don't fully understand the problem and why this change is fixing it.

What I do want to mention is that we are going to be removing project generation support from the TFLM makefiles in the near future. I would strongly recommend switching over to the create_tflm_tree.py script.

See more details in https://github.com/tensorflow/tflite-micro/issues/332#issuecomment-888546457

njeffrie:
I believe https://github.com/tensorflow/tflite-micro/pull/367 fixes this issue.

PR #370, opened 2021-08-06
Fixed person detection example for ESP platform

vikramdattu:
BUG=Broken person_detection example for ESP platform

Following changes to fix the example:

* Use int8 model
* Avoid build issues with target rule: Changed the rule from person_detection_int8 to person_detection for esp
* Convert uint8 camera buffer to int8

### DIFF
--- a/tensorflow/lite/micro/examples/person_detection/Makefile.inc
+++ b/tensorflow/lite/micro/examples/person_detection/Makefile.inc
@@ -77,8 +77,13 @@ $(eval $(call microlite_test,detection_responder_test_int8,\
 $(DETECTION_RESPONDER_TEST_SRCS),$(DETECTION_RESPONDER_TEST_HDRS)))
 
 # Builds a standalone object recognition binary.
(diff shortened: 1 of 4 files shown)


vikramdattu:
@advaitjain related MR on tensorflow repo: https://github.com/tensorflow/tensorflow/pull/47063

advaitjain (changes requested):
Apologies for the delay here.

With https://github.com/tensorflow/tflite-micro/pull/337, we are now putting the models into the repo. For example the person_detect model is at https://github.com/tensorflow/tflite-micro/tree/main/tensorflow/lite/micro/models

It would be good to have the ESP integration also make use of this model.

Also, just wanted to flag https://github.com/tensorflow/tflite-micro/issues/390#issuecomment-899871900

If you can confirm that specifying TARGET_ARCH as an explicit makefile command line option will work for you, that would be great.

advaitjain:
Closing old PR. Let's reopen if needed.

PR #385, opened 2021-08-09
Add special TARGET_ARCH=project_generation to cortex_m_generic.

advaitjain:
* This enables use to use the create_tflm_tree.py script correctly for all the examples.

 * The special behavior for `TARGET=cortex_m_generic TARGET_ARCH=project_generation` is to not exclude the micro_speech example (and its dependencies) is the list of sources.

 * We do not want to have micro_speech dependencies as part of a non-project generation build because it means that libtensorflow-microlite.a will have kissfft symbols which can lead to symbol collisions.

Also, improved the CI for project generation with cortex_m_generic, which needed some of the TF_LITE_REMOVE_VIRTUAL_DELETE calls to be moved from private to public sections.

BUG=http://b/193823889

### DIFF
--- a/tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
+++ b/tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
@@ -26,7 +26,7 @@ EXAMPLES="-e hello_world -e magic_wand -e micro_speech -e person_detection"
 
 TEST_OUTPUT_DIR=$(mktemp -d)
 
-#readable_run \
+readable_run \
   python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py \
   ${TEST_OUTPUT_DIR} \
   ${EXAMPLES}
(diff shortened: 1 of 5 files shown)


tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -22,15 +22,14 @@ cd "${ROOT_DIR}"
 
 source tensorflow/lite/micro/tools/ci_build/helper_functions.sh
 
+EXAMPLES="-e hello_world -e magic_wand -e micro_speech -e person_detection"
+
 TEST_OUTPUT_DIR=$(mktemp -d)
 
-readable_run \
+#readable_run \
```
petewarden, line 29: Just confirming this comment is intended? It's a bit surprising to have a comment with a \ continuation.

advaitjain: Fixed. This was left behind from my testing.

petewarden (approved):

PR #386, opened 2021-08-09
Fix a couple of issues with the BUILD file caught with the Google buildifier check.

rockyrhodes:
BUG=cleanup


deqiangc (dismissed):
LGTM

advaitjain (commented):
We actually want to disable layering_check because getting to parity between internal and open-source checks is tricky (http://b/177257332).

Disabling the checks is currently the easiest way to avoid build errors due to this discrepancy.

advaitjain (approved):

deqiangc (approved):

PR #388, opened 2021-08-10
Refactor Makefile: split the kernel test into its own Makefile.inc.

deqiangc:
After this change, only non-test kernel files are in Makefile.

BUG=http://b/195938386


njeffrie (dismissed):

tensorflow/lite/micro/kernels/Makefile.inc
```diff
@@ -0,0 +1,91 @@
+# This file includes kernel test targets only.
+
```
advaitjain, line 2: let's add a copyright notice. I'm surprised that the test_code_style did not catch this.

advaitjain (changes requested):
one nit about copyright notice. Other than that it looks good to me.

tensorflow/lite/micro/kernels/Makefile.inc
advaitjain: created http://b/196073950 to update the CI to catch this in the future. For now, let's add a copyright here and merge this PR.

https://github.com/tensorflow/tflite-micro/blob/8e86e1c29cec3fc115e481ab4cf946ab0e769d4d/tensorflow/lite/micro/tools/ci_build/test_code_style.sh#L50-L57

deqiangc: Done with adding copyright notice.

advaitjain (approved):

PR #393, opened 2021-08-11
Delete (most of the) Arduino-specific code.

advaitjain:
Moving forward, TFLM support for Arduino is going to be maintained in a stand-alone repository at: https://github.com/tensorflow/tflite-micro-arduino-examples

BUG=http://b/193823889


njeffrie (approved):
I will defer to Pete since he's much more familiar with the Arduino code, but it seems clear all this will have to go sooner or later.

My only concern is whether this breaks our Arduino library release. If not (or if the Arduino repo is ready to create releases) then this LGTM.

petewarden (approved):

PR #405, opened 2021-08-13
Only perform a build for the vision_p6 PR presubmit checks (do not run the tests)

advaitjain:
The Xtensa VisionP6 make test command is very slow: http://b/196459172

The primary reason for this is that the person_detect_int8 test takes a very long time to complete for Vision P6 because of the lack of any optimizations.

The workaround in this PR is to only do a build as part of the checks prior to submitting a PR. A better solution would be to have a way to explicitly exclude certain tests from running when we do a `make test`.

http://b/196583864 tracks the work needed for future improvements.

BUG=http://b/196459172


.github/workflows/xtensa.yml
```diff
@@ -55,22 +55,36 @@ jobs:
           echo ${{ secrets.TFLM_BOT_PACKAGE_READ_TOKEN }} | docker login ghcr.io -u tflm-bot --password-stdin
           docker run --rm -v `pwd`:/opt/tflite-micro ghcr.io/tflm-bot/xtensa:0.1 /opt/tflite-micro/tensorflow/lite/micro/tools/ci_build/test_xtensa_hifi5.sh
 
+  vision_p6_presubmit:
+    runs-on: ubuntu-latest
+
+    if: |
+      (github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'ci:run'))
+
+    name: Vision P6 (presubmit)
```
deqiangc, line 64: nit: name: Build Vision P6 (presubmit)

.github/workflows/xtensa.yml
```diff
@@ -55,22 +55,36 @@ jobs:
           echo ${{ secrets.TFLM_BOT_PACKAGE_READ_TOKEN }} | docker login ghcr.io -u tflm-bot --password-stdin
           docker run --rm -v `pwd`:/opt/tflite-micro ghcr.io/tflm-bot/xtensa:0.1 /opt/tflite-micro/tensorflow/lite/micro/tools/ci_build/test_xtensa_hifi5.sh
 
+  vision_p6_presubmit:
+    runs-on: ubuntu-latest
+
+    if: |
+      (github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'ci:run'))
```
deqiangc, line 62: nit: Would github.event_name == 'workflow_dispatch' still be useful here?  just in case someone trying to manually trigger workflow for xtensa only?

deqiangc (dismissed):
LGTM in general. some minor comments.

.github/workflows/xtensa.yml
advaitjain: done

.github/workflows/xtensa.yml
advaitjain: done

deqiangc (approved):

PR #410, opened 2021-08-13
Have make fail if C++ array generation fails

advaitjain:
Prior to this change, the generate_cc_arrays.py script could fail (for example because PIL not being installed) but the error would be silently ignored. Instead, the build would fail much later because of missing header files and the root cause would be hard to debug.

Follow-up PR will also use this error handling approach for other places where we call into Python and bash scripts from the make.

Manually tested by introducing an error in the Python script and confirming that the make command fails:
```
make -f tensorflow/lite/micro/tools/make/Makefile build
find: ‘../google/’: No such file or directory
tensorflow/lite/micro/tools/make/downloads/flatbuffers already exists, skipping the download.
tensorflow/lite/micro/tools/make/downloads/pigweed already exists, skipping the download.
tensorflow/lite/micro/tools/make/downloads/person_model_int8 already exists, skipping the download.
Traceback (most recent call last):
  File "/home/advaitjain/github/tflite-micro/tensorflow/lite/micro/tools/generate_cc_arrays.py", line 22, in <module>
    from PIL import Ima
ImportError: cannot import name 'Ima' from 'PIL' (/usr/lib/python3/dist-packages/PIL/__init__.py)
tensorflow/lite/micro/examples/person_detection/Makefile.inc:58: *** Something went wrong: .  Stop.
```

BUG=http://b/196630709

### DIFF
--- a/tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
+++ b/tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
@@ -49,9 +49,10 @@ popd > /dev/null
 
 rm -rf ${TEST_OUTPUT_DIR}
 
-# Check that we can export a TFLM tree with additional makefile options.
-TEST_OUTPUT_DIR_CMSIS=$(mktemp -d)
+# Remove existing state prior to testing project generation for cortex-m target.
+make -f tensorflow/lite/micro/tools/make/Makefile clean clean_downloads
 
+TEST_OUTPUT_DIR_CMSIS=$(mktemp -d)
 
 readable_run \
   python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py \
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -49,9 +49,10 @@ popd > /dev/null
 
 rm -rf ${TEST_OUTPUT_DIR}
 
-# Check that we can export a TFLM tree with additional makefile options.
-TEST_OUTPUT_DIR_CMSIS=$(mktemp -d)
+# Remove existing state prior to testing project generation for cortex-m target.
```
deqiangc, line 52: nit: is this only for cortex-m?

deqiangc (approved):

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
advaitjain: for now, yes. We are not testing any other target platform with the project generation script.

PR #411, opened 2021-08-13
Update the download script for TFLite Micro person detection example

MeghnaNatraj:
This change makes the script a lot easier to read and run  (uses `download_and_convert_data.py` instead of `download_mscoco.sh` and `build_visualwakewords_data.py` to download the dataset).

The output of each step is exactly the same, except few files have been renamed (instructions have been updated accordingly).

BUG=http://b/153503072


advaitjain (approved):
Thanks for this change! Is there a bug that we can link to? The CI checks for BUG=<bug number> in the PR description.

MeghnaNatraj:
I've updated the description - is that sufficient? (internal link to bug: b/153503072)

advaitjain:
> I've updated the description - is that sufficient? (internal link to bug: b/153503072)

yes, that works.

PR #413, opened 2021-08-13
Add a test for testing binary size comparison. 

deqiangc:
This change is on top of PR #254. 

It creates a new test that can be changed to cause test binary change.
It adds a workflow that can be triggered by adding label ci:test in a PR and auto addes a commit to reflect that PR's binary size change.

This is to enable more test on the mechanism to auto report binary size change in a PR.


deqiangc:
Close this PR. We will first merge PR #254 first and then add changes on top of it.

PR #415, opened 2021-08-14
Add Pillow to Dockerfile.micro

advaitjain:
I have already updated the docker container used for the style check -- this was needed to get the presubmit checks for #410 to pass.

https://github.com/tensorflow/tflite-micro/blob/f9e7d495df13951ac589f279d703d5f00402d5a1/.github/workflows/ci.yml#L75-L76

I did take a shortcut and directly pushed an updated image to `ghcr.io/tflm-bot/tflm-ci:latest` instead of following the more traceable approach of creating a new versioned tag and updating ci.yml

This was done to quickly allow me to merge #410.

BUG=http://b/196065539


ci/Dockerfile.micro
```diff
@@ -31,7 +31,7 @@ RUN ln -s /usr/bin/clang-format-12 /usr/bin/clang-format
 # integration.
 RUN pip install yapf
 
-RUN pip install six
+RUN pip install six Pillow
```
deqiangc, line 34: nit: line 32 and 37 both does pip with one package on one line. Just wondering whether it makes sense to have one package at one line as a rule of thumb.  This is really NIT and totally up to you.

deqiangc (approved):

njeffrie (approved):

ci/Dockerfile.micro
advaitjain: done. also added a comment about why Pillow was added

njeffrie (approved):

PR #417, opened 2021-08-16
Create a binary size test.

deqiangc:
Also create a script to generate size report of a such binary.
Modify the size_comp.py to be more generic (not limited to xtensa).

The next PR will update the size_check.yml so that we can experiment 
a workflow that  creates a commit that keeps tracks of the binary size 
of a given benchmark target for each PR.

Binary_size_test.cc in tensorflow/lite/micro/tools/ci_build/binary_size_test 
is created to generate an image with different data section by changing kSize.  
They will be removed after the workflow is verified. 

test_size.sh current build the above binary_size_test target so that we can experiment
this workflow to ensure it works properly and does not create all around merge conflict.
test_size.sh will be changed to build a real benchmark target once the workflow is verified to work.

BUG=http://b/196637015


tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -516,16 +516,6 @@ ifneq ($(OPTIMIZED_KERNEL_DIR),)
...
-
-  # The first ifneq is needed to be compatible with make versions prior to 4.2
-  # which do not support .SHELLSTATUS. While make 4.2 was released in 2016,
-  # Ubuntu 18.04 only has version 4.1
-  ifneq ($(.SHELLSTATUS),)
-    ifneq ($(.SHELLSTATUS),0)
-      $(error Error with specialize_files.py $(MICROLITE_CC_KERNEL_SRCS))
-    endif
-  endif
-
```
advaitjain, line 528: is this intentional?

advaitjain (changes requested):

tensorflow/lite/micro/tools/make/Makefile
deqiangc: oh. No.  Thanks for catching this!  This is a mistake by copy.

Fixed now.

tensorflow/lite/micro/tools/ci_build/binary_size_test/binary_size_test.cc
```diff
@@ -0,0 +1,38 @@
...
+TF_LITE_MICRO_TESTS_BEGIN
+
+TF_LITE_MICRO_TEST(BinarySizeChangeBykSize) {
+  // Just some code to create a binary and keep data section.
+  for (int i = 0; i < kSize; i++) {
+    random_array[i] = i + 1;
+  }
+
+  for (int i = 0; i < kSize; i++) {
+    TF_LITE_MICRO_EXPECT_EQ(random_array[i], i + 1);
```
njeffrie, line 34: Won't this test fail for values of i > 7? They first 8 elements are initialized but kSize is 64.

tensorflow/lite/micro/tools/ci_build/test_size.sh
```diff
@@ -0,0 +1,49 @@
...
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+ROOT_DIR=${SCRIPT_DIR}/../../../../..
+cd "${ROOT_DIR}"
+
+
+SIZEFILE_DIR=${ROOT_DIR}/ci
+MAKEFILE_DIR=${ROOT_DIR}/tensorflow/lite/micro/tools/make
+BENCHMARK_TARGET=binary_size_test
```
njeffrie, line 28: Nit: this isn't really a benchmark right? s/BENCHMARK/TEST?

tensorflow/lite/micro/tools/ci_build/test_size.sh
```diff
@@ -0,0 +1,49 @@
...
+# Build for x86.
+TARGET=linux
+TARGET_ARCH=x86_64
+readable_run make -f tensorflow/lite/micro/tools/make/Makefile clean
+readable_run make -j8 -f tensorflow/lite/micro/tools/make/Makefile build BUILD_TYPE=${BUILD_TYPE} TARGET=${TARGET} TARGET_ARCH=${TARGET_ARCH} ${BENCHMARK_TARGET}
+
+# Capture size of the test binary.
+GENDIR=${MAKEFILE_DIR}/gen/${TARGET}_${TARGET_ARCH}_${BUILD_TYPE}/
+BINDIR=${GENDIR}/bin/
+size ${BINDIR}/${BENCHMARK_TARGET} > ${SIZEFILE_DIR}/size_log.txt
```
njeffrie, line 48: I'm curious, is the size checked anywhere? Perhaps I'm missing some context on the overall purpose of this test.

njeffrie (commented):
Some random comments / questions. I'm not sure I have the full context, so take comments with a grain of salt.

advaitjain (approved):

advaitjain:
Maybe worth another comment in the PR description that this change is for prototyping purposes only and will be heavily modified prior to turning on for build targets that we care about, to address the questions that Nat has.

tensorflow/lite/micro/tools/ci_build/binary_size_test/binary_size_test.cc
deqiangc: I think line #22 mislead: the only reason that I initialized the global array there is to put them in data section rather than bss section.  Because the uninitialized global variable goes to bss section and I suspect that due to that, changing kSize is probably not going to lead to binary size to change, which is the goal.

deqiangc: so the test will not fail, b/c line 28 to line 30 re-populate the array up to kSize. 

Sorry for the confusion, I will add a comment in line #22.

tensorflow/lite/micro/tools/ci_build/test_size.sh
deqiangc: You are right that this right now just points to a test binary. 
The intention is that this will become the benchmark target when we have a defined benchmark target for size.  

I will add more comments in PR description.

tensorflow/lite/micro/tools/ci_build/test_size.sh
deqiangc: You are correct that this is not used in this CL.  The next CL will add a size_check workflow, which will takes what is output from here and create a commit to store the size history.

Right now, we need to do some test to ensure this mechanism even works without leading to merge conflict all around.

tensorflow/lite/micro/tools/ci_build/binary_size_test/binary_size_test.cc
deqiangc: I will add comment in the code at the next PR. But I will add more comments in the PR description.

tensorflow/lite/micro/tools/ci_build/test_size.sh
njeffrie: Never mind, the PR description answers this quesion

tensorflow/lite/micro/tools/ci_build/binary_size_test/binary_size_test.cc
njeffrie: Ahh got it, thanks.

tensorflow/lite/micro/tools/ci_build/test_size.sh
njeffrie: Ack.

njeffrie (approved):

advaitjain (approved):

Issue #325, opened 2021-07-26
Hifi5 optimzations for the TFLM kernels


advaitjain:
This is an umbrella issue to track the work needed to add in optimized kernels for Hifi5.

For the most part, these optimizations are wrappers between the TFLM kernels and https://github.com/foss-xtensa/nnlib-hifi5

advaitjain:
Closing. We will have more granular issues as needed.

PR #424, opened 2021-08-17
Cadence HiFi5 NN Library: Softmax operator.

cad-audio:
Integrating int8 input and int8 output optimized variant of softmax for Hifi4 and Hifi5.

The difference in the binary size for kernel_softmax_test as measured with:
```
make -f tensorflow/lite/micro/tools/make/Makefile TARGET=xtensa TARGET_ARCH=hifi4 XTENSA_CORE=F1_190305_swupgrade -j8 test_kernel_softmax_test
xt-size tensorflow/lite/micro/tools/make/gen/xtensa_hifi4_default/bin/kernel_softmax_test
```

suggests that the memory footprint of the optimized implementation is the same as the footprint of the reference implementation.
|       |  text   | data    |   bss   |    dec   |
| ----- | ------- | ------- | ------- | -------- |
| with this PR |  183668 | 30612 | 13064 | 227344 | 
| w/o this PR |  183796 | 30804 | 13064 | 227664 |

Note that the impact on arena usage has not been measured.

As expected, the keyword_benchmark size is unchanged.

BUG=#325


cad-audio:
@advaitjain

advaitjain (approved):

Issue #425, opened 2021-08-17
PPD kernel is not using stable sort


mansnils:
The detection_postprocess (PPD) kernel in TFL has partially changed to stable sort. In order for TFLM to be bit-exact to TFL it should also do this.
Some more context: https://github.com/tensorflow/tensorflow/issues/47158

PR #426, opened 2021-08-17
Replace std::partial_sort with merge_sort in PPD

mansnils:
TFL has changed to std::stable_sort but TFLM can not use it because it
is using the heap. Instead TFLM use merge sort.

BUG=Fixes https://github.com/tensorflow/tflite-micro/issues/425


tensorflow/lite/micro/kernels/detection_postprocess.cc
```diff
@@ -350,6 +350,57 @@ void DecreasingPartialArgSort(const float* values, int num_values,
       [&values](const int i, const int j) { return values[i] > values[j]; });
 }
 
+template <typename Compare>
+void insertion_sort(int* start, int* end, Compare compare) {
```
advaitjain, line 354: let's update all the function names to be CamelCase.

InsertionSort, TopDownMerge ...

tensorflow/lite/micro/kernels/detection_postprocess.cc
```diff
@@ -350,6 +350,57 @@ void DecreasingPartialArgSort(const float* values, int num_values,
...
+  for (int* i = start; i != end; ++i) {
+    std::rotate(std::upper_bound(start, i, *i, compare), i, i + 1);
+  }
+}
+
+template <typename Compare>
```
advaitjain, line 360: do these functions need to be templated? If it is not required right now, then I would prefer to not have it.

advaitjain (changes requested):
It would be good to have a test that checks the sorting that is implemented with this PR.

tensorflow/lite/micro/kernels/detection_postprocess.cc
```diff
@@ -350,6 +350,57 @@ void DecreasingPartialArgSort(const float* values, int num_values,
...
+
+  if (num_values < threshold) {
+    insertion_sort(values, values + num_values, compare);
+    return;
+  }
+
+  const int half_num_values = num_values / 2;
+
+  merge_sort(values, scratch, half_num_values, compare);
+  merge_sort(values + half_num_values, scratch, num_values - half_num_values,
```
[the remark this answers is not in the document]
petewarden: After taking a look, I think this seems reasonable. The recursion is less than log2(N) deep on the size of the array, with nothing but a few function arguments on the stack (no local array variables for example), so I think it should be okay even with just a couple of kilobytes of stack.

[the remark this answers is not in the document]
mansnils: Ok

tensorflow/lite/micro/kernels/detection_postprocess.cc
mansnils: Will do

tensorflow/lite/micro/kernels/detection_postprocess.cc
mansnils: It is needed to pass the compare function object around, which is used in a few places and it in that way the lambda expression can be defined in one place, like it was before. So in that sense I prefer to have them... ok?

tensorflow/lite/micro/kernels/detection_postprocess.cc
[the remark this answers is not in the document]
mansnils: Done

tensorflow/lite/micro/kernels/detection_postprocess.cc
mansnils: Done

tensorflow/lite/micro/kernels/detection_postprocess.cc
advaitjain: sounds ok to me.

tensorflow/lite/micro/kernels/detection_postprocess_test.cc
```diff
@@ -460,4 +460,57 @@ TF_LITE_MICRO_TEST(
...
+      0.0, 0.0,  0.0, 0.0, 1.0,  // box #6
+  };
+
+  // class scores - two classes without background
+  const float kInputData2[] = {0.015625, 0.007812, 0.003906, 0.015625,
+                               0.015625, 0.007812, 0.019531, 0.019531,
+                               0.007812, 0.003906, 0.003906, 0.003906};
+
+  constexpr int kInputElements1 =
+      1 * 6 * 5;  // kInputShape1[1] * kInputShape1[2] * kInputShape1[3];
```
advaitjain, line 484: nit: I think that if you make kInputShape a constexpr, then you could initialize kInputElements1 with kInpuShape

advaitjain (dismissed):
There seem to be some conflicts that need to be resolved. Also, feel free to apply the `ci:run` label to run all the checks that need to pass prior to the PR being merged.

tensorflow/lite/micro/kernels/detection_postprocess_test.cc
mansnils: done, I realized the uint8 unit tests had been removed

advaitjain (approved):

tensorflow/lite/micro/kernels/detection_postprocess_test.cc
advaitjain: uint8 removal for detection_postprocess was likely an oversight on our part. Can you confirm that the CMSIS-NN team needs continued support for this? https://github.com/tensorflow/tflite-micro/issues/457

PR #429, opened 2021-08-17
Generate cc and header files for micro_speech.

njeffrie:
Generate the model and audio inputs. Add wav file generation in generate_cc_arrays.py.

BUG=196065539


tensorflow/lite/micro/tools/generate_cc_arrays.py
```diff
@@ -70,16 +70,26 @@ def generate_array(input_fname):
     for byte in image_bytes:
       out_string += hex(byte) + ","
     return [len(image_bytes), out_string]
+  elif input_fname.endswith('.wav'):
+    wav = wave.open(input_fname, mode='r')
```
advaitjain, line 74: super nit: wav --> wav_file or something more descriptive. I letter difference between module name and variable name makes me nervous.

advaitjain (approved):
very cool!

We'll need to update the Xtensa docker container to add Wave as well, right?

tensorflow/lite/micro/tools/generate_cc_arrays.py
njeffrie: Good call. Done.

njeffrie:
I added pip3 install Wave wherever we have pip3 install Pillow, and I don't see any pip commands in xtensa. Perhaps I'm looking in the wrong place, however.

advaitjain (approved):

PR #422, opened 2021-08-17
Move make variables in Makefile

mansnils:
Moving GENDIR and other make variables up before target makefiles are
included, as cortex_m_corstone_300_makefile.inc is using the variables.

Fix for: https://github.com/tensorflow/tflite-micro/issues/390

Issue #407, opened 2021-08-13
Move hardware / platform support to repositories that are external to tflite-micro


advaitjain:
As an extension of https://github.com/tensorflow/tensorflow/issues/47413, we would like to move hardware platform support and platform-specific customization of the examples out of the tflite-micro repository and into platform specific repositories.

The goal with such a transition is:
 * Give more control over the TFLM integration to the maintainers of the specific platform, including set up of the IDE / build system etc.
 * Allow platform maintainers to decide at what cadence they would like to pull in updates from TFLM, make releases ...
 * Allow platform maintainers and easier way to customize and document the TFLM examples for their specific target.
 * Reduce the amount of code in the TFLM repository that can not be tested or maintained by the TFLM maintainers.
 * Simplify the examples in the TFLM repository (by not attempting to support various hardware variations, peripherals ...)

We are going to accelerate this transition by starting to delete the existing platform support because this is a common pain point and the existing way of supporting various hardware platforms is error-prone and not scalable. Some related discussions on this topic:
 * https://github.com/tensorflow/tflite-micro/issues/332#issuecomment-888436216
 * https://github.com/tensorflow/tflite-micro/issues/294#issuecomment-883922085
 * https://github.com/tensorflow/tflite-micro/issues/264#issuecomment-877307077
 * https://github.com/tensorflow/tflite-micro/issues/246#issuecomment-875958553
 * https://github.com/tensorflow/tflite-micro/issues/350
 * https://github.com/tensorflow/tflite-micro/issues/170#issuecomment-860888987

advaitjain:
Here is a (partial) list of existing integrations that will be affected by this change along with my guess of some of the relevant stakeholders:

 * apollo3evb, sparkfun_edge: @oclyke
 * arc: @gerbauz
 * arduino: migrated to https://github.com/tensorflow/tflite-micro-arduino-examples
 * ceva: @yair-ehrenwald
 * ecm3531: \<unknown\>
 * esp: @vikramdattu
 * himax: @bigcat-himax
 * mbed: @freddan80 
 * openmvcam: @kwagyeman
 * riscv32_mcu: @petewarden 
 * spresense: @takayoshi-k
 * stm32f4hal: @alxhoff
 * xcore: @andrewxcav

advaitjain:
[...]

PR #432, opened 2021-08-17
Build all the benchmarks as part of a `make build` command.

advaitjain:
With this change:

 * `make build` also build the benchmarks
 * As part of fixing `make build` for EthosU, we are switching it over to using the person_detection.tflite file from the repository instead of downloading it.
 * With this switch for EthosU, we no longer need to download the person_detect_int8 model and the corresponding calls are removed.

This PR also includes the changes from https://github.com/tensorflow/tflite-micro/pull/422 since the Ethos U build will continue to be broken without that.

Manually tested the following command:
```
make -f tensorflow/lite/micro/tools/make/Makefile CO_PROCESSOR=ethos_u OPTIMIZED_KERNEL_DIR=cmsis_nn TARGET=cortex_m_corstone_300 TARGET_ARCH=cortex-m55 test
```

BUG=http://b/197017812


tensorflow/lite/micro/examples/network_tester/network_tester_test.cc
```diff
@@ -24,8 +24,8 @@ limitations under the License.
 #include "tensorflow/lite/schema/schema_generated.h"
 
 #ifdef ETHOS_U
-#include "tensorflow/lite/micro/examples/person_detection/person_image_data.h"
-#include "tensorflow/lite/micro/tools/make/downloads/person_model_int8/person_detect_model_data.h"
+#include "tensorflow/lite/micro/examples/person_detection/testdata/person_image_data.h"
```
deqiangc, line 27: More of a question:  person_detection and network_tester both depends on this file. Is this more for convenience or are there other reasons?

tensorflow/lite/micro/tools/make/ext_libs/person_detection_int8_vela_convert.sh
```diff
@@ -38,49 +38,40 @@ cd "${ROOT_DIR}"
...
   echo "// This file is generated by $0." > ${CONVERTED_PERSON_MODEL_INT8}
-  echo '#include "tensorflow/lite/micro/tools/make/downloads/person_model_int8/person_detect_model_data.h"' >> \
+  echo '#include "tensorflow/lite/micro/models/person_detect_model_data.h"' >> \
        ${CONVERTED_PERSON_MODEL_INT8}
   echo -n "const " >> ${CONVERTED_PERSON_MODEL_INT8}
-  xxd -i ${DOWNLOADED_PERSON_MODEL_INT8_PATH}/person_detect_vela.tflite >> \
-      ${CONVERTED_PERSON_MODEL_INT8}
-  sed -i 's/tensorflow_lite_micro_tools_make_downloads_person_model_int8_person_detect_vela_tflite/g_person_detect_model_data/' \
+  xxd -i ${MODEL_DIR}/person_detect_vela.tflite >> ${CONVERTED_PERSON_MODEL_INT8}
+  sed -i 's/tensorflow_lite_micro_tools_make_gen_cortex_m_corstone_300_cortex_m55_default_genfiles_tensorflow_lite_micro_models_person_detect_vela_tflite/g_person_detect_model_data/' \
```
deqiangc, line 70: nit: maybe comment on where 
tensorflow_lite_micro_tools_make_gen_cortex_m_corstone_300_cortex_m55_default_genfiles_tensorflow_lite_micro_models_person_detect_vela_tflite come from?

tensorflow/lite/micro/benchmarks/Makefile.inc
```diff
@@ -9,21 +9,23 @@ tensorflow/lite/micro/benchmarks/micro_benchmark.h
...
+tensorflow/lite/micro/examples/person_detection/testdata/no_person.bmp
+
 ifeq ($(CO_PROCESSOR),ethos_u)
+  PERSON_DETECTION_BENCHMARK_GENERATOR_INPUTS += \
+    tensorflow/lite/micro/models/person_detect.tflite
+else
   # Ethos-U use a Vela optimized version of the original model.
   PERSON_DETECTION_BENCHMARK_SRCS += \
-  $(MAKEFILE_DIR)/downloads/person_model_int8/person_detect_model_data_vela.cc
+  $(GENERATED_SRCS_DIR)tensorflow/lite/micro/models/person_detect_model_data_vela.cc
```
deqiangc, line 22: nit and totally up to you:  $(GENERATED_SRCS_DIR)/tensorflow would more clearly indicate the expectation.  I think // would do no harm. WDYT?

tensorflow/lite/micro/examples/network_tester/network_tester_test.cc
advaitjain: The network tester example uses the person detect model for the EthosU target. Its not a very good design to have one example reach into the other example's directory for input data. But it is what we have right now.

Once we clean up the examples at a high level with https://github.com/tensorflow/tflite-micro/issues/407, we should certainly revisit how we organize the data and models that are shared across examples.

tensorflow/lite/micro/tools/make/ext_libs/person_detection_int8_vela_convert.sh
advaitjain: yeah, I'm letting all this be for now since it is internals of the Vela system that I am not too familiar with.

deqiangc (approved):

tensorflow/lite/micro/benchmarks/Makefile.inc
advaitjain: The // resulted in the vela compiler's outputs paths being a bit odd. More importantly, we've had double slash cause the MacOS build to fail previously so I am trying to avoid it.

I do agree with you that having `GENERATED_SRCS_DIR` not have a trailing / would be nicer. But that is how the Makefiles have been written so we'd have to make the change in multiple places.

advaitjain: For completeness, here is the MacOS fix: https://github.com/tensorflow/tflite-micro/pull/367

tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -237,6 +237,16 @@ endif
 # runtime that can be linked in to other programs.
 MICROLITE_LIB_NAME := libtensorflow-microlite.a
 
+# Where compiled objects are stored.
```
fanghuaqi, line 240: If this part is placed here, then `TARGET_ARCH` and `BUILD_TYPE` stored in $(MAKEFILE_DIR)/targets/$(TARGET)_makefile.inc will not be able to affect the `GENDIR` variable, for example, for `TARGET=riscv32_mcu`, the TARGET_ARCH will be default set to x86-64, but before it should be `TARGET_ARCH := riscv32_mcu` defined in https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/tools/make/targets/riscv32_mcu_makefile.inc#L2

tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -550,16 +556,6 @@ ALL_SRCS := \
 	$(MICROLITE_CC_KERNEL_SRCS) \
 	$(MICROLITE_TEST_SRCS)
 
-# Where compiled objects are stored.
```
fanghuaqi, line 553: These lines should be kept, and previous hunks should be removed.

fanghuaqi (commented):
`TARGET_ARCH` and `BUILD_TYPE` is not able to be replaced by the ones defined in target makefile.

PR #435, opened 2021-08-18
Fix misc formatting and spelling errors.

njeffrie:
BUG=clean up some internal-only warnings

### DIFF
--- a/tensorflow/lite/micro/build_def.bzl
+++ b/tensorflow/lite/micro/build_def.bzl
@@ -1,20 +1,20 @@
 def micro_copts():
-  return [
-      "-Wall",
-      "-Werror",
-      "-DFLATBUFFERS_LOCALE_INDEPENDENT=0",
-  ]
+    return [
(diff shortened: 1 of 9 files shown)


advaitjain (approved):

advaitjain (changes requested):
We want to continue to disable the layering check. That is the easiest way to ensure that external and internal errors are consistent.

See https://github.com/tensorflow/tflite-micro/pull/386#pullrequestreview-725882136 for additional context. 

I'm happy to switch to something else is we think that that would be preferred (including keeping the layering check and manually fix any errors when we import from github). Let me know what you think.

njeffrie:
Makes sense. I had removed due to warnings in the build, but I'll refrain from removing them now that I understand their purpose.

advaitjain (approved):

advaitjain (approved):

advaitjain (approved):

PR #450, opened 2021-08-20
Make size check workflow error out if memory footprint increase

deqiangc:
between head of main and this PR.

This error can be disabled by putting MEM_CHANGE_OK= tag.
In either case, a memory footprint comparison report will be produced
as part of this workflow action.

The plan is for the code reviewer to audit the report when the
PR increase memory footprint and only
allow the PR that increase memory footprint to be merged with
MEM_CHANGE_OK=what_is_the_reason_for_this_increase_and_why_this_is_ok

The PR shall be up to date for the comparison to be accurate.
So before applying MEM_CHANGE_OK=, please make the PR branch up to date.

BUG=http://b/196637015

### DIFF
--- a/.github/workflows/size_check.yml
+++ b/.github/workflows/size_check.yml
@@ -20,9 +20,16 @@ jobs:
 
     name: Binary Size Estimation
     steps:
+      - name: Get build with PR
       - uses: actions/checkout@v2
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Get build from main
+        uses: actions/checkout@v2
+        with:
+          path: main_ref
+
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/tools/ci_build/test_size.sh
```diff
@@ -14,38 +14,63 @@
...
+# The caller is expected to store this  __BINARY_TARGET_PATH back to its local
+# variable if it needs to use the generated binary target with path later on.
+__BINARY_TARGET_PATH=
+function build_target() {
+  local binary_target=$1
+  local build_type=$2
+  local target=$3
+  local target_arch=$4
+  readable_run make -f tensorflow/lite/micro/tools/make/Makefile clean
+  readable_run make -f tensorflow/lite/micro/tools/make/Makefile third_party_downloads
```
advaitjain, line 37: I think we should only do a clean once. Else the build times will be quite large if we expand to multiple targets.

tensorflow/lite/micro/tools/ci_build/test_size.sh
```diff
@@ -14,38 +14,63 @@
...
+  readable_run make -j8 -f tensorflow/lite/micro/tools/make/Makefile build build_type=${build_type} TARGET=${target} TARGET_ARCH=${target_arch} ${binary_target}
```
advaitjain, line 38: might be worth making this a catch-all `MAKEFILE_OPTIONS` parameter, since these can vary depending on the TARGET. And we're missing `OPTIMIZED_KERNEL_DIR`

tensorflow/lite/micro/tools/ci_build/test_size.sh
```diff
@@ -14,38 +14,63 @@
...
 
+  # Return the relative binary with path and name.
+  __BINARY_TARGET_PATH="tensorflow/lite/micro/tools/make/gen/${target}_${target_arch}_${build_type}/bin/${binary_target}"
```
advaitjain, line 41: one recommendation for returning values from bash functions is to echo them.

https://www.linuxjournal.com/content/return-values-bash-functions

tensorflow/lite/micro/tools/ci_build/test_size.sh
```diff
@@ -14,38 +14,63 @@
...
+size ${CURRENT_BINARY} > ${ROOT_DIR}/ci/size_log.txt
 
-# Next, make sure that the release build succeeds.
-# Build for x86.
-TARGET=linux
-TARGET_ARCH=x86_64
-readable_run make -f tensorflow/lite/micro/tools/make/Makefile clean
-readable_run make -j8 -f tensorflow/lite/micro/tools/make/Makefile build BUILD_TYPE=${BUILD_TYPE} TARGET=${TARGET} TARGET_ARCH=${TARGET_ARCH} ${BENCHMARK_TARGET}
+# Get a clone of the main repo as the reference.
+REF_ROOT_DIR="${ROOT_DIR}/tmp/main_ref"
```
advaitjain, line 59: prefer `mktemp -d`

https://github.com/tensorflow/tflite-micro/blob/4e091a46e935afcb27035e538044e37892d4a983/tensorflow/lite/micro/tools/make/flatbuffers_download.sh#L64

advaitjain (changes requested):
I'm ok starting with the structure from this PR.

However, thinking ahead to where we might want to check few different targets, it might be worth refactoring into bash and Python functions so that we can have the following flow:

For a given platform (e.g. Xtensa)

 * have a list of targets we care about (e.g. keyword_benchmark, person_detection_benchmark, all the kernel tests).
 * be able to reuse the build artifacts for each of these
 * report the size differences for each of the targets
 * raise an error if any one of the sizes has changed.
 * flag in the error message which of the targets has increased in size.

Also, I realized today that for the Xtensa build, fetching the docker image takes ~3mins so we might want to include the size check as part of the current Xtensa workflow instead of creating a new one.

PR #451, opened 2021-08-20
Add README instructions for generating projects for cortex m generic, ensure include paths are added to cortex_m_generic makefile

cdknorow:
cortex-m-generic and CMSIS, update makefile to ensure Include paths for CMSIS
are added to the project file


cdknorow:
@advaitjain this is kind of hackish, not sure if its worth adding, but I felt I'd make a pull request anyway

cdknorow:
@googlebot I consent

advaitjain:
> @advaitjain this is kind of hackish, not sure if its worth adding, but I felt I'd make a pull request anyway

My hope is that you can switch to using the new project generation fairly easily. I would prefer having you switch over to that rather than fixing the makefile based generation (or adding in additional documentation for it).

It is also part of our CI so we have more guarantees that future updates will not break that workflow:
https://github.com/tensorflow/tflite-micro/blob/a9f2e03b943990bc958ce3b92ca76ffc27fcf6d6/tensorflow/lite/micro/tools/ci_build/test_project_generation.sh#L57-L69

cdknorow:
@advaitjain got it, I'll look into that next.

Issue #455, opened 2021-08-23
Tensor data is incorrectly loaded from FlatBuffers on big-endian targets


mateusz-holenko:
Currently `InitializeTfLiteTensorFromFlatbuffer` and `InitializeTfLiteEvalTensorFromFlatbuffer` refer to a data in a FlatBuffer using a direct pointer - an example is:

 https://github.com/tensorflow/tflite-micro/blob/a9f2e03b943990bc958ce3b92ca76ffc27fcf6d6/tensorflow/lite/micro/micro_allocator.cc#L440

Because data in FlatBuffers is always little-endian, on big-endian machines it causes problems because any type wider than 8 bits will be read with bytes swapped.

advaitjain:
Cross-referencing https://github.com/tensorflow/tflite-micro/issues/304 and https://github.com/tensorflow/tflite-micro/pull/456#issuecomment-903921078

advaitjain:
With https://github.com/tensorflow/tflite-micro/pull/689, we are removing endian changes at runtime. Instead, we would like for the byte-swapping to be external to the interpreter.

Issue #304, opened 2021-07-20
Remove runtime code paths for big endian machines


advaitjain:
The binary footprint of TFLM has been creeping up and we are in a phase where we would like to reduce that and push functionality into offline pre-processing or brnaches that can be pruned by the compiler/linker.

One such example is big endian support.
https://github.com/tensorflow/tflite-micro/blob/9d1f67e41aadfa55ac6b8bbf6038402ce22ee1e8/tensorflow/lite/micro/micro_allocator.cc#L369-L390

The recommendation for TFLM users that would like big endian support is to have a pre-processing tool that converts the flatbuffer to have the same endianess as the target platform.

We would be happy to discuss alternate strategies if this suggestion is not going to work in practice. However, we would like to have users not rely on the current way of supporting big-endian.

advaitjain:
tagging @kun-lu20 and @skribm9

advaitjain:
In addition to impact on the TFLM footprint (which we need some testing to validate if indeed there is any), there is also an impact on readability and maintainability of the code.

For example, if we go down the path of https://github.com/tensorflow/tflite-micro/issues/456 then we risk introducing a fair amount of complexity.

mariannems:
Following this, is there any guidelines to run the TFLM examples on big endian devices ?

mariannems:
I would be interested into working on that, but I still have some concerns.

I see how to make a tool that would byte swap the buffers inside the model. That could be done with something similar to [randomize_weights.py](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/tools/randomize_weights.py). It could open the .tflite, loop through the buffers and byte swap them as needed according to the TensorType, and then write the new .tflite using the flatbuffer_utils functions.

But, if the support for big endian is removed from tflm, does it means that the tool would also need to swap all the other values, such as indices or lengths, and other parameters that are written as int32 ? If so, I don't see how it could be made easily. For example, `model_object.Pack(builder)` would generate indices, which would not have been byte swapped.

Any thoughts on that ? Thanks

joshortner:
@mariannems Is there a solution to the issue you raised? I'm working on a big endian machine, used the functions in flatbuffer_utils.py to convert the model, but it seems like some of the indices in the model weren't converted causing asserts to trigger. Maybe I'm missing something simple.

kun-lu20:
[...]

PR #456, opened 2021-08-23
Fix tensor data loading on big-endian targets 

mateusz-holenko:
This is a proof of concept of how a solution for #455 might look like. 

We make `InitializeTfLiteTensorFromFlatbuffer` and `InitializeTfLiteEvalTensorFromFlatbuffer` use `FlatBufferVectorToTfLiteTypeArray` instead of directly assigning a pointer into FlatBuffer data to the tensor's data field. 

This does not fully solve the problem, because it only handles `int32`s (so, as it stands, big-endian targets will be able to correctly read all 8-bit types and `int32`), but it's enough to get the `hello_world` example to run and give correct outputs on SPARC (leon3).

### DIFF
--- a/tensorflow/lite/micro/examples/hello_world/main_functions.cc
+++ b/tensorflow/lite/micro/examples/hello_world/main_functions.cc
@@ -33,7 +33,7 @@ TfLiteTensor* input = nullptr;
 TfLiteTensor* output = nullptr;
 int inference_count = 0;
 
-constexpr int kTensorArenaSize = 2000;
+constexpr int kTensorArenaSize = 3000;
 uint8_t tensor_arena[kTensorArenaSize];
 }  // namespace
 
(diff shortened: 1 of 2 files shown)


advaitjain:
Per https://github.com/tensorflow/tflite-micro/issues/304, I would like to avoid adding this complexity into the TFLM inference engine.

Rather, my preference here would be to add a separate pre-processing tool that converts the flatbuffer to big endian and be able to assume that the endianness of the flatbuffer is the same as the target that the inference is being run on.

PiotrZierhoffer:
Hey @advaitjain , as discussed with @petewarden we'll leave it hanging for a while until you discuss the best way forward. The problem is there and the solution is, clearly, not generic (focusing on `int32`), so we'll proceed according to what you decide

mateusz-holenko:
@googlebot I fixed it.

advaitjain:
closing this PR due to https://github.com/tensorflow/tflite-micro/issues/455#issuecomment-971923810

Issue #460, opened 2021-08-24
Not possible to download third party software in a clean repo


mansnils:
The creating of the download folder from the Makefile was probably removed by mistake. Now it is not possible to run make since the download script expects the download folder to be existing.

advaitjain:
We had a bad merge which resulted in this regression.

Fixed with:
https://github.com/tensorflow/tflite-micro/commit/94ddbfe8ba348d56bc94486ee0120eb6e556fd98

PR #461, opened 2021-08-24
Create download directory if it not exists

mansnils:
Fix for: https://github.com/tensorflow/tflite-micro/issues/460


advaitjain:
We have reverted the bad merge that caused this issue: https://github.com/tensorflow/tflite-micro/issues/460#issuecomment-904865970

Issue #438, opened 2021-08-19
TCN models containing expand dims and conv1d layers induce "Arena size is too small issue"


swapnilsayansaha:
I am trying to upload some lightweight TCN models to STM32L476RG Nucleo running Mbed OS.

Example model 1 (the tcn layer is from keras tcn library https://github.com/philipperemy/keras-tcn):
```
Model: "model"
__________________________________________________________________________________________________
Layer (type)                    Output Shape         Param #     Connected to                     
==================================================================================================
input_1 (InputLayer)            [(None, 400, 6)]     0                                            
__________________________________________________________________________________________________
tcn (TCN)                       (None, 32)           31840       input_1[0][0]                    
__________________________________________________________________________________________________
tf.reshape (TFOpLambda)         (None, 32, 1)        0           tcn[0][0]                        
__________________________________________________________________________________________________
max_pooling1d (MaxPooling1D)    (None, 16, 1)        0           tf.reshape[0][0]                 
__________________________________________________________________________________________________
flatten (Flatten)               (None, 16)           0           max_pooling1d[0][0]              
__________________________________________________________________________________________________
pre (Dense)                     (None, 32)           544         flatten[0][0]                    
__________________________________________________________________________________________________
velx (Dense)                    (None, 1)            33          pre[0][0]                        
__________________________________________________________________________________________________
vely (Dense)                    (None, 1)            33          pre[0][0]                        
==================================================================================================
Total params: 32,450
Trainable params: 32,450
Non-trainable params: 0
```
The above model consumes 162 kB of flash (1 MB capacity) and a total of ~50 kB (128 kB capacity) of RAM on the Mbed board (includes other dependencies) for arena of size 32X1000.

Example model 2 (which is even lighter!):
```
__________________________________________________________________________________________________
Layer (type)                    Output Shape         Param #     Connected to                     
==================================================================================================
input_2 (InputLayer)            [(None, 140, 1)]     0                                            
__________________________________________________________________________________________________
conv1d (Conv1D)                 (None, 140, 2)       22          input_2[0][0]                    
__________________________________________________________________________________________________
batch_normalization (BatchNorma (None, 140, 2)       8           conv1d[0][0]                     
__________________________________________________________________________________________________
zero_padding1d (ZeroPadding1D)  (None, 140, 2)       0           batch_normalization[0][0]        
__________________________________________________________________________________________________
conv1d_1 (Conv1D)               (None, 140, 11)      242         zero_padding1d[0][0]             
[...]

PR #473, opened 2021-08-27
lstm:training micro speech lstm model

nyadla-sys:
BUG= training micro speech lstm model and example
Signed-off-by: Niranjan Yadla <nyadla@cadence.com>


nyadla-sys:
@advaitjain, Could you please merge this change?

advaitjain:
I have removed some of the files that should no longer be needed with https://github.com/tensorflow/tflite-micro/pull/473/commits/860877b1a571905cd78bc39e90962dd3b72746a0

There are some updates to the documentation as well as including the wav file to spectrogram conversion that will be addressed with follow-up PRs.

advaitjain (dismissed):

advaitjain (changes requested):
One last remaining item is that we will need to add a separate LICENSE file to third_party/xtensa.

I'm assuming that since this started out as a contribution into tensorflow/lite/micro, we can use the tflite-micro LICENSE: https://github.com/tensorflow/tflite-micro/blob/main/LICENSE

Let me know what you think.

nyadla-sys:
@advaitjain, I have just updated LICENSE file to xtensa folder
Please review and merge #438 PR

advaitjain (approved):

PR #477, opened 2021-09-01
Experimental pull request for review of use of ExternalContext to share network context between application and operators.

cad-audio:
@petewarden , @advaitjain 

Could you please review this experimental PR which tries to add and use Get/SetExternalContext interface to share network context from application to operators?
Note, this is a dirty implementation, only for review of flow and use of interface.
(For e.g. TfLiteExternalContext* external_contexts_[kTfLiteMaxExternalContexts] is not a private array currently).

### DIFF
--- a/tensorflow/lite/c/common.h
+++ b/tensorflow/lite/c/common.h
@@ -55,7 +55,8 @@ typedef enum TfLiteExternalContextType {
   kTfLiteGemmLowpContext = 1,    // include gemm_support.h to use.
   kTfLiteEdgeTpuContext = 2,     // Placeholder for Edge TPU support.
   kTfLiteCpuBackendContext = 3,  // include cpu_backend_context.h to use.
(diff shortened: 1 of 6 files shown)


jenselofsson:
I think this looks promising. Have you thought about the use case where one would want to send multiple TfLiteExternalContext:s (eg different ones depending on which is the recieving kernel), and how the kernels would identify which one is intended for them?

cad-audio:
> I think this looks promising. Have you thought about the use case where one would want to send multiple TfLiteExternalContext:s (eg different ones depending on which is the recieving kernel), and how the kernels would identify which one is intended for them?

For the current use case, we would need single context for the whole network. But I see your point, it would be good to have flexibility to share multiple kernel specific contexts. One quick way to do that would be to extend _TfLiteExternalContextType_ enum to have more entries.

PR #481, opened 2021-09-03
Some improvements to build scripts

shlmregev:
1. Add TENSORFLOW_ROOT as a prefix to paths in helper_functions.inc. This enables us to include this file from an external directory. Useful when building a custom op library.
2. Add a script size_hexagon_binary.sh to measure the size of a Hexagon binary, and a matching make macro/rule to invoke it. Similar scripts can be created for any target.
3. Lint fixes to generate_cc_array.py
4. Move HEXAGON_TFLM_LIB from the ext_lib makefile to hexagon_makefile.inc. This library is needed by any binary, so it belongs on the LDFLAGS.

Corresponding internal change: http://cl/394723466

BUG=http://b/183286935 and future improvements in http://b/199160755


njeffrie (dismissed):

advaitjain:
My bad for not reviewing the internal change completely. I'm not completely confident that we want to add the size support to the Makefile.

Let's merge this PR as is and then revisit. Created http://b/199160755 to figure out if that is a direction that TFLM wants to support or not.

advaitjain (dismissed):

advaitjain (approved):

Issue #486, opened 2021-09-08
Build OPTIMIZED_KERNEL_DIR=cmsis_nn with a manually provided CMSIS-NN lib


mansnils:
As default TFLM is built using CMSIS-NN code that is downloaded to the TFLM tree.
It is also possible to build TFLM using CMSIS-NN code from an external path (CMSIS_PATH=......).

It is now  possible to build CMSIS-NN a lib.

So it should also be possible to build TFLM with a manually provided CMSIS-NN lib.

Issue #170, opened 2021-06-13
Trouble using Make to build examples.


embeddetech:
I'm seeing build errors with the different TFLM build attempts I've been making. I have not tried a Bazel build, if there are build instructions on how to do a Bazel build for this repo on Windows I would try that, currently I am trying to build with a Make project. I am successfully building Tensorflow Lite (from the main Tensorflow repo) with CMake, but with that build I don't have clarity on what is just needed for TFLM.

Build attempt #1:

Following the first Micro Speech example "Deploy to ARC EM SDP" found here:
https://github.com/tensorflow/tflite-micro/tree/main/tensorflow/lite/micro/examples/micro_speech

Command line commands from an empty folder:
git clone https://github.com/tensorflow/tflite-micro tflite-micro
cd tflite-micro
make -f tensorflow/lite/micro/tools/make/Makefile TARGET=arc_emsdp ARC_TAGS=reduce_codesize OPTIMIZED_KERNEL_DIR=arc_mli generate_micro_speech_mock_make_project

Result: Several packages are downloaded, but the end result is "make: *** No rule to make target 'generate_micro_speech_mock_make_project'.  Stop."

Build attempt #2:

Following the  Micro Speech example "Deploy to ESP32" found here:
https://github.com/tensorflow/tflite-micro/tree/main/tensorflow/lite/micro/examples/micro_speech

ESP IDF should be properly installed.

Command line commands from an empty folder:
git clone https://github.com/tensorflow/tflite-micro tflite-micro
cd tflite-micro
make -f tensorflow/lite/micro/tools/make/Makefile TARGET=esp generate_micro_speech_esp_project

Result:
There are a couple "FIND: Parameter format not correct" messages and "File not found - *.cc", "File not found - *.h". It appears that pigweed is downloaded, along with some other packages. Ultimately, the build fails with "make: *** No rule to make target 'generate_micro_speech_esp_project'.  Stop.".

Build attempt #3:

Following the Micro Speech example "Deploy to Sparkfun Edge" found here:
https://github.com/tensorflow/tflite-micro/tree/main/tensorflow/lite/micro/examples/micro_speech

Command line commands from an empty folder:
git clone https://github.com/tensorflow/tflite-micro tflite-micro
cd tflite-micro
make -f tensorflow/lite/micro/tools/make/Makefile TARGET=sparkfun_edge OPTIMIZED_KERNEL_DIR=cmsis_nn micro_speech_bin

[...]

PR #487, opened 2021-09-08
Add option to provide an external CMSIS-NN lib

mansnils:
Add new build option for when building with OPTIMIZED_KERNEL_DIR=cmsis_nn and also update the CMSIS-NN README.

This is a fix for: https://github.com/tensorflow/tflite-micro/issues/486

### DIFF
--- a/tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
+++ b/tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
@@ -1,6 +1,12 @@
 # Enable u-arch specfic behaviours
 ifneq (,$(filter $(TARGET_ARCH), x86_64))
-    # CMSIS-NN optimizations not supported
+  # CMSIS-NN optimizations not supported
+endif
+
+ifneq (,$(CMSIS_NN_LIBS))
+  ifeq (,$(CMSIS_PATH))
+    $(error CMSIS_NN_LIBS provided but not CMSIS_PATH)
+  endif
 endif
 
 # Unless an external path is provided we force a download during the first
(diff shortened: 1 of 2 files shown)


freddan80 (approved):

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
```diff
@@ -14,10 +20,12 @@ ifeq ($(CMSIS_PATH), $(CMSIS_DEFAULT_DOWNLOAD_PATH))
   endif
 endif
 
-THIRD_PARTY_CC_SRCS += \
-  $(call recursive_find,$(CMSIS_PATH)/CMSIS/NN/Source,*.c)
-THIRD_PARTY_CC_HDRS += \
-  $(call recursive_find,$(CMSIS_PATH)/CMSIS/NN/Include,*.h)
+ifeq (,$(CMSIS_NN_LIBS))
+  THIRD_PARTY_CC_SRCS += $(call recursive_find,$(CMSIS_PATH)/CMSIS/NN/Source,*.c)
```
advaitjain, line 24: What do you think of replace this with a find shell command (and then we can replace the recursive_find macro and have the makefiles be a bit simpler)?

something like
https://github.com/tensorflow/tflite-micro/blob/de29703ba0847677ce7cc7df1165e222501946da/tensorflow/lite/micro/tools/make/ext_libs/xtensa.inc#L30-L31

advaitjain (approved):
apologies for the delayed review. Going to approve and merge this PR as is, but I'm curious what you think of removing `recursive_find`?

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
mansnils: Will that work on Window? I recall something like that was the reason for recursive_find. If yes I am fine with replacing it.

advaitjain: Unclear. We are already using various find commands, so I think we can switch this one over as well.

https://github.com/tensorflow/tflite-micro/blob/96e109101e60737e17115d16660591c38b90d69d/tensorflow/lite/micro/tools/make/Makefile#L386-L390

With the new project generation and #407, we expect the majority of Windows use-cases (which were around makefile-based project generation) to be supported via external TFLM examples repositories.

We do have users attempting to `make` on Windows and that is failing for reasons that we have been open about not being able to support (e.g. https://github.com/tensorflow/tflite-micro/issues/170)

PR #499, opened 2021-09-13
Made Xtensa kernels available via bazel.

jackagoogle:
Creates a micro_ops implementation for each of the target kernel implementations. i.e. `micro_ops` for the reference ops, `xtensa_hif4_micro_ops` for the Xtensa HiFi4 ops, etc.

This resembles how the Makefile works, but using Bazel macros instead. This paradigm should extend to all target implementations, not just Xtensa.


petewarden:
Is this still active, or has it been superseded by the more recent changes?

jackagoogle:
> Is this still active, or has it been superseded by the more recent changes?

This can be closed.

PR #502, opened 2021-09-13
Remove some of the hardware specific code that has not been updated for a while.

advaitjain:
We are moving towards removing ***all*** hardware-specific example code out of the TFLM repository, and this PR is a step in that direction.

BUG=https://github.com/tensorflow/tflite-micro/issues/407


njeffrie (approved):
Just checking, do we have clear instructions for hardware vendors to create their forks of the examples? I want to make sure there's a clear path forward if an example stops working on their hardware.

advaitjain:
> Just checking, do we have clear instructions for hardware vendors to create their forks of the examples? I want to make sure there's a clear path forward if an example stops working on their hardware.

Yes, https://github.com/tensorflow/tflite-micro/issues/407 links to the updated instructions: https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/new_platform_support.md

PR #506, opened 2021-09-15
Add reference kernel support for SVDF op with all 8 bit weights and state

rockyrhodes:
The extra memory usage comes from 7 OpDataSvdf structs.  I only added 4 bytes to each, but with 16-byte alignment I got unlucky and added 16 bytes total to each of these allocations.

BUG=179816930


tensorflow/lite/micro/kernels/svdf_test.cc
```diff
@@ -743,81 +703,17 @@ TF_LITE_MICRO_TEST(SvdfQuantized2x2Input2x4OutputShouldMatchGolden) {
...
-      tflite::testing::time_weights_data_16x1x1, activation_state_data_mutable,
-      tflite::testing::bias_data_16x1x1, scratch_buffer, output_data,
-      tflite::testing::input_data_16x1x1, input_size,
-      tflite::testing::golden_output_relu_16x1x1);
-}
-#endif
-
-TF_LITE_MICRO_TEST(SvdfQuantized1x16Input64x1OutputShouldMatchGolden) {
+// Template parameter sets type of both time_weights and activation_state.
+template <typename T>
```
njeffrie, line 715: I'm not sure these templates are making our tests clearer. I don't have a strong preference, but it could be simpler to just add copies of SvdfQuantized1x16Input64x1OutputShouldMatchGolden and SvdfQuantized1x16Input64x1OutputReluShouldMatchGolden with int8 activation states. I'm curious about Advait's feelings about this.

tensorflow/lite/micro/kernels/svdf.h
```diff
@@ -47,6 +48,8 @@ extern const int kSvdfInputActivationStateTensor;
 extern const int kSvdfOutputTensor;
 
 // TensorflowLite Micro-specific reference implementation for Integer SVDF.
+// One version for each of int8_t and int16_t time_weights/state.
+template <typename T>
 void EvalIntegerSvdfReference(TfLiteContext* context, TfLiteNode* node,
```
advaitjain, line 53: move template definition to the header: https://google.github.io/styleguide/cppguide.html#Self_contained_Headers

In this case, that would be really annoying so we could have two functions

`EvalInt8SvdfReference` and `EvalInt16SvdfReference` in the header. And these two can be implemented with a call to the function template in the .cc file

tensorflow/lite/micro/kernels/svdf_test.cc
```diff
@@ -712,11 +673,11 @@ TF_LITE_MICRO_TEST(SvdfQuantized2x2Input2x4OutputShouldMatchGolden) {
...
 
-  float input_scale = 2.5f / INT8_MAX;              // Range is [-2.5, 2.5]
-  float feature_weights_scale = 1.f / INT8_MAX;     // Range is [-1, 1]
-  float time_weights_scale = 1.f / INT16_MAX;       // Range is [-1, 1]
-  float activation_state_scale = 16.f / INT16_MAX;  // Range is [-16, 16]
-  float output_scale = 1.f / INT8_MAX;              // Range is [-1, 1]
+  float input_scale = 2.5 / INT8_MAX;
+  float feature_weights_scale = 1.0 / INT8_MAX;
+  float time_weights_scale = 1.0 / std::numeric_limits<T>::max();
+  float activation_state_scale = 1.49 / std::numeric_limits<T>::max();
```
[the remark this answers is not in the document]
rockyrhodes: Yes.  The previous range yielded activation state that was too coarsely quantized.

tensorflow/lite/micro/kernels/svdf_test.cc
advaitjain: I'm happy to duplicate, especially since we want to remove the int16 version completely once everything is switch over to int8.

tensorflow/lite/micro/kernels/svdf.cc
```diff
@@ -66,11 +66,26 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
+        case kTfLiteInt8: {
+          EvalIntegerSvdfReference<int8_t>(
+              context, node, input, weights_feature, weights_time, bias, params,
+              activation_state, output, data);
+          return kTfLiteOk;
+          break;
+        }
+        default:
+          TF_LITE_KERNEL_LOG(context, "Type %s not currently supported.",
+                             TfLiteTypeGetName(weights_time->type));
```
advaitjain, line 86: nit: we don't need this macro anymore. Can directly use MicroPrintf (http://b/158205789)

https://github.com/tensorflow/tflite-micro/blob/26a28789f3a3328bbd4d2033375850562c3cc8e8/tensorflow/lite/micro/kernels/read_variable.cc#L70-L72

tensorflow/lite/micro/kernels/svdf_test.cc
rockyrhodes: I changed the templating a bit as Advait suggested below.  Let me know if everyone is okay with this.

advaitjain (dismissed):

tensorflow/lite/micro/kernels/svdf.cc
rockyrhodes: Done.

tensorflow/lite/micro/kernels/svdf.h
rockyrhodes: Done.

advaitjain (approved):

advaitjain (approved):

PR #512, opened 2021-09-15
lstm:xtensa:update colab link and model.zip

nyadla-sys:
BUG=cleanup

### DIFF
--- a/third_party/xtensa/examples/micro_speech_lstm/train/README.md
+++ b/third_party/xtensa/examples/micro_speech_lstm/train/README.md
@@ -51,21 +51,16 @@ Train the model in the cloud using Google Colaboratory.
 
 ## Trained Models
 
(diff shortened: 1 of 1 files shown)


nyadla-sys:
@advaitjain Please review and merge this #512  PR .
Updated colab links to point to tflite-micro github

advaitjain (changes requested):
I don't think we should put in model.zip at all. We have the tflite files checked in already at https://github.com/tensorflow/tflite-micro/blob/main/third_party/xtensa/examples/micro_speech_lstm/micro_speech_lstm.tflite and it should be sufficient to link to that from the readme, right?

nyadla-sys:
@advaitjain Sure,let me delete model.zip.
I guess  tflite file is sufficient

nyadla-sys:
@advaitjain updated the same ,please review PR

nyadla-sys:
@advaitjain Please approve this pR

advaitjain (approved):

PR #528, opened 2021-09-22
Add 8-bit version of keyword benchmark.

rockyrhodes:
Scrambled data was generated from an 8-bit model in my svdf_8bit branch
like this:

cp googledata/speech/greco3/models/shotword/x_google/en/hifiep_default_20190510_requant/artifacts/hotword.tflm.tflite /tmp/hotword.tflite

blaze run //third_party/tensorflow/lite/tools:randomize_weights -- --input_tflite_file=/tmp/hotword.tflite --output_tflite_file=/tmp/hotword_scrambled.tflite

blaze run third_party/tflite_micro/google/hotword/tflite_flatbuffer_align /tmp/hotword_scrambled.tflite /tmp/hotword_scrambled_aligned.tflite

xxd -i /tmp/hotword_scrambled_aligned.tflite /tmp/hotword_scrambled_aligned.cc

BUG=523


njeffrie (changes requested):
Can we split this into a separate benchmark, perhaps called keyword_benchmark_8bit? My main concern is that it will no longer be possible to run the benchmark with optimized kernels, and our size estimates will not be accurate.

rockyrhodes:
Sure.  I'll split it into a new benchmark next week.

PR #529, opened 2021-09-22
Remove unused code.

advaitjain:
BUG=cleanup and http://b/189926408#comment28


advaitjain:
@jenselofsson, this function was added with https://github.com/tensorflow/tensorflow/pull/38121 and I'm curious if it is still needed.

njeffrie (commented):
wasn't this added as part of the ARM offline memory planning work? I wonder if there is still a use for it.

freddan80:
@advaitjain @njeffrie, correct. We added it for debug purposes in case we'd run in to issues with the offline memory planning, but haven't used it that much tbh. If there's strong motivation to remove it, I'm fine with that. Otherwise, I'd prefer to keep it. Let me know.

njeffrie (approved):
I'm in agreement that removing this will be best for readability.

freddan80:
> The motivation for removing it it both readability about a seemingly unused function as well as the fact that certain TFLM-users are seeing compiler warnings as a result of this. Additionally, we are starting to think about improvements to the memory framework and any cleanup will be useful for that work too.
> 
> I think such debug functionality can now be implemented using the approach of the RecodringMicroAllocator i.e. by creating a class that derives from MicroAllocator which is then passed into the interpreter
> 
> https://github.com/tensorflow/tflite-micro/blob/392a7f30e42d823be5e8e8c2ff557d7e1224fa91/tensorflow/lite/micro/micro_interpreter.h#L54-L62
> 
> This interpreter API did not exist when [tensorflow/tensorflow#38121](https://github.com/tensorflow/tensorflow/pull/38121) was created.

Understand. No issues from my side.

PR #530, opened 2021-09-22
__restrict__ is an unnecessary optimization in the reference code.

advaitjain:
If needed, we will need to add an ifdef similar to https://github.com/tensorflow/tflite-micro/blob/180de4b73b28bde0d39dc9f2a0bad68d6a9e0fc0/tensorflow/lite/kernels/internal/portable_tensor_utils.h#L26-L28

BUG=cleanup and http://b/189926408#comment28

### DIFF
--- a/tensorflow/lite/micro/kernels/svdf_common.cc
+++ b/tensorflow/lite/micro/kernels/svdf_common.cc
@@ -227,10 +227,9 @@ void EvalInt8SvdfReference(TfLiteContext* context, TfLiteNode* node,
 
 static inline void ApplyTimeWeightsBiasAndActivation(
     int batch_size, int memory_size, int num_filters, int num_units, int rank,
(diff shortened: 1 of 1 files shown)


deqiangc (commented):
Even though it is not required right now, people may evaluate TFLM on a new platform based on reference codes. Therefore, adding the following

 #if defined(_MSC_VER) 
 #define __restrict__ __restrict 
 #endif 

to our common util and start using it may be worthy in the long run.  __restrict__ helps compiler and can save code size AFAIK. Would you like to consider either adding a todo with tracking bug OR just add the #ifdef? 

Other than that, LGTM

deqiangc (approved):

advaitjain:
I view __restrict__ as premature optimization in this case. It is mostly inherited from reference code in TfLite and there hasn't been any profiling to see what (if any) the benefits are.

PR #536, opened 2021-09-25
Wrap 16-bit Kiss FFT in a C++ namespace

shlmregev:
This will enable us to support multiple copies of Kiss FFT for different
resolutions within the same application without running into linker
errors.

BUG=200737980


advaitjain (dismissed):

tensorflow/lite/experimental/microfrontend/lib/kiss_fft_int16.cc
```diff
@@ -0,0 +1,8 @@
+#include "tensorflow/lite/experimental/microfrontend/lib/kiss_fft_common.h"
+
+#define FIXED_POINT 16
+namespace kissfft_fixed16 {
+#include "kiss_fft.c"
+#include "tools/kiss_fftr.c"
```
advaitjain, line 6: This is making me a bit nervous since the .c files being included are in the third_party/kissfft.

What do you think of this alternative (we can merge the current PR as-is and I can do this follow-up change):
 * we move kiss_fft_int16.c to third_party/kissfft (for the bazel build)
 * have another copy of kiss_fft_int16 in micro/tools/make/downloads/kissfft as part of the TFLM download and path (for the makefile build) 

Having the various kiss_fft_int16, _float, _int32 live next to the rest of the KissFFT code would be easier to maintain, I think.

shlmregev: It's up to you. I took the approach of not contaminating the "downloads" folder, so that it's as close as possible to the original, minus the patch. Having a single copy of the wrapper is also an advantage.

advaitjain (approved):

tensorflow/lite/experimental/microfrontend/lib/kiss_fft_int16.cc
advaitjain: ok, I'll think about what direction we would want to go.

PR #538, opened 2021-09-27
Generate keyword benchmark data from model.

rockyrhodes:
BUG=http://b/179816930


tensorflow/lite/micro/benchmarks/BUILD
```diff
@@ -29,10 +33,10 @@ cc_library(
 cc_library(
     name = "keyword_scrambled_model_data",
     srcs = [
-        "keyword_scrambled_model_data.cc",
+        "//tensorflow/lite/micro/models:generated_keyword_scrambled_model_cc",
```
njeffrie, line 36: Note (not actionable): There was a debate when I created the models directory which models should live here vs with their example or benchmark. The decision was to only put models here that are needed in >1 place. Still, I like putting the keyword model here since we'll probably share it with the memory footprint example.

tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -679,10 +679,10 @@ ifneq ($(findstring $(EXPLICITLY_SPECIFIED_TEST),$(MICROLITE_TEST_SRCS)),)
   MICROLITE_TEST_SRCS := $(filter-out $(EXPLICITLY_SPECIFIED_TEST), $(MICROLITE_TEST_SRCS))
   EXPLICITLY_SPECIFIED_TEST_SRCS := \
   $(EXPLICITLY_SPECIFIED_TEST) \
-  tensorflow/lite/micro/benchmarks/keyword_scrambled_model_data.cc \
+  $(GENERATED_SRCS_DIR)tensorflow/lite/micro/models/keyword_scrambled_model_data.cc \
```
njeffrie, line 682: It's odd to me that this was in the Makefile to start with. What happens if we remove it?

rockyrhodes: The memory_arena_threshold_test fails to compile with the Makefile without this.

njeffrie (approved):

tensorflow/lite/micro/tools/make/Makefile
advaitjain: this should depend on the tflite file, right? Else the build will fail if we no longer do the tflite to model_data generation as part of the keyword_benchmark.

PR #576, opened 2021-10-06
Add a NonPersistentMemoryPlannerShim to support a memory plan in c structure format

deqiangc:
Add a NonPersistentMemoryPlannerShim to support a memory plan in c structure format
This memory plan specifies the memory layout of all non persistent
buffers in the arena.

This is an early stage exploration of what the next version of
offline memory planning in TFLM might look like; thus this is
an experimental feature subjected to change.

This corresponds to internal c/401255189

BUG=http://b/199230276


deqiangc:
> I agreed that the implementation should be kept simple, for many reasons.
> 
> However, in order to support current functionality, that is, the numerous calls to [RequestScratchBufferInArena](https://github.com/tensorflow/tflite-micro/blob/1d06ad15715289c16e44cfe04cade90d743c25e0/tensorflow/lite/c/common.h#L762-L768) in the code base, I'd assume some kind of support is necessary... Or am I missing something? Each call to [RequestScratchBufferInArena](https://github.com/tensorflow/tflite-micro/blob/1d06ad15715289c16e44cfe04cade90d743c25e0/tensorflow/lite/c/common.h#L762-L768) would eventually lead to a call to AddBuffer(), which is unsupported by the [NonPersistentMemoryPlannerShim](https://github.com/tensorflow/tflite-micro/blob/1d06ad15715289c16e44cfe04cade90d743c25e0/tensorflow/lite/micro/memory_planner/non_persistent_buffer_planner_shim.cc#L28-L33).
> 
> One other question, will the existing a. Offline memory planer (via tflite metadata) and b. GreedyMemoryPlanner remain in the TFLM code base for compatibility?
> 
> Tagging @mansnils and @dominicsymes.

Hello, Fredrik,

For the second question, yes. The GreedyMemoryPlanner and its associated support for offline tflite metadata support remains in the code base AND remains as the default.   The NonPersistentMemoryPlannerShim is kind of a OPT in replacement  only if a user choose to do so as illustrated in https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/offline_memory_plan.md

As to your first question, yes, there is a another CL in work to address that and also better hook up with MicroAllocator.  

Thanks and best regards!
Deqiang

Issue #589, opened 2021-10-13
Null pointer bias in fully connected op is not working


mansnils:
Null pointer bias should be supported but is not really working.

A null pointer bias is received in Prepare but not in Eval.

How to reproduce: Try running a model with a FC operator without bias data and then compare the output with that of TFL.

PR #590, opened 2021-10-13
Handle nullptr bias in fully_connected op

### LINKED ISSUE #685: Fully connected node without bias has invalid parameters when run with xtensa nnlib on HiFi4
I am trying to deploy a model trained in tensorflow on the HiFi4 using TFLM. After training a simple NN in tensorflow (a one layer multilayer perception without a bias term) and converting to tflite using the standard optimizations, I can try to run the model on the HiFi4 and run inference. Everything appears to work fine until I call invoke on the interpreter, in which case the return code is -1.

After adding some more logging, I have noticed that the input to the fully connected nnlib node is 
```
EvalQuantizedInt8: batch=0 p_out=0x003ed630 p_weight=0x003d7b7c p_inp=0x003ecf30 p_bias=0x00000000
```
Hence `p_bias` is being set to null. However, this is unfortunately prohibited by the xtensa nnlib library:

https://github.com/foss-xtensa/nnlib-hifi4/blob/42da7e5c17844a0a83dc6b32f06258504bcdd373/xa_nnlib/algo/kernels/fc/hifi4/xa_nn_fully_connected.c#L53

If I retrain my tensorflow model but make sure to include a bias term, everything works fine. I believe a similar error was patched in the CMSIS library: https://github.com/ARM-software/CMSIS_5/issues/920. Would it be possible to add some similar protection here?

mansnils:
* Validate tensor index in GetEvalInput.
* Check for nullptr bias tensor in fully connected kernel.

BUG=Fixes https://github.com/tensorflow/tflite-micro/issues/589, fixes #685  and http://b/205306469


tensorflow/lite/micro/kernels/kernel_util.h
```diff
@@ -39,6 +40,13 @@ inline TfLiteEvalTensor* GetMutableEvalInput(const TfLiteContext* context,
 // Returns the TfLiteEvalTensor struct for a given input index in a node.
 inline const TfLiteEvalTensor* GetEvalInput(const TfLiteContext* context,
                                             const TfLiteNode* node, int index) {
+  const int tensor_index = tflite::ValidateTensorIndexing(
+      context, index, node->inputs->size, node->inputs->data);
+
+  if (tensor_index < 0) {
+    return nullptr;
+  }
```
advaitjain, line 48: let's move this check to GetMutableEvalInput

tensorflow/lite/kernels/kernel_util.h
```diff
@@ -144,6 +144,17 @@ TfLiteStatus GetIntermediatesSafe(const TfLiteContext* context,
                                   TfLiteTensor** tensor);
 #endif  // TF_LITE_STATIC_MEMORY
 
+inline int ValidateTensorIndexing(const TfLiteContext* context, int index,
+                                  int max_size, const int* tensor_indices) {
```
advaitjain, line 148: let's have this function not be inlined. I'm happy to make this change in TF -- that might be faster, let me know.

advaitjain (changes requested):
Looks good to me. Is there a test case that can be added?

tensorflow/lite/micro/kernels/kernel_util.h
mansnils: ok

mansnils:
@advaitjain  Does it look ok?

advaitjain:
sorry for the delay, looking at this now.

advaitjain:
> In this PR I will just copy ValidateTensorIndexing to lite/micro/kernels/kernel_util. Once that is merged I am happy to make a new set of PRs were ValidateTensorIndexing is moved to a common place between TFL and TFLM. If that sounds ok..

Thanks. Thinking more about it, I think having a separate `ValidateTensorIndexing` for TFLM is the way to go.

I have made some fixes to the Xtensa kernels as well (including https://github.com/tensorflow/tflite-micro/pull/711) and once the CI passes, we can merge this PR.

advaitjain (dismissed):

advaitjain (approved):

PR #614, opened 2021-10-21
Add NonPersistentMemoryPlanner integration test with MicroAllocator

deqiangc:
Add a buffer_request_count to keep track of buffer request to do error
checking.

Also rename variable names in NonpersistentMemoryPlannerShim per extra
review comments.

This corresponds to internal cl/403489054

BUG=http://b/199230276

PR #616, opened 2021-10-21
Add build file format check and fix in test_code.sh

deqiangc:
Add build file format check and fix in test_code.sh
and also run format fix on all BUILD and bzl files.
 
After this PR, BUILD and bzl files will be checked in presubmit.

BUG=http://b/203803273

### DIFF
--- a/tensorflow/lite/micro/BUILD
+++ b/tensorflow/lite/micro/BUILD
@@ -103,10 +103,10 @@ cc_library(
         "//tensorflow/lite/core/api:error_reporter",
         "//tensorflow/lite/core/api:op_resolver",
         "//tensorflow/lite/kernels/internal:compatibility",
-        "//tensorflow/lite/schema:schema_fbs",
-        "//tensorflow/lite/schema:schema_utils",
         "//tensorflow/lite/micro/memory_planner:greedy_memory_planner",
         "//tensorflow/lite/micro/memory_planner:micro_memory_planner",
+        "//tensorflow/lite/schema:schema_fbs",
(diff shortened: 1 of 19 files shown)


tensorflow/lite/micro/BUILD
```diff
@@ -391,6 +391,8 @@ cc_test(
         ":micro_error_reporter",
         ":test_helpers",
         "//tensorflow/lite/c:common",
+        "//tensorflow/lite/micro/memory_planner:memory_plan_struct",
```
njeffrie, line 394: Let's land #614 first so these are not added in this change.

tensorflow/lite/micro/memory_planner/non_persistent_buffer_planner_shim.cc
```diff
@@ -20,16 +20,21 @@ limitations under the License.
 namespace tflite {
 
 NonPersistentMemoryPlannerShim::NonPersistentMemoryPlannerShim(
-    const BufferPlan* offline_buffer_plan)
-    : offline_buffer_plan_(offline_buffer_plan) {}
+    const BufferPlan* buffer_plan)
```
njeffrie, line 23: Same as above.

tensorflow/lite/micro/memory_planner/non_persistent_buffer_planner_shim.cc
```diff
@@ -20,16 +20,21 @@ limitations under the License.
 namespace tflite {
```
advaitjain, line 20: unrelated change appears to have crept in.

PR #643, opened 2021-11-01
Fix xtensa build with reference kernels.

njeffrie:
Previously, xtensa kernels which depended on nnlib were included by default even when building for reference kernels. This cauesed an include error due to missing nnlib headers.

TESTED=make -f tensorflow/lite/micro/tools/make/Makefile test_integration_tests_seanet_conv TARGET=xtensa TARGET_ARCH=hifi4 XTENSA_CORE=Fusion_F1 -j8

Test ensures that TFLM targets can be built / run on the xtensa ISS with reference kernels.

BUG=http://b/204814982


advaitjain (approved):

advaitjain:
can you update the PR description with the command that you ran to manually test that this PR is ok?

PR #676, opened 2021-11-11
Use template to generate BUILD and integrations_tests.cc in generate_per_layer_tests.py

deqiangc:
This change refactor generate_per_layer_tests.py to use
Mako to generate BUILD and integrations_tests.cc files, both
of which were generated directly from the code before this change.
Using Mako template simplify the generate_per_layer_tests.py
from 426 lines to 286 lines.

Tested = The generated files are compared against what the script
generated before the change and also bazel build :all 

BUG=http://b/205752554


njeffrie (approved):
Very cool!

njeffrie (approved):

PR #684, opened 2021-11-15
Change generate_cc_arrays.py to correctly dump multichannel WAV files.

shlmregev:
BUG=b/203458038

### DIFF
--- a/tensorflow/lite/micro/tools/generate_cc_arrays.py
+++ b/tensorflow/lite/micro/tools/generate_cc_arrays.py
@@ -21,6 +21,7 @@ from __future__ import print_function
 
 import argparse
 import os
+import struct
 import wave
 
 from PIL import Image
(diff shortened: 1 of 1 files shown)


njeffrie (dismissed):

tensorflow/lite/micro/tools/generate_cc_arrays.py
```diff
@@ -73,11 +74,13 @@ def generate_array(input_fname):
...
     wav_file = wave.open(input_fname, mode='r')
+    num_channels = wav_file.getnchannels()
     out_string = ''
     for _ in range(wav_file.getnframes()):
       frame = wav_file.readframes(1)
-      out_string += str(int.from_bytes(frame, byteorder='little',
-                                       signed=True)) + ','
+      samples = struct.unpack('<%dh' % num_channels, frame)
+      for sample in samples:
+        out_string += str(sample) + ','
```
advaitjain, line 83: This is a comment about the existing code, but it can be inefficient (https://google.github.io/styleguide/pyguide.html#310-strings)

Can we change to:
```py
n_frames = wav_file.getnframes()
frames = wav_file.readframes(n_frames)
samples = struct.unpack('<%dh' % (num_channels * n_frames), frames)
out_string = ','.join(map(str, samples))
```

advaitjain (changes requested):
one nit comment about exisitng code that would be nice to make more compliant with the Python style guide.

advaitjain (approved):

PR #686, opened 2021-11-15
Allow null inputs outputs in tflite flatbuffer

njeffrie:
Changes required to allow for null input and output arrays for subgraphs and operators.

Null inputs and outputs work on TFLite because they create the node TfLiteIntArrays using malloc even when the size is 0.

BUG=http://b/206152717

### DIFF
--- a/tensorflow/lite/micro/micro_allocator.cc
+++ b/tensorflow/lite/micro/micro_allocator.cc
@@ -25,14 +25,14 @@ limitations under the License.
 #include "tensorflow/lite/core/api/op_resolver.h"
 #include "tensorflow/lite/core/api/tensor_utils.h"
 #include "tensorflow/lite/kernels/internal/compatibility.h"
-#include "tensorflow/lite/schema/schema_generated.h"
-#include "tensorflow/lite/schema/schema_utils.h"
 #include "tensorflow/lite/micro/compatibility.h"
 #include "tensorflow/lite/micro/memory_helpers.h"
 #include "tensorflow/lite/micro/memory_planner/greedy_memory_planner.h"
 #include "tensorflow/lite/micro/memory_planner/micro_memory_planner.h"
 #include "tensorflow/lite/micro/micro_error_reporter.h"
 #include "tensorflow/lite/micro/simple_memory_allocator.h"
+#include "tensorflow/lite/schema/schema_generated.h"
+#include "tensorflow/lite/schema/schema_utils.h"
 
 namespace tflite {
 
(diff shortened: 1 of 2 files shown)


rockyrhodes (approved):

tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -300,6 +303,9 @@ TfLiteStatus FlatBufferVectorToTfLiteTypeArray(
     *result = const_cast<kTfLiteArrayType*>(
         reinterpret_cast<const kTfLiteArrayType*>(flatbuffer_array));
   } else {
+    // Fixed b/206152717 by removing this check for little endian. This check is
```
advaitjain, line 306: does this mean some models that work on little endian can not work on big endian? And there is no path to having support for a nullptr flatbuffer_array for big endian?

If that is the case, I would strongly suggest removing big endian support completely since this can be quite confusing.

https://github.com/tensorflow/tflite-micro/issues/304
https://github.com/tensorflow/tflite-micro/issues/455

I'm happy to make a separate PR that removes this support but wanted to first make sure that my understanding is correct as far as big-endian and this current PR is concerned.

tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -146,14 +146,16 @@ TfLiteStatus AllocationInfoBuilder::AddTensors(const SubGraph* subgraph,
 
   uint32_t operators_size = NumSubgraphOperators(subgraph);
 
-  for (size_t i = 0; i < subgraph->inputs()->size(); ++i) {
+  for (size_t i = 0;
+       subgraph->inputs() != nullptr && i < subgraph->inputs()->size(); ++i) {
```
advaitjain, line 150: is it the case that some of subgraph-> inputs, outputs, etc can be nullptr but not others? or will it be the case that either all will be nullptr or none of them?

If it is the latter, having a check in the begining of the function that tests for this specific condition (and a comment explaining under what conditions that can happen) might be more readable.

advaitjain (changes requested):
a few clarifying questions.

tensorflow/lite/micro/micro_allocator.cc
njeffrie: Yes that is correct. This would only be the case for models with zero-length input / output arrays (those using resource variables or CALL_ONCE).

I agree in principle, but I'm unsure what effects removing big endian support would have.

tensorflow/lite/micro/micro_allocator.cc
njeffrie: Yes, some subgraphs can have nonnull inputs / outputs. This is the case for SEANet.

tensorflow/lite/micro/micro_allocator.cc
njeffrie: Let's discuss removing big endian in a follow-on CL separately?

advaitjain (approved):

Issue #631, opened 2021-10-27
hello_world_model_data.h - a mysteriously appearing file?


ICTSoEasy:
Where does tensorflow/lite/micro/examples/hello_world/hello_world_model_data.h come from? 

I've been from pillar to post trying to get this to run on a pico - even the TinyML book has it just appearing!

Finfalter:
Run the [tests on your development machine](https://github.com/tensorflow/tflite-micro/tree/main/tensorflow/lite/micro/examples/hello_world#run-the-tests-on-a-development-machine)
```bash
make -f tensorflow/lite/micro/tools/make/Makefile third_party_downloads
make -f tensorflow/lite/micro/tools/make/Makefile test_hello_world_test
```
With these (the first line, actually) _**hello_world_model_data.cc**_ and _**hello_world_model_data.h**_ are generated during the built. They are stored in 
```bash
tensorflow/lite/micro/tools/make/gen/linux_x86_64_default/genfiles/tensorflow/lite/micro/examples/hello_world
```
This is at least the case for a 64-bit x86 Linux machine.

In `tensorflow/lite/micro/tools/make/Makefile`, lines 233 ff., there are some hints where the generated files are stored in a general case:
```bash
# Where compiled objects are stored.
GENDIR := $(MAKEFILE_DIR)/gen/$(TARGET)_$(TARGET_ARCH)_$(BUILD_TYPE)/
# [..]
GENERATED_SRCS_DIR := $(GENDIR)genfiles/
```

njeffrie:
@Finfalter is exactly correct.

We have been trying to move from examples with model_data.h and model_data.cc to a single model.tflite, and generate the cc and header during the build. If you're curious about the generation process, we use [this script](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/tools/generate_cc_arrays.py) to generate the cc and header files.

I agree w.r.t. improving documentation - I will look at including a section about data generation in our README.

PR #694, opened 2021-11-17
Add doc explaining data generation in TFLM examples

njeffrie:
BUG=Fixes https://github.com/tensorflow/tflite-micro/issues/631


advaitjain (approved):
It would be nice to consolidate a bunch of documentation like this into some type of examples / best practices guide. Let's continue documenting as issues come up and think about a more consolidated format later.

Issue #702, opened 2021-11-18
Add --rename-cc-to-cpp option to create_tflm_tree.py to assist with Micropython build


mocleiri:
Hello, 

I have switched my micropython tensorflow lite firmware to build all within micropython using the create_tflm_tree.py script to generate the files to build.

Previously  I was building the static library via the tflm Makefile and then linking it into the target firmware during the Micropython build.

Micropython supports C++ but it expects the files to have the **.cpp** suffix.  Unfortunately all of the files in tensorflow have the **.cc** suffix.

I got around this by adding a new switch **--rename-cc-to-cpp** that will as part of the generation of the files also take care of renaming them so that they end with .cpp as the Micropython build expects.

I'm successfully using this switch to generate files for stm32, rp2 and esp32:

- https://github.com/mocleiri/tensorflow-micropython-examples/blob/feature/31-support-stm32-boards/.github/workflows/build_esp32.yml#L71
- https://github.com/mocleiri/tensorflow-micropython-examples/blob/feature/31-support-stm32-boards/.github/workflows/build_stm32.yml#L61
- https://github.com/mocleiri/tensorflow-micropython-examples/blob/feature/31-support-stm32-boards/micropython-modules/microlite/prepare-tflm-rp2.sh#L3

I'd like to upstream this switch so I don't have to maintain my own fork with it.

mocleiri:
I'm closing this issue because the associated pull request has been merged.  Note that the option was changed to 
```
--rename_cc_to_cpp
```
in the final merged commit

PR #704, opened 2021-11-18
Add Support for building Micropython using project generator

mocleiri:
Add a new option **--rename-cc-to-cpp** to the create_tflm_tree.py script that is needed for building tflm within a Micropython build.

This is upstreaming of some changes I made to switch my Microlite Micropython firmware over to the new generator files approach from building separately and linking the static library in at firrmware build time.

My microlite Micropython module is built using the USER_C_MODULE micropython extension point so I am not able to easily change the core makefiles to support .cc files instead or in addition to .cpp files.

Since tflm has switched to generating the project files for incorporation into projects adding this switch on the tflm side makes sense. 

@advaitjain made some style changes to conform to the Google Python style as well as the shell style (https://google.github.io/styleguide/shellguide.html#s5.7-quoting)

BUG=#702

### DIFF
--- a/tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
+++ b/tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
@@ -17,35 +17,35 @@
 set -e
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-ROOT_DIR=${SCRIPT_DIR}/../../../../..
+ROOT_DIR="${SCRIPT_DIR}/../../../../.."
 cd "${ROOT_DIR}"
 
 source tensorflow/lite/micro/tools/ci_build/helper_functions.sh
 
 # First, we test that create_tflm_tree without any examples can be used to build a
 # static library.
-TEST_OUTPUT_DIR=$(mktemp -d)
+TEST_OUTPUT_DIR="$(mktemp -d)"
 
 readable_run \
   python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py \
-  ${TEST_OUTPUT_DIR}
+  "${TEST_OUTPUT_DIR}"
 
-readable_run cp tensorflow/lite/micro/tools/project_generation/Makefile ${TEST_OUTPUT_DIR}
-pushd ${TEST_OUTPUT_DIR} > /dev/null
+readable_run cp tensorflow/lite/micro/tools/project_generation/Makefile "${TEST_OUTPUT_DIR}"
+pushd "${TEST_OUTPUT_DIR}" > /dev/null
 readable_run make -j8 libtflm
 popd > /dev/null
 
-rm -rf ${TEST_OUTPUT_DIR}
+rm -rf "${TEST_OUTPUT_DIR}"
 
 # Next, we test that create_tflm_tree can be used to build example binaries.
 EXAMPLES="-e hello_world -e magic_wand -e micro_speech -e person_detection"
 
-TEST_OUTPUT_DIR=$(mktemp -d)
+TEST_OUTPUT_DIR="$(mktemp -d)"
 
 readable_run \
   python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py \
-  ${TEST_OUTPUT_DIR} \
-  ${EXAMPLES}
+  "${TEST_OUTPUT_DIR}" \
+  "${EXAMPLES}"
 
 # Confirm that print_src_files and print_dest_files output valid paths (and
 # nothing else).
@@ -54,51 +54,52 @@ FILES="$(python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree
            ${TEST_OUTPUT_DIR} \
            --print_src_files --print_dest_files --no_copy)"
 
-readable_run ls ${FILES} > /dev/null
+readable_run ls "${FILES}" > /dev/null
 
 # Next, make sure that the output tree has all the files needed buld the
 # examples.
-readable_run cp tensorflow/lite/micro/tools/project_generation/Makefile ${TEST_OUTPUT_DIR}
-pushd ${TEST_OUTPUT_DIR} > /dev/null
+readable_run cp tensorflow/lite/micro/tools/project_generation/Makefile "${TEST_OUTPUT_DIR}"
+pushd "${TEST_OUTPUT_DIR}" > /dev/null
 readable_run make -j8 examples
 popd > /dev/null
 
-rm -rf ${TEST_OUTPUT_DIR}
+rm -rf "${TEST_OUTPUT_DIR}"
 
 # Remove existing state prior to testing project generation for cortex-m target.
 make -f tensorflow/lite/micro/tools/make/Makefile clean clean_downloads
 
-TEST_OUTPUT_DIR_CMSIS=$(mktemp -d)
+TEST_OUTPUT_DIR_CMSIS="$(mktemp -d)"
 
 readable_run \
   python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py \
   --makefile_options="TARGET=cortex_m_generic OPTIMIZED_KERNEL_DIR=cmsis_nn TARGET_ARCH=project_generation" \
-  ${TEST_OUTPUT_DIR_CMSIS} \
-  ${EXAMPLES}
+  "${TEST_OUTPUT_DIR_CMSIS}" \
+  "${EXAMPLES}"
 
-readable_run cp tensorflow/lite/micro/tools/project_generation/Makefile ${TEST_OUTPUT_DIR_CMSIS}
+readable_run \
+  cp tensorflow/lite/micro/tools/project_generation/Makefile "${TEST_OUTPUT_DIR_CMSIS}"
 
-pushd ${TEST_OUTPUT_DIR_CMSIS} > /dev/null
+pushd "${TEST_OUTPUT_DIR_CMSIS}" > /dev/null
 
-PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded/bin \
+PATH="${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded/bin" \
   readable_run \
   make -j8 BUILD_TYPE=cmsis_nn
 
 popd > /dev/null
 
-rm -rf ${TEST_OUTPUT_DIR_CMSIS}
+rm -rf "${TEST_OUTPUT_DIR_CMSIS}"
 
 # Test that C++ files are renamed to .cpp
 
-TEST_OUTPUT_DIR_RENAME_CC=$(mktemp -d)
+TEST_OUTPUT_DIR_RENAME_CC="$(mktemp -d)"
 
 readable_run \
   python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py \
-  --rename-cc-to-cpp \
-  ${TEST_OUTPUT_DIR_RENAME_CC}
+  --rename_cc_to_cpp \
+  "${TEST_OUTPUT_DIR_RENAME_CC}"
 
-CC_FILES=$(find ${TEST_OUTPUT_DIR_RENAME_CC} -name "*.cc" | head)
-CPP_FILES=$(find ${TEST_OUTPUT_DIR_RENAME_CC} -name "*.cpp" | head)
+CC_FILES="$(find ${TEST_OUTPUT_DIR_RENAME_CC} -name "*.cc" | head)"
+CPP_FILES="$(find ${TEST_OUTPUT_DIR_RENAME_CC} -name "*.cpp" | head)"
 
 if test -n "${CC_FILES}"; then
   echo "Expected no .cc file to exist"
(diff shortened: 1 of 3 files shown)


tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -87,3 +87,33 @@ PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded
...
+
+readable_run \
+  python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py \
+  --makefile_options="TARGET=cortex_m_generic OPTIMIZED_KERNEL_DIR=cmsis_nn TARGET_ARCH=project_generation" \
+  --rename-cc-to-cpp \
+  ${TEST_OUTPUT_DIR_CMSIS} \
+  ${EXAMPLES}
+
+CC_FILES=$(find ${TEST_OUTPUT_DIR_CMSIS} -name "*.cc" | head)
+CPP_FILES=$(find ${TEST_OUTPUT_DIR_CMSIS} -name "*.cpp" | head)
```
mocleiri, line 107: This looks for any .cc and any .cpp files and then extracts the first one.

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -87,3 +87,33 @@ PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded
...
+
+if test -n "$CC_FILES"; then
+  echo "Expected no .cc file to exist"
+  exit 1;
+fi
+
+if test -z "$CPP_FILES"; then
+  echo "Expected a .cpp file to exist"
+  exit 1;
+fi
```
mocleiri, line 119: I wanted to do a test on the existence of the file but couldn't get the syntax quite right.  Instead the first one tests if there is any value for the .cc files search (it errors out if there are any).  The second one tests if there is no value in the .cpp files search and errors out if there are no .cpp files.

petewarden (approved):

advaitjain (changes requested):
My preference would be to have a separate and standalone function that walks the entire output tree and renames files. The fewer boolean options to functions, the more maintainable the overall script.

And let's use pathlib or os to do the renaming, similar to https://stackoverflow.com/q/2900035

```py
def _rename_cc_to_cpp(output_dir):
  # walk the entire output_dir and rename all the .cc fles to .cpp
```

And then the call would be:
```
  if args.no_copy is False:
    _copy(src_files, dest_files)

  if args.examples is not None:
    _create_examples_tree(args.output_dir, args.examples)

  if args.rename_cc_to_cpp:
    _rename_cc_to_cpp(args.output_dir)

```

mocleiri:
Thanks for your feedback.

I'll rework my changes to take the approach you are suggesting

mocleiri:
@advaitjain I've reworked my changes following your recommendations and also rebased onto the latest main branch.

tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py
```diff
@@ -179,6 +179,13 @@ def _create_examples_tree(prefix_dir, examples_list):
         # in-place find and replace.
         print(line, end="")
 
+def _rename_cc_to_cpp (output_dir):
+  for path, subdirs, files in os.walk(output_dir):
+    for name in files:
+      if fnmatch(name, "*.cc"):
+        baseName, extension = os.path.splitext(name)
+        os.rename (path + '/' + name, path + '/' + baseName + ".cpp")
+
```
mocleiri, line 188: Find all *.cc files in output_dir and then rename them to baseName.cpp

tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py
```diff
@@ -35,6 +35,7 @@
 import re
 import shutil
 import subprocess
+from fnmatch import fnmatch
```
mocleiri, line 39: Used for selecting only the *.cc files.

tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py
mocleiri: This whole block to find all of the files in a subdirectory came from this example: 
https://stackoverflow.com/a/13214966

tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py
```diff
@@ -179,6 +179,13 @@ def _create_examples_tree(prefix_dir, examples_list):
         # in-place find and replace.
         print(line, end="")
 
+def _rename_cc_to_cpp (output_dir):
+  for path, subdirs, files in os.walk(output_dir):
+    for name in files:
+      if fnmatch(name, "*.cc"):
```
advaitjain, line 185: let's make it simpler and use `name.endswith(".cc")` instead of fnmatch.

tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py
```diff
@@ -206,6 +213,11 @@ def main():
                       action="append",
                       help="Examples to add to the output tree. For example: "
                       "-e hello_world -e micro_speech")
+  parser.add_argument("--rename-cc-to-cpp",
+                      "-r",
+                      action="store_true",
+                      help="Rename all .cc files to .cpp in the destination files location.  Used for micropython" +
+                      " which only builds .cpp files.")
```
advaitjain, line 220: delete "Used for micropython". There may be other uses.

tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py
```diff
@@ -206,6 +213,11 @@ def main():
                       action="append",
                       help="Examples to add to the output tree. For example: "
                       "-e hello_world -e micro_speech")
+  parser.add_argument("--rename-cc-to-cpp",
+                      "-r",
```
advaitjain, line 217: please remove `-r`. It isn't descriptive enough.

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -87,3 +87,33 @@ PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded
...
 
 #rm -rf ${TEST_OUTPUT_DIR_CMSIS}
+
+# Test that C++ files are renamed to .cpp
+rm -rf ${TEST_OUTPUT_DIR_CMSIS}
+
+# Remove existing state prior to testing project generation for cortex-m target.
+# make -f tensorflow/lite/micro/tools/make/Makefile clean clean_downloads
+
+TEST_OUTPUT_DIR_CMSIS=$(mktemp -d)
```
advaitjain, line 97: let's have a different variable here. TEST_OUTPUT_DIR_RENAME_CC?

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -87,3 +87,33 @@ PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded
 popd > /dev/null
 
 #rm -rf ${TEST_OUTPUT_DIR_CMSIS}
+
+# Test that C++ files are renamed to .cpp
+rm -rf ${TEST_OUTPUT_DIR_CMSIS}
```
advaitjain, line 92: the commented out like 89 seems to be an oversight. Let's move this rm to the CMSIS block and add a new output dir that tests the file renaming.

```
#rm -rf ${TEST_OUTPUT_DIR_CMSIS}

# Test that C++ files are renamed to .cpp
rm -rf ${TEST_OUTPUT_DIR_CMSIS}
```

should become:
```
rm -rf ${TEST_OUTPUT_DIR_CMSIS}

# Test that C++ files can be renamed to .cpp
```

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -87,3 +87,33 @@ PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded
...
+
+# Remove existing state prior to testing project generation for cortex-m target.
+# make -f tensorflow/lite/micro/tools/make/Makefile clean clean_downloads
```
advaitjain, line 95: remove commented out lines

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -87,3 +87,33 @@ PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded
...
+
+TEST_OUTPUT_DIR_CMSIS=$(mktemp -d)
+
+readable_run \
+  python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py \
+  --makefile_options="TARGET=cortex_m_generic OPTIMIZED_KERNEL_DIR=cmsis_nn TARGET_ARCH=project_generation" \
```
advaitjain, line 101: my preference would be to have this not use any additional makefile_options, i.e. use

```
readable_run \
  python3 tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py \
  ${TEST_OUTPUT_DIR_RENAME_CC}
```

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -87,3 +87,33 @@ PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded
...
+  --rename-cc-to-cpp \
+  ${TEST_OUTPUT_DIR_CMSIS} \
+  ${EXAMPLES}
+
+CC_FILES=$(find ${TEST_OUTPUT_DIR_CMSIS} -name "*.cc" | head)
+CPP_FILES=$(find ${TEST_OUTPUT_DIR_CMSIS} -name "*.cpp" | head)
+
+# echo "CC_FILE='$CC_FILES', CPP_FILE='$CPP_FILES'"
```
advaitjain, line 109: remove commented out line

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -87,3 +87,33 @@ PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded
...
+
+if test -n "$CC_FILES"; then
+  echo "Expected no .cc file to exist"
+  exit 1;
```
advaitjain, line 113: also log the ${CC_FILES} so that the error message is more actionable.

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -87,3 +87,33 @@ PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded
...
+  --rename-cc-to-cpp \
+  ${TEST_OUTPUT_DIR_CMSIS} \
+  ${EXAMPLES}
+
+CC_FILES=$(find ${TEST_OUTPUT_DIR_CMSIS} -name "*.cc" | head)
+CPP_FILES=$(find ${TEST_OUTPUT_DIR_CMSIS} -name "*.cpp" | head)
+
+# echo "CC_FILE='$CC_FILES', CPP_FILE='$CPP_FILES'"
+
+if test -n "$CC_FILES"; then
```
advaitjain, line 111: `"${CC_FILES}"`

https://google.github.io/styleguide/shellguide.html#variable-expansion

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
```diff
@@ -87,3 +87,33 @@ PATH=${PATH}:${ROOT_DIR}/tensorflow/lite/micro/tools/make/downloads/gcc_embedded
...
+  echo "Expected no .cc file to exist"
+  exit 1;
+fi
+
+if test -z "$CPP_FILES"; then
```
advaitjain, line 116: `"${CPP_FILES}"`

https://google.github.io/styleguide/shellguide.html#variable-expansion

advaitjain (changes requested):
Thanks for the updated PR.

A few more small nitpick comments. We can merge this PR once they are addressed.

mocleiri:
@advaitjain Thanks for your detailed review.  

I've made all of the changes you identified. 

 I verified the test still passes before force pushing the updated commit just now.

mocleiri:
@advaitjain I've addressed you comments that I missed in my first try.

Please advise if you see anything else I missed.  I'm happy to resubmit if you see anything.

tensorflow/lite/micro/tools/ci_build/test_project_generation.sh
mocleiri: The above diff isn't showing the change but I implemented it as you said with the rm for the CMSIS directory being uncommented and then leading into the new test block for the renaming feature.

Can be seen in the latest commit from here: https://github.com/tensorflow/tflite-micro/commit/768fb9f7997d1fddf8c3bdadd48acf5e4c0225b5#diff-227b53ea64f9bfbf4262fdcbd7fee872cc32deb19774f0a947850db6828080deR89

advaitjain (approved):

advaitjain:
I push some updates to match the google style etc. with [`6ff229a` (#704)](https://github.com/tensorflow/tflite-micro/pull/704/commits/6ff229aee30b413bcfecdc7ec4f41419724001c6). Note that the parameter is  now `--rename_cc_to_cpp` instead of `--rename-cc-to-cpp`

advaitjain (approved):

advaitjain (approved):

PR #712, opened 2021-11-20
Fix Unused Operator Output Buffer Sharing

njeffrie:
Handle case where operator outputs are unused, but written to from within the op.

TESTED=Ran various end-to-end tests and verified audio artifacts were gone.

BUG=http://b/205145012#comment15


shlmregev (approved):
I'm not very familiar with the memory planner. Maybe add diqiang as a reviewer?

advaitjain (approved):

PR #713, opened 2021-11-22
adding mergify config for jobs in ci.yml workflow

jwithers:
Basic mergify config for testing. Only includes the jobs in the ci.yml workflow: Bazel, Cortex-M, Code Style, Project Generation and Makefile x86 (all (presubmit)).

Requires a ci:mergify label on the pr to activate for purposes of testing. Also does not include a requirement for at least one approved review, which is likely something we want doing this in anger. 

BUG=none, evaluating a ci internal change

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,11 @@
+pull_request_rules:
+  - name: Automatic merge on approval
+    conditions:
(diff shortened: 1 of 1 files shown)


jwithers:
Have added ci:mergify to the project label list.

advaitjain (approved):

Issue #637, opened 2021-10-29
ARC target transition into platform specific repository.


gerbauz:
This issue will be used to track changes related to ARC specific files to meet requirements for moving hardware platform support and platform-specific customization of the examples out of the tflite-micro repository.

Related issue:
* [Move hardware / platform support to repositories that are external to tflite-micro #407](https://github.com/tensorflow/tflite-micro/issues/407)

advaitjain:
Thanks for the heads up. Yes, we can help you make the transition prior to removing the ARC-specific code.

Having make/targets/arc is fine, with the caveat that the Makefile-based project generation will be removed. So any functionality in the tools/make/arc directory that depends on the Makefile-based project generation will have to be implemented differently.

Happy to discuss what the alternatives would be.

Ideally, there is a Linux tool chain that can be used to build and run the kernel tests without requiring any project generation, and only Makefiles needed for that are in the TFLM repository.

And everything that is not required for this Linux tool chain is in the ARC-specific examples repository.

gerbauz:
Thank you for the answer.

> Ideally, there is a Linux tool chain that can be used to build and run the kernel tests without requiring any project generation, and only Makefiles needed for that are in the TFLM repository.
> 
> And everything that is not required for this Linux tool chain is in the ARC-specific examples repository.

Also, based on your [answer](https://github.com/tensorflow/tflite-micro/issues/407#issuecomment-956577367), for now, as we are not able to build using only Linux toolchain or using some options like docker image, we'll have to wait until external build systems will be integrated and only after that we can finally support kernel tests, right?

gerbauz:
Hi @advaitjain,

As I see, xtensa using docker to run their tools, don't they?
https://github.com/tensorflow/tflite-micro/blob/1a23ad6adf007ad6ef3826523f41ff43fb686a13/.github/workflows/xtensa.yml#L38-L39

Can we look for same support for ARC target?

gerbauz:
Hi @advaitjain,

Still interested in using docker for ARC tests. What do you think?

Thank you.

PR #737, opened 2021-11-25
Removing ARC code related to examples and updating make files and test scripts.

gerbauz:
This PR removes ARC unused files (probably temporary) and contains some scripts to run tests for ARC target.

BUG=#637


gerbauz:
Updated.

deqiangc (commented):
@advaitjain 

This PR deletes the test in kernels such as tensorflow/micro/kernels/arc_mili/conv_slicing_test.cc.   Although this kind of test is not commonly seen and encouraged, I am not sure of removing them either.  I will defer to @advaitjain  for a second pair of eyes/thoughts.

gerbauz:
> This PR deletes the test in kernels such as tensorflow/micro/kernels/arc_mili/conv_slicing_test.cc. Although this kind of test is not commonly seen and encouraged, I am not sure of removing them either. I will defer to @advaitjain for a second pair of eyes/thoughts.

Yes, this are tests specific to ARC slicing code. They haven't been updated for a long time, so we decided to remove them and maybe return back to them later.

gerbauz:
Updated.

gerbauz:
Updated.

advaitjain (approved):

Issue #706, opened 2021-11-19
Null pointer bias in convolution kernels is not working


keandersslab:
Most convolution kernels (conv/depthwise conv/transpose conv) support nullptr bias tensors.
Eval functions prepare a const TfLiteEvalTensor* bias tensor which is later fed to tflite::micro::GetTensorData<int32_t>(bias)
This will print "HALTED" in console and inference is effectively stopped.

This can be solved by substituting this for the GetTensorData() call:
tflite::micro::GetTensorData<int32_t>(const_cast<TfLiteEvalTensor*>(bias))

PR #738, opened 2021-11-25
Convolution kernel wrappers now accept nullptr bias tensors.

keandersslab:
This solves https://github.com/tensorflow/tflite-micro/issues/706


keandersslab:
Time for review soon ?

kjeir:
Hi.
This PR has been in review for 5 months now. Does somebody have an oppurtunity to have a look ? 
Thanks

paulinesho:
Hi @kjeir the team discussed this issue today and preferred to go with this approach instead: https://github.com/tensorflow/tflite-micro/pull/1114. We think that using `GetOptionalTensorData` introduced in that PR will make it more explicit about what behavior is intended and hence more readable. Let me know if we can close this or if there's any other issue.

keandersslab:
I agree that #1114 is a better solution.
I will close this PR

keandersslab:
Solved by #1114

PR #739, opened 2021-11-25
Quantize operator:

cad-audio:
Updated the quantize kernel to support all the data types similar to reference.
uncommented the test cases in quantize_test.

### DIFF
--- a/tensorflow/lite/micro/kernels/quantize_test.cc
+++ b/tensorflow/lite/micro/kernels/quantize_test.cc
@@ -49,7 +49,6 @@ void ValidateQuantizeGoldens(TfLiteTensor* tensors, int tensors_size,
   }
 }
 
-#if !defined(XTENSA)
 template <typename T>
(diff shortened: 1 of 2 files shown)


njeffrie (changes requested):
Would it be possible to reuse the EvalQuantizeReference method rather than re-creating the logic in the xtensa quantize operator?

freddan80:
@njeffrie @cad-audio would it be possible to add [uint8 support](https://github.com/tensorflow/tensorflow/blob/40670552faa92fb353bd0e4e5efed26979f239c5/tensorflow/lite/kernels/quantize.cc#L316) for this one op as an "exception"? We have received requests from customers for this feature since sensor data, e.g. images, usually comes as uint from the driver. That would remove the need to re-quantize in the app.

freddan80:
Sorry for scope creeping this ticket... But I thought it was a good time to ask :)

advaitjain:
Closing old PR. Let's reopen if needed.

Issue #741, opened 2021-11-26
Port MIRROR_PAD from TFlite


mansnils:
This issue tracks my work porting operator MIRROR_PAD from Tflite.

The port will be submitted in a number of PRs according to https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/porting_reference_ops.md - although PR2 won't be needed since the operator is not using anything from reference_ops.h.

beekeeper23:
Thank you for your contribution! Is there any update or any timeline on your work?

mansnils:
Hi @beekeeper23,
It is up for review now: https://github.com/tensorflow/tflite-micro/pull/779

beekeeper23:
Perfect, thank you!

PR #742, opened 2021-11-29
Extract a function for parsing operator MIRROR_PAD

mansnils:
Extract the parsing of op parameters from the flatbuffer out of a
switch statement case, into a standalone function which can be
called by the micro op resolver.

This PR is part of the work to port operator MIRROR_PAD
from TFL to TFLM, as tracked in issue https://github.com/tensorflow/tflite-micro/issues/741.


deqiangc (changes requested):
@advaitjain 

I feel this PR is in the wrong Repo. This change (which is PR1 in main/tensorflow/lite/micro/docs/porting_reference_ops.md) should be in tensorflow main repo and then sync to tflite-micro.

Add the teammate in case I misunderstand the purpose/context of this PR.

mansnils:
Hi @deqiangc 
I think you are right, I am closing this one.

Issue #743, opened 2021-11-29
TRANSPOSE CONV not really supporting dilation


mansnils:
Parameters params->dilation_height_factor and params->dilation_width_factor have been seen to return random numbers and they are always set to 1 in the TFL equivalent operator, which means it won't be bit-exact towards TFL. So it should always be 1 in TFLM as well.

PR #744, opened 2021-11-29
Update TRANSPOSE_CONV

mansnils:
* Removes dilation (set to 1).
* Corrects input height to pad calculation.

BUG =  https://github.com/tensorflow/tflite-micro/issues/743

### DIFF
--- a/tensorflow/lite/micro/kernels/transpose_conv.cc
+++ b/tensorflow/lite/micro/kernels/transpose_conv.cc
@@ -1,4 +1,4 @@
-/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


deqiangc (changes requested):
@advaitjain @njeffrie 

It seems to me that this change would pass test due to the underlying reference transpose_conv implementation assume dilation = 1 (not using dilation parameter). However, it is less clear to me that whether that is just because that is not yet implemented or a intentional choice.  If it is former,
I would like to have a 
TFLITE_CHECK_EQ(params->dilation_height_factor ,1)  instead of this change.

At the very least, I would like the constant 1 not to be hard coded, but use
constexpr int kDilationHeightFactor = 1;
constexpr int kDilationWidthFactor = 1;
and document the reasons why we hard code these numbers.  Thanks!

I add two teammates that may be more familiar on this in the loop to correct me.

advaitjain:
I think we need to combine the fixes from https://github.com/tensorflow/tflite-micro/pull/744 and https://github.com/tensorflow/tflite-micro/pull/736. If possible, it would also be very useful to have some unit tests that fail without these PRs but pass with them.

@njeffrie, would you be able to take that on?

njeffrie:
Created #762 to address these change along with the changes in #736

PR #751, opened 2021-11-30
For both MicroAllocator and RecordingMicroAllocator, define function to return the default tail usage.

deqiangc:
For both MicroAllocator and RecordingMicroAllocator, define function to return 
   the default tail usage. 
    Some original expression only account for memory planner size and this
    change account for micro allocator as well.
    This is another baby step for cleaning up magic number that make test
    brittle.
    
BUG=http://b/207157610


tensorflow/lite/micro/recording_micro_allocator.cc
```diff
@@ -18,12 +18,29 @@ limitations under the License.
...
 namespace tflite {
 
+size_t RecordingMicroAllocator::GetDefaultTailUsage() {
+  // RecordingMicroAllocator inherits from MicroAllocator and its tail usage is
+  // similar with MicroAllocator with SimpleMemoryAllocator and MicroAllocator
+  // being replaced.
+  return MicroAllocator::GetDefaultTailUsage(
+             /*is_memory_planner_given=*/false) +
+         AlignSizeUp(sizeof(RecordingSimpleMemoryAllocator),
+                     alignof(RecordingSimpleMemoryAllocator)) -
```
advaitjain, line 36: What do you think of adding a  templated AlignSizeUp

```cc
template <typename T>
size_t AlignSizeUp() {
  AlignSizeUp(sizeof(T), alignof(T));
}
```

That would make this expression easier to read:

```
 return AlignSizeUp<RecordingSimpleMemoryAllocator>() - AlignSizeUp<SimpleMemoryAllocator>() +
         AlignSizeUp<RecordingMicroAllocator>() -      AlignSizeUp<MicroAllocator>()

**Review by advaitjain — CHANGES_REQUESTED** [review]

File: tensorflow/lite/micro/recording_micro_allocator.h [hunk 2]
```diff
@@ -58,6 +58,11 @@ class RecordingMicroAllocator : public MicroAllocator {
                                          size_t arena_size,
                                          ErrorReporter* error_reporter);
 
+  // Returns the fixed amount of memory overhead of RecordingMicroAllocator. You
+  // can image this number as the arena usage after RecordingMicroAllocator is
```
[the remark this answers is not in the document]
deqiangc: Done

tensorflow/lite/micro/recording_micro_allocator.cc
deqiangc: Create a bug http://b/208703041 and add it into the TODO. Prefer another PR for it to keep the change scope smaller in this PR.

advaitjain (approved):

PR #755, opened 2021-12-02
Add integration test generation support for multiple inputs and types

njeffrie:
Support generating layer tests for more operators {transpose_conv, fully_connected, add, sub, strided_slice, depthwise_conv2d and leaky_relu} as well as int8, int16, int32 and float input and output data types.

BUG=201585693

TESTED=
For op in (conv, transpose_conv, add, sub, strided_slice, leaky_relu):
```
bazel run //tensorflow/lite/micro/integration_tests:generate_per_layer_tests -- --input_tflite_file=/tmp/seanet.tflite --output_dir=/home/njeff/tflite_micro/tensorflow/lite/micro/integration_tests/seanet/<op>
bazel run tensorflow/lite/micro/integration_tests/seanet/<op>:integration_test --verbose_failures
make -f tensorflow/lite/micro/tools/make/Makefile test_integration_tests_seanet_<op>_test
```


tensorflow/lite/micro/integration_tests/generate_per_layer_tests.py
```diff
@@ -143,43 +141,106 @@ def __init__(self, output_dir, model_paths, inputs):
...
+                          high=np.iinfo(dtype).max,
+                          dtype=dtype,
+                          size=input_tensor0().shape),
+        np.random.randint(low=np.iinfo(dtype).min,
+                          high=np.iinfo(dtype).max,
+                          dtype=dtype,
+                          size=input_tensor1().shape)
+    ]
+
+  def generate_inputs_transpose_conv(self, interpreter, dtype):
```
deqiangc, line 175: All these generate_inputs_xxx function seems to be generating input tensors according to each OP's specific interface.  I imagine these will keep on growing and wonder whether this would be easy to maintain. WDYT of refactoring this to be a single GenerateInputForOp class in a sperate file and takes op and interpreter and so on?

deqiangc (commented):
Could you also document the command and steps that you use to verify this script is working? 

On a related note, if this is not a one-time script for one mode, I wonder what your thoughts are in terms of script testing itself. I am on the fence on whether we need to have some tests for the script that generate test:), just wondering how to maintain this long term.

deqiangc (changes requested):
I am still shamefully not familiar with github tool.  So double comments
Could you also document the command and steps that you use to verify this script is working?

advaitjain:
> Could you also document the command and steps that you use to verify this script is working?
> 
> On a related note, if this is not a one-time script for one mode, I wonder what your thoughts are in terms of script testing itself. I am on the fence on whether we need to have some tests for the script that generate test:), just wondering how to maintain this long term.

+1 to figuring out how to test this script. All we are doing today is to ensure that we can `bazel build` the `py_binary`. Let's figure out some options in http://b/205042963

tensorflow/lite/micro/integration_tests/generate_per_layer_tests.py
njeffrie: My hope is that we can mostly share implementations between ops. For example, all single-input single-output ops can share one implementation, most 2-intput 1 output ops share (add, sub, mul, div, etc) and only a few special ops need a specific impl (transpose_conv).

I'll move this into a separate class / file like you suggested.

njeffrie:
Added steps used to test this script.

My feeling is that an integration test will look like a script which runs the commands I used to test the model on several different models, for several different ops.

I expect that this will not work perfectly for new models and ops - it's a convenient utility designed to be modified and extended for new models.

deqiangc (dismissed):

njeffrie:
I added a presubmit test which generates conv layers for person detection and runs both the bazel and make targets generated.

deqiangc (commented):
Do you mind elaborate a bit on why this test is not put in under test_x86.sh and therefore gets automatically coverage instead of a new workflow?  I heard from @advaitjain that each new workflow starts a new process, checkout build etc and take some time and our presubmit is already a bit slow.

njeffrie:
Moved test to separate workflow as per discussion

deqiangc (approved):

advaitjain (approved):

PR #768, opened 2021-12-09
Upload the script that geneartes the simple_add_model.tflite in models

deqiangc:
Add a script that creates a simple tflite model with just ADD operation.

BUG=http://b/210896418


tensorflow/lite/micro/examples/memory_footprint/create_adder_model.py
```diff
@@ -0,0 +1,51 @@
...
+from absl import app
+import tensorflow as tf
+
+
+def main(_):
+  input_shape = (1, 1, 1)
+  x1 = tf.keras.layers.Input(input_shape)
+  x2 = tf.keras.layers.Input(input_shape)
+
+  # equivalent to `added = tf.keras.layers.add([x1, x2])`
```
advaitjain, line 26: remove comment?

tensorflow/lite/micro/examples/memory_footprint/create_adder_model.py
```diff
@@ -0,0 +1,51 @@
...
+  added = tf.keras.layers.Add()([x1, x2])
+  model = tf.keras.models.Model(inputs=[x1, x2], outputs=added)
+  model.save("simple_add_model")
+
+  converter = tf.lite.TFLiteConverter.from_saved_model("simple_add_model")
```
advaitjain, line 31: do we need to go to a saved_model? can we directly go from keras to tflite, similar to https://github.com/tensorflow/tflite-micro/blob/ca14acc51a92912da5b18da3f97391d773842925/tensorflow/lite/micro/testing/generate_test_models.py#L64-L65

tensorflow/lite/micro/examples/memory_footprint/create_adder_model.py
```diff
@@ -0,0 +1,51 @@
...
+  # Enforce integer only quantization
+  converter.target_spec.supported_ops = [tf.lite.OpsSet.TFLITE_BUILTINS_INT8]
+  converter.inference_input_type = tf.int8
+  converter.inference_output_type = tf.int8
+
+  # Convert the model to the TensorFlow Lite format with quantization and
+  # quantization requires a representative data set
+  def representative_dataset():
+    for i in range(500):
+      yield ([tf.random.normal(input_shape), tf.random.normal(input_shape)])
```
advaitjain, line 42: let's set both the random and operation-level seed so that the output is reproducible.

Copied the following from `tf.random.normal` docstring:
```
Outputs random values from a normal distribution.

Example that generates a new set of random values every time:

>>> tf.random.set_seed(5);
>>> tf.random.normal([4], 0, 1, tf.float32)
<tf.Tensor: shape=(4,), dtype=float32, numpy=..., dtype=float32)>

Example that outputs a reproducible result:

>>> tf.random.set_seed(5);
>>> tf.random.normal([2,2], 0, 1, tf.float32, seed=1)
<tf.Tensor: shape=(2, 2), dtype=float32, numpy=
array([[-1.3768897 , -0.01258316],
      [-0.169515   ,  1.0824056 ]], dtype=float32)>

In this case, we are setting both the global and operation-level seed to
ensure this result is reproducible.  See `tf.random.set_seed` for more
information.
```

I'd be interested in a discussion about what tests we can put around such funcitonality. For example, maybe we want a Python test that confirms that the model created with this script is the model that is currently checked into the repository?

advaitjain (commented):
happy to talk more about adding some tests etc. so that whatever pattern we specify in this PR can be used for future scripts as well.

tensorflow/lite/micro/examples/memory_footprint/create_adder_model.py
deqiangc: done

tensorflow/lite/micro/examples/memory_footprint/create_adder_model.py
deqiangc: Done. Thanks for pointing this to me!

tensorflow/lite/micro/examples/memory_footprint/create_adder_model.py
deqiangc: Done  with fixing the random seed. Thanks for pointing out to me! 

It is a really good point about how to verify this. As discussed offline, this is not straightforward and we can discuss further in b/210896418.

Neverthess, in this PR, I updated the model using the new script, which is reproducible and also use a more realistic tensor size.

advaitjain (approved):

Issue #772, opened 2021-12-10
Enable conv optimization in kernel


tingyan19:
Update kernel code to enable conv optimization. Complete optimization will need support nnlib, and will come at a later time. 

At this point, related nnlib functions are stubbed.

Issue #638, opened 2021-10-29
CMSIS-NN third party library builds with -Os flag.


jenselofsson:
When building with OPTIMIZED_KERNEL_DIR=cmsis_nn, the glue code in tensorflow/lite/micro/kernels/cmsis_nn is built with the KERNEL_OPTIMIZATION_LEVEL=-O2.

The CMSIS-NN library are built as part of the THIRD_PARTY_OBJS, which is uses the optimization flag -Os (CORE_OPTIMIZATION_FLAG). It would be preferable to be able to specify a custom optimization flag for a third party library, since at least CMSIS-NN would see a performance increase when being compiled with for example -O3.

My initial thought is to add a CMSIS_NN_OPTIMIZATION_FLAG (or something to that effect) to [cmsis_nn.inc](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc) which is then used in the Makefile, and to split the CMSIS-NN header-/source-files from the THIRD_PARTY_CC_HDRS/SRCS list into its own list.

But that solution would not scale well as the number of third party libraries and custom flags would increase, so it might be preferable to have a more complex, but scalable solution.

jenselofsson:
@advaitjain @petewarden Do you have any thoughts on this?

advaitjain:
Sorry for the delay here, we'll talk internally and get back to you on this.

Cross-linking https://github.com/tensorflow/tflite-micro/pull/774#discussion_r767025930

PR #774, opened 2021-12-10
Enable optimization for conv2d in HiFi4 (#772)

tingyan19:
Apply patches to enable optimization for conv2d in HiFi4

BUG=#772

TESTED=

docker run --rm -v "$(pwd)":/opt/tflite-micro ghcr.io/tflm-bot/xtensa:0.1 /opt/tflite-micro/tensorflow/lite/micro/tools/ci_build/test_xtensa_hifi3z.sh


tensorflow/lite/micro/kernels/xtensa/conv_hifi.cc
```diff
@@ -54,23 +54,142 @@ TfLiteStatus ConvPrepareHifi(TfLiteContext* context, TfLiteNode* node) {
...
+                            const TfLiteEvalTensor* filter,
+                            const TfLiteEvalTensor* bias,
+                            TfLiteEvalTensor* output) {
+  const RuntimeShape& input_shape = tflite::micro::GetTensorShape(input);
+  const RuntimeShape& filter_shape = tflite::micro::GetTensorShape(filter);
+  /* Dilation is currently not supported on HiFi 4 NN Library */
+  if ((params.dilation_width_factor == 1) &&
+      (params.dilation_height_factor == 1) &&
+      input_shape.Dims(1) >= filter_shape.Dims(1) &&
+      input_shape.Dims(2) >= filter_shape.Dims(2)) {
```
advaitjain, line 93: this if-else block is quite confusing to read.

The else block has a return statement and then there is a call to the reference kernel after the else block. Refactoring to something like the following might help:

```cc
if (input_shape.Dims(1) < filter_shape.Dims(1)  || input_shape.Dims(2) >= filter_shape.Dims(2)) {
// first set of calls to nnlib
return
}

// Dilation not supported in nnlib
if ((params.dilation_width_factor == 1) &&
      (params.dilation_height_factor == 1) &&
      input_shape.Dims(1) >= filter_shape.Dims(1) &&
      input_shape.Dims(2) >= filter_shape.Dims(2)) {
 // calls to nnlib
  return;
}

// Use the reference kernels
}
```

I would also like know if the nnlib call when filter is large than the input is necessary, and whether that supports dilation or not.

tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -112,6 +112,7 @@ endif
 
 CORE_OPTIMIZATION_LEVEL := -Os
 KERNEL_OPTIMIZATION_LEVEL := -O2
+THIRD_PARTY_C_OPTIMIZATION_LEVEL := -Os
```
advaitjain, line 115: Let's keep this change for a separate PR. I think we will want to be careful here since we might want some 3P libraries with -Os (e.g. flatbuffers) and others with -O2 (CMSIS, nnlib). Perhaps make a bug and we can discuss further?

Related bug: http://b/161489252 and https://github.com/tensorflow/tflite-micro/issues/638

tensorflow/lite/micro/tools/make/targets/xtensa_makefile.inc
```diff
@@ -34,6 +34,8 @@ endif
 #   TARGET_ARCH=hifi4 --> -DHIFI4
 TARGET_ARCH_DEFINES := -D$(shell echo $(TARGET_ARCH) | tr [a-z] [A-Z])
 
+THIRD_PARTY_C_OPTIMIZATION_LEVEL := -O3 -LNO:simd
```
advaitjain, line 37: let's keep this change for a separate PR

advaitjain (changes requested):

tensorflow/lite/micro/tools/make/targets/xtensa_makefile.inc
tingyan19: Rolled back in commit https://github.com/tensorflow/tflite-micro/pull/774/commits/90c7ba043d30dc2879e5bb82e8e62c34f0be4418

and created bug: 

http://b/210160638

tensorflow/lite/micro/tools/make/Makefile
tingyan19: Rolled back in commit https://github.com/tensorflow/tflite-micro/commit/90c7ba043d30dc2879e5bb82e8e62c34f0be4418

and created bug:

http://b/210160638

tensorflow/lite/micro/kernels/xtensa/conv_hifi.cc
tingyan19: Created an cleanup bug http://b/210163787 to refactor it soon.

advaitjain (dismissed):

advaitjain (approved):

PR #786, opened 2021-12-14
Set resource variable buffers to 0 by default.

njeffrie:
Ensures uninitialized resource variables on TFLM work as expected.

Bug=http://b/209832301


tensorflow/lite/micro/micro_resource_variable.cc
```diff
@@ -92,6 +92,9 @@ TfLiteStatus MicroResourceVariables::Allocate(int id, TfLiteContext* context,
     variable.bytes = tensor->bytes;
     variable.resource_buffer =
         context->AllocatePersistentBuffer(context, tensor->bytes);
+    // Zero out resource buffers by deafult. Buffers can be initialized to
+    // nonzero values using ASSIGN_VARIABLE.
+    memset(variable.resource_buffer, 0, variable.bytes);
```
tingyan19, line 97: Does it make more sense if we check whether the resource buffer is allocated successfully before calling memset() on it?

njeffrie: Good call. Move to after the null check.

tingyan19 (approved):

Issue #787, opened 2021-12-15
Enable HiFi4 add and leaky relu optimization


tingyan19:
Enable latest add and leaky relu optimization. Include new source code and related tests.

PR #788, opened 2021-12-15
Add optimized add and leaky relu (#787)

tingyan19:
Added optimized add and leaky relu kernel wrapper and updated internal tests.

BUG=#787

Test=docker run --rm -v "$(pwd)":/opt/tflite-micro ghcr.io/tflm-bot/xtensa:0.1 /opt/tflite-micro/tensorflow/lite/micro/tools/ci_build/test_xtensa_hifi3z.sh INTERNAL


tensorflow/lite/micro/kernels/xtensa/add.cc
```diff
@@ -0,0 +1,201 @@
...
+  op_params.output_offset = data->output_offset;
+  op_params.output_multiplier = data->output_multiplier;
+  op_params.output_shift = data->output_shift;
+  SetActivationParams(data->output_activation_min, data->output_activation_max,
+                      &op_params);
+  bool need_broadcast = reference_ops::ProcessBroadcastShapes(
+      tflite::micro::GetTensorShape(input1),
+      tflite::micro::GetTensorShape(input2), &op_params);
+
+  switch (output->type) {
```
njeffrie, line 81: Thoughts on extracting this into a common EvalAddReference method? I'm hesitant to copy this logic, since it easily gets out of sync when we add a new reference kernel method.

tensorflow/lite/micro/kernels/xtensa/leaky_relu.cc
```diff
@@ -0,0 +1,116 @@
...
+#include "tensorflow/lite/micro/kernels/kernel_util.h"
+#include "tensorflow/lite/micro/kernels/leaky_relu.h"
+#include "tensorflow/lite/micro/kernels/xtensa/fixedpoint_utils.h"
+#include "tensorflow/lite/micro/kernels/xtensa/xtensa.h"
+#include "tensorflow/lite/micro/micro_error_reporter.h"
+
+namespace tflite {
+
+template <typename T>
+void QuantizeLeakyRelu(const LeakyReluOpData& data,
```
njeffrie, line 32: same comment w.r.t. extracting this into leaky_relu_common.cc and using a shared reference method.

tensorflow/lite/micro/tools/ci_build/test_xtensa_hifi3z.sh
```diff
@@ -50,6 +50,20 @@ if [[ ${1} == "INTERNAL" ]]; then
     OPTIMIZED_KERNEL_DIR=xtensa \
     XTENSA_CORE=HIFI_190304_swupgrade \
     test_integration_tests_seanet_conv -j$(nproc)
+
+  readable_run make -f tensorflow/lite/micro/tools/make/Makefile \
```
njeffrie, line 54: These tests will fail on GH right? Should they be protected behind a flag of some sort, or perhaps only be used in the internal presubmit?

njeffrie (approved):
After leaving the comments about extracting common code out into shared methods, I realized we do duplicate code in lots of other xtensa kernels. I'd be OK leaving the common code extraction to a follow-on change which fixes this for all optimized kernels if you want to file a fixit bug to do this.

tensorflow/lite/micro/tools/ci_build/test_xtensa_hifi3z.sh
tingyan19: Yes, they are protected behind "if [[ ${1} == "INTERNAL" ]]; then" and we only append "INTERNAL" in internal tests.

tensorflow/lite/micro/kernels/xtensa/add.cc
tingyan19: Created a separate bug in https://issuetracker.google.com/211470746 to handle this.

tensorflow/lite/micro/kernels/xtensa/leaky_relu.cc
tingyan19: Created a separate bug in https://issuetracker.google.com/211470746 to handle this.

njeffrie (approved):

PR #806, opened 2021-12-20
Mergify docs

jwithers:
Adding documentation for our mergify setup. Not sure this is going to want to stay stand alone like this and probably is going to want to be merged with docs/continuious_integration.md when we know it is stable.

Also this will get a couple of changes as the code it documents changes, the stand-alone apply_cirun.yml is probably changing for examples

BUG= #806


jwithers:
Outdated. Closing.

PR #835, opened 2022-01-02
Integrate and demonstrate use of Pigweed's unit test module

rkuester:
Integrate and demonstrate the use of Pigweed's unit test module. This series of commits adds a header and Makefile build helper for writing tests using Pigweed's unit test module, pw_unit_test, in the same manner as tests using the existing micro_test. They add two example uses: a pw_unit_test rendition of the test in examples/hello_world and a pw_unit_test rendition of the unit test for kernels/fill. The Pigweed-based tests coexist in parallel with the micro_test-based tests.

Of note is the bump from C++11 to C++17 required to use Pigweed. See the relevant commit's log for some discussion.


advaitjain:
Closing this PR. Let's see if we can come back to this in the future.

PR #837, opened 2022-01-04
Store scratch buffer arrays separately for each subgraph.

njeffrie:
Allocate scratch buffers separately for each subgraph. This avoids
allocating extra memory, and fixes correctness bugs where scratch arrays
were allocated multiple times, causing scratch buffers to be invalid in
all subgraphs except the last run during allocation (maximum index).

TESTED=ran release test which failed previously, verified scratch
buffers are only allocated once.

BUG=http://b/211906909

### DIFF
--- a/tensorflow/lite/micro/micro_allocator.cc
+++ b/tensorflow/lite/micro/micro_allocator.cc
@@ -107,7 +107,7 @@ class AllocationInfoBuilder {
   // Add allocation information for the scratch buffers.
   TfLiteStatus AddScratchBuffers(
       internal::ScratchBufferRequest* scratch_buffer_requests,
-      ScratchBufferHandle* scratch_buffer_handles);
+      ScratchBufferHandle* scratch_buffer_handles, int subgraph_idx);
 
   // Returns a pointer to the built AllocationInfo array.
   const AllocationInfo* Finish() const { return info_; }
@@ -221,21 +221,24 @@ TfLiteStatus AllocationInfoBuilder::GetOfflinePlannedOffsets(
 
 TfLiteStatus AllocationInfoBuilder::AddScratchBuffers(
     internal::ScratchBufferRequest* scratch_buffer_requests,
-    ScratchBufferHandle* scratch_buffer_handles) {
+    ScratchBufferHandle* scratch_buffer_handles, int subgraph_idx) {
   // Set up allocation info for buffers.
-  for (size_t i = tensor_count_; i < tensor_count_ + buffer_count_; ++i) {
+  for (size_t i = 0, buffers_added = 0; i < buffer_count_; ++i) {
     internal::ScratchBufferRequest* current_request =
-        &(scratch_buffer_requests[i - tensor_count_]);
-    ScratchBufferHandle* current_handle =
-        &(scratch_buffer_handles[i - tensor_count_]);
+        &(scratch_buffer_requests[i]);
+    if (current_request->subgraph_idx == subgraph_idx) {
+      ScratchBufferHandle* current_handle =
+          &(scratch_buffer_handles[buffers_added]);
 
-    AllocationInfo* current = &info_[i];
-    current->output_ptr = reinterpret_cast<void**>(&current_handle->data);
-    current->bytes = current_request->bytes;
-    current->first_created = current_request->node_idx;
-    current->last_used = current_request->node_idx;
-    current->offline_offset = kOnlinePlannedBuffer;
-    current->needs_allocating = true;
+      AllocationInfo* current = &info_[tensor_count_ + buffers_added];
+      current->output_ptr = reinterpret_cast<void**>(&current_handle->data);
+      current->bytes = current_request->bytes;
+      current->first_created = current_request->node_idx;
+      current->last_used = current_request->node_idx;
+      current->offline_offset = kOnlinePlannedBuffer;
+      current->needs_allocating = true;
+      buffers_added++;
+    }
   }
   return kTfLiteOk;
 }
@@ -570,8 +573,7 @@ SubgraphAllocations* MicroAllocator::StartModelAllocation(const Model* model) {
 }
 
 TfLiteStatus MicroAllocator::FinishModelAllocation(
-    const Model* model, SubgraphAllocations* subgraph_allocations,
-    ScratchBufferHandle** scratch_buffer_handles) {
+    const Model* model, SubgraphAllocations* subgraph_allocations) {
   if (!model_is_allocating_) {
     TF_LITE_REPORT_ERROR(error_reporter_,
                          "MicroAllocator: Model allocation finished before "
@@ -586,10 +588,12 @@ TfLiteStatus MicroAllocator::FinishModelAllocation(
     TFLITE_DCHECK(subgraph != nullptr);
 
     TF_LITE_ENSURE_STATUS(AllocateScratchBufferHandles(
-        scratch_buffer_handles, scratch_buffer_request_count_));
+        &subgraph_allocations[subgraph_idx].scratch_buffer_handles,
+        subgraph_allocations[subgraph_idx].scratch_buffer_request_count_));
     TF_LITE_ENSURE_STATUS(CommitStaticMemoryPlan(
         model, subgraph_allocations[subgraph_idx].tensors,
-        *scratch_buffer_handles, subgraph_idx));
+        subgraph_allocations[subgraph_idx].scratch_buffer_handles,
+        subgraph_idx));
     TF_LITE_ENSURE_STATUS(AllocateVariables(
         subgraph, subgraph_allocations[subgraph_idx].tensors));
   }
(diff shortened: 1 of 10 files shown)


tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -221,21 +221,24 @@ TfLiteStatus AllocationInfoBuilder::GetOfflinePlannedOffsets(
...
-    current->first_created = current_request->node_idx;
-    current->last_used = current_request->node_idx;
-    current->offline_offset = kOnlinePlannedBuffer;
-    current->needs_allocating = true;
+        &(scratch_buffer_requests[i]);
+    if (current_request->subgraph_idx == subgraph_idx) {
+      ScratchBufferHandle* current_handle =
+          &(scratch_buffer_handles[buffers_added]);
+
+      AllocationInfo* current = &info_[tensor_count_ + buffers_added];
```
deqiangc, line 233: Might be worthy of adding a comment about why  +tensor_count_?

tensorflow/lite/micro/micro_allocator_test.cc
```diff
@@ -356,7 +352,6 @@ TF_LITE_MICRO_TEST(TestMockModelAllocation) {
 
 TF_LITE_MICRO_TEST(TestMockModelAllocationWithGivenMemoryPlanner) {
```
deqiangc, line 353: Can we add a test specific for where the issue happens: one model that contains two subgraphs?

tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -221,21 +221,24 @@ TfLiteStatus AllocationInfoBuilder::GetOfflinePlannedOffsets(
 
 TfLiteStatus AllocationInfoBuilder::AddScratchBuffers(
     internal::ScratchBufferRequest* scratch_buffer_requests,
-    ScratchBufferHandle* scratch_buffer_handles) {
+    ScratchBufferHandle* scratch_buffer_handles, int subgraph_idx) {
```
deqiangc, line 224: IIUC, AllocationInfoBuilder is created per subgraph (https://source.corp.google.com/piper///depot/google3/third_party/tflite_micro/tensorflow/lite/micro/micro_allocator.cc;l=900;bpv=1;bpt=1;rcl=419362405).

The tensor_count_ inside AllocationInfoBuilder is already set per subgraph at construction time, so WDYT of passing subgraph_idx as a constructor parameter to AllocationInfoBuilder and avoid passing subgraph_idx all the time?

tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -221,21 +221,24 @@ TfLiteStatus AllocationInfoBuilder::GetOfflinePlannedOffsets(
...
   // Set up allocation info for buffers.
-  for (size_t i = tensor_count_; i < tensor_count_ + buffer_count_; ++i) {
+  for (size_t i = 0, buffers_added = 0; i < buffer_count_; ++i) {
```
deqiangc, line 226: the buffer_count_ here is really total_buffer_count_ (total for all subgraphs), is that understanding correct?

tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -880,7 +887,7 @@ TfLiteStatus MicroAllocator::CommitStaticMemoryPlan(
 
   const SubGraph* subgraph = model->subgraphs()->Get(subgraph_idx);
   size_t allocation_info_count =
```
deqiangc, line 889: Do you think it may be worthy to add a comment here that there is a potential optimization here:
ideally, we just need subgraph->tensors()-size() + number of scratch request for the current subgraph.
But we ignore this for this optimization is unlikely to gain much for most realistic mode with large tensors.

tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -880,7 +887,7 @@ TfLiteStatus MicroAllocator::CommitStaticMemoryPlan(
...
-      subgraph->tensors()->size() + scratch_buffer_request_count_;
+      subgraph->tensors()->size() + total_scratch_requests_;
```
deqiangc, line 890: subgraph->tensors()->size() + subgraph_allocations[subgraph_idx].scratch_buffer_request_count_ should work, right?

njeffrie: Good call. I've added it as an argument; subgraph_allocations is not accessible from within this method.

tensorflow/lite/micro/micro_allocator_test.cc
njeffrie: Done

tensorflow/lite/micro/micro_allocator.cc
njeffrie: Done.

tensorflow/lite/micro/micro_allocator.cc
njeffrie: correct.

tensorflow/lite/micro/micro_allocator.cc
njeffrie: It is - I would slightly prefer to keep it this way to keep AllocationInfoBuilder small. If we need to add subgraph_index into another method I think then it should be part of AllocationInfoBuilder. I don't have a strong preference one way or another, though.

tensorflow/lite/micro/micro_allocator_test.cc
```diff
@@ -115,7 +116,7 @@ void AllocateAndVerifyMockWeightTensor(
 void EnsureUniqueVariableTensorBuffer(const Model* model,
                                       TfLiteEvalTensor* eval_tensors,
                                       const int variable_tensor_idx) {
-  for (size_t i = 0; i < GetModelTensorCount(model); i++) {
+  for (size_t i = 0; i < GetModelTensorCount(model, 0); i++) {
```
deqiangc, line 119: nit: /*subgraph_idx=*/0?

tensorflow/lite/micro/micro_allocator_test.cc
```diff
@@ -401,6 +396,57 @@ TF_LITE_MICRO_TEST(TestMockModelAllocationWithGivenMemoryPlanner) {
...
+      subgraph_allocations[1].scratch_buffer_handles[0].data);
+
+  // The memory layout is expected to be:
+  // subgraph 0:
+  // [tensor1 (128 bytes)] [tensor 0 (128 bytes)] [scratch (16 bytes)]
+  // subgraph 1:
+  // [tensor0 (128 bytes)] [scratch (16 bytes)]
+  TF_LITE_MICRO_EXPECT_EQ(subgraph0_tensor1_ptr + 128, subgraph0_tensor0_ptr);
+  TF_LITE_MICRO_EXPECT_EQ(subgraph0_tensor0_ptr + 128, subgraph0_scratch_ptr);
+  TF_LITE_MICRO_EXPECT_EQ(subgraph1_tensor0_ptr + 128, subgraph1_scratch_ptr);
```
deqiangc, line 447: another invariant is:
TF_LITE_MICRO_EXPECT_EQ(subgraph0_tensor0_ptr, subgraph1_tensor0_ptr).  

WDYT adding that here as well. I think that clear illustrate our assumption between subgraphs: everything is replanned (both tensor and scratch buffer).

tensorflow/lite/micro/micro_allocator_test.cc
deqiangc: and other place in which GetModelTensorCount is called. On another thought, if you have to change every place with a new argument, might just change the function to GetSubgraphTensorCountFromModel(). Up to you though

tensorflow/lite/micro/test_helpers.h
```diff
@@ -161,6 +161,9 @@ const Model* GetSimpleModelWithSubgraphsAndIf();
 // of "if" subgraph overlaps with the input tensor 2 of subgraph 1.
 const Model* GetModelWithIfAndSubgraphInputTensorOverlap();
 
+// Returns a flatbuffer model with two subgraphs.
```
deqiangc, line 164: nitpicking: may want to comment number of tensor and scratch buffer request here?

tensorflow/lite/micro/micro_context.cc
```diff
@@ -35,11 +35,15 @@ void* MicroContext::AllocatePersistentBuffer(size_t bytes) {
...
                                                        int* buffer_idx) {
   return allocator_.RequestScratchBufferInArena(
-      bytes, graph_.GetCurrentSubgraphIndex(), buffer_idx);
+      graph_.GetAllocations(), bytes, graph_.GetCurrentSubgraphIndex(),
+      buffer_idx);
 }
 
 void* MicroContext::GetScratchBuffer(int buffer_idx) {
-  ScratchBufferHandle* handle = scratch_buffer_handles_ + buffer_idx;
+  ScratchBufferHandle* handle =
```
deqiangc, line 43: nit: maybe add a todo with bug here:  I feel this logic should probably be abstracted into either micro_allocator or micro_graph because it is very related with the implementation of how scratch buffer is implemented and it is better to let micro_context just be a shim layer

deqiangc (approved):
LGTM overall.  Most comments are nitpicking out of my preference of having as much as possible.

tensorflow/lite/micro/micro_allocator_test.cc
njeffrie: Done.

njeffrie:
Closing since this is redundant after #947

PR #839, opened 2022-01-04
Review approval applies tags

jwithers:
Upon an approved review, the ci:run and ci:ready_to_merge tags are applied to the pr. 

This is a touch unwieldy and took a bit to figure out because it has to be two different files, ready_to_merge_apply and ready_to_merge trigger. This is due to the permissions issues around changing labels on PRs coming from forks. The pull_request_review trigger doesn't have the privilege escalation features of pull_request_target, thus the labeling portion has to handed off to another job triggered by workflow_run. 

These files can't be consolidated because workflow_run triggers on workflows, not jobs.

And the ugly bit with passing the pr number by artifacts is that the workflow_run event doesn't have a PR source, so you have to pass it in. There is no practical way you could get a false launch with this setup, although I think you could possibly end up with a situation where an approved review doesn't get tagged if you approve a spate of PRs and one gets delayed exactly correctly for the artifact to be overwritten by the following pr at the precise point between writing the artifact and reading it. I don't think this is a particularly high probability event and the worst case scenario is you will have to add the tags manually to some random PR out of a string.

BUG=https://issuetracker.google.com/issues/210946861

### DIFF
--- a/.github/workflows/ready_to_merge_apply.yml
+++ b/.github/workflows/ready_to_merge_apply.yml
@@ -1,0 +1,56 @@
+name: Ready To Merge Apply
+
+###
+#
+# Applies ci:ready_to_merge tag on pr number set in an artifact by ready_to_merge_trigger.yml.
+# Please see comments in that file for more
+#
+###
+
+on:
+  workflow_run:
+    workflows: ["Ready To Merge Trigger"]
+    types:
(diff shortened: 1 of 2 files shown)


advaitjain (approved):
Certainly more complex than would be ideal.

PR #842, opened 2022-01-05
Patch to fix project generation failure only seen in google3

deqiangc:
This is the patch inside http://cl/419841588

BUG=internal presubmit failure

### DIFF
--- a/tensorflow/lite/micro/tools/make/Makefile
+++ b/tensorflow/lite/micro/tools/make/Makefile
@@ -404,6 +404,7 @@ $(TFL_CC_HDRS)
 # platform-specific headers.
 # some kiss fft source file has to be included in header sections because
 # the implemenation of three different resolution fft from one single c file.
+# See http://b/201319430 for additional context.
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -402,12 +402,9 @@ $(TFL_CC_HDRS)
 
 # TODO(b/165940489): Figure out how to avoid including fixed point
 # platform-specific headers.
+# some kiss fft source file has to be included in header sections because
+# the implemenation of three different resolution fft from one single c file.
```
advaitjain, line 406: perhaps add the following to the comment:

See http://b/201319430 for additional context

advaitjain (changes requested):

tensorflow/lite/micro/tools/make/Makefile
deqiangc: Done

advaitjain (approved):

PR #844, opened 2022-01-07
Create a script to report size data of multiple binaries into a csv file

deqiangc:
create_size_log_x86.sh build some chosen binaries for linux x86 release flavor and append the size report into size_data/size_linux_x86_64_release.csv. 
Currently, this script fails if there is any failure in building or profiling target binaries.
Eventually, this script will also calls into another script to detect  size increase and fail if any substantial size increases.

This script is a modified version of https://github.com/advaitjain/tflite-micro/blob/local-continuous-builds/tensorflow/lite/micro/docs/local_continuous_builds/xtensa_test.sh and is for x86 target.  Each hardware target is likely to have its own script due to the need for separate container. 

BUG=https://issuetracker.google.com/211673155


tensorflow/lite/micro/tools/metrics/create_size_log_x86.sh
```diff
@@ -0,0 +1,139 @@
...
+
+end_size_report ${SIZE_LOG}
+
+if [[ ${KEYWORD_BENCHMARK_STATUS} != 0 || ${BASELINE_MEMORY_FOOTPRINT_STATUS} != 0 || ${INTERPRETER_MEMORY_FOOTPRINT_STATUS} != 0 ]]
+then
+  echo "Failure in profiling."
+  exit -1
+fi
+
+## TODO: run difference detection and also return error code if detecting large increase
```
advaitjain, line 138: nit: add bug with the TODO.

tensorflow/lite/micro/tools/metrics/create_size_log_x86.sh
```diff
@@ -0,0 +1,139 @@
...
+ROOT_DIR=${SCRIPT_DIR}/../../../../..
+GEN_FILES_DIR=${ROOT_DIR}/tensorflow/lite/micro/tools/make/gen/
+cd "${ROOT_DIR}"
+
+source tensorflow/lite/micro/tools/ci_build/helper_functions.sh
+
+START_HEADER="################## START ##############################"
+END_HEADER="#################### END ############################"
+HEAD_SHA=`git rev-parse HEAD`
+SIZE_LOG=${SCRIPT_DIR}/size_data/size_linux_x86_64_release.csv
```
advaitjain, line 116: my guess is that we would benefit from a separate log file for each target. That will make it easier to write analysis scripts since they can be focused on a single target's log file instead of having to parse the combined log file and separate out per-binary logs.

advaitjain, line 116: Let's put the log files in a separate top-level folder. Maybe a structure like:
```
tflite-micro/continuous_builds_data/size_profiling/x86_linux_release/<binary_target>.csv
```

And put a README.md in tflite-micro/continuous_builds_data describing what this folder will contain, that all the logs are generated via automated workflows ...

tensorflow/lite/micro/tools/metrics/size_data/size_linux_x86_64_release.csv
```diff
@@ -0,0 +1,19 @@
+################## START ##############################
```
advaitjain, line 1: I'm unsure if the START and END delimiting lines are adding a whole lot to the readability of this log file.

And I'm also wondering if there is value in having the logs be more structured like yaml or json so that any future analysis can benefit from that?

In that scenario we would benefit from parsing the output of the size command in Python and outputting whatever format we feel is scalable.

tensorflow/lite/micro/tools/metrics/create_size_log_x86.sh
deqiangc: done

tensorflow/lite/micro/tools/metrics/create_size_log_x86.sh
deqiangc: Done.

tensorflow/lite/micro/tools/metrics/size_data/size_linux_x86_64_release.csv
deqiangc: Done as we discussed.

advaitjain (approved):

PR #853, opened 2022-01-14
Avoid unintentional memory overwrite in IF operator

deqiangc:
IF operator copies its input to the active branch subgraph's input and copy the active subgraph's output's to IF's output. Because IF's tensors and the active branch subgraph's tensors are separately planned and can have overlap.  This PR fix the unintentionl overwrite due to those overlap by using an intermediate buffer. 

Bug=https://b/214244865


tensorflow/lite/micro/kernels/if.cc
```diff
@@ -111,32 +197,118 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
-       i < graph_info.NumSubgraphOutputs(active_branch_subgraph_index); ++i) {
-    const TfLiteEvalTensor* output =
-        tflite::micro::GetEvalOutput(context, node, i);
-
-    TfLiteEvalTensor* subgraph_output =
-        graph_info.GetSubgraphOutput(active_branch_subgraph_index, i);
+  // The input tensors of the IF subgraph may already get overwritten; these
+  // tensors may still be used by other OPs and we need to restore it first from
+  // intermediate buffers.
+  CopyFromIntermediateToIfInput(context, node, graph_info,
```
shlmregev, line 299: Writing to the if op's input tensors can overwrite the active subgraph's output tensors before we get a chance to copy them to the intermediate (line 307).

tensorflow/lite/micro/kernels/if.cc
```diff
@@ -111,32 +197,118 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
+                                active_branch_subgraph_index);
+  // The output tensors of the IF subgraph and active branch subgraphs are
```
shlmregev, line 301: Perhaps a clearer comment would be:
"The output tensors of the IF op's subgraph and the active branch subgraph"

tensorflow/lite/micro/kernels/if.cc
```diff
@@ -111,32 +197,118 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
                     graph_info.InvokeSubgraph(active_branch_subgraph_index));
 
-  for (size_t i = 0;
-       i < graph_info.NumSubgraphOutputs(active_branch_subgraph_index); ++i) {
-    const TfLiteEvalTensor* output =
-        tflite::micro::GetEvalOutput(context, node, i);
-
-    TfLiteEvalTensor* subgraph_output =
-        graph_info.GetSubgraphOutput(active_branch_subgraph_index, i);
+  // The input tensors of the IF subgraph may already get overwritten; these
```
shlmregev, line 296: IIUC, what might get overwritten are the IF op's input tensors, not the IF op subgraph's input tensors. A slight difference if true. Maybe a better way to phrase it would be:
"The IF op's input tensor may have already been overwritten.."

tensorflow/lite/micro/kernels/if.cc
```diff
@@ -111,32 +197,118 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
+  bool cond_value = cond->data.b[0];
+
+  tflite::MicroContext* micro_context = tflite::GetMicroContext(context);
+  MicroGraph& graph_info = micro_context->graph();
+
+  // Currently we copy the input / output between the subgraphs.
+  int active_branch_subgraph_index =
+      cond_value ? op_data->then_subgraph_index : op_data->else_subgraph_index;
+
+  // The input tensors of the IF subgraph and active branch subgraphs are
```
shlmregev, line 282: Perhaps a clearer comment would be:
"The input tensors of the IF op's subgraph and the active branch subgraph.."

shlmregev (changes requested):

tensorflow/lite/micro/kernels/if.cc
shlmregev: I see that happen with the model that I'm testing now.

njeffrie: +1 - I think that in order to avoid this issue, we need to save the intermediate inputs and have a separate intermediate output buffer.

Before invoking the subgraph, the following can happen:
1. Copy If inputs to input intermediate
2. Copy input intermediate to subgraph inputs

After invoking the subgraph the following can happen in this order:
1. Copy subgraph outputs to output intermediate
2. Copy input intermediate back to if op inputs
3. Copy output intermediate to if op outputs

Wdyt?

tensorflow/lite/micro/kernels/if.cc
```diff
@@ -76,24 +121,65 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
...
 
-  const TfLiteTensor* cond;
-  TF_LITE_ENSURE_OK(context, GetInputSafe(context, node, 0, &cond));
-  bool cond_value = cond->data.b[0];
+    TfLiteEvalTensor* subgraph_output =
+        graph_info.GetSubgraphOutput(active_branch_subgraph_index, i);
 
-  tflite::MicroContext* micro_context = tflite::GetMicroContext(context);
-  MicroGraph& graph_info = micro_context->graph();
+    // These checks must occur in Eval since TfLiteEvalTensors are not available
```
njeffrie, line 144: While the eval tensors are not available, equivalent TfLiteTensors are available during prepare. I would slightly prefer to keep this logic within prepare, since they are designed to ensure correct model formation.

Another way to put it - the only way this check could catch something not caught during prepare would be if our eval tensors have a different type or shape from the tensors in the model and during prepare, which would indicate a much larger issue with TFLM overall.

tensorflow/lite/micro/kernels/if.cc
```diff
@@ -34,13 +34,58 @@ namespace {
...
+  kFromSubgraphToIntermediate,
+  kFromIntermediateToIf,
 };
 
 void* Init(TfLiteContext* context, const char* buffer, size_t length) {
   TFLITE_DCHECK(context->AllocatePersistentBuffer != nullptr);
   return context->AllocatePersistentBuffer(context, sizeof(OpData));
 }
 
+size_t CalculateIntermediateBufferSize(TfLiteContext* context,
```
njeffrie, line 56: If we do end up needing two separate buffers, this could be something like "AllocateIntermediateScratchBuffers" and allocate both input inermediate and output intermediate buffers.

This is quite a minor suggestion - feel free to ack and do whichever option you prefer.

njeffrie (commented):
I have a slight preference for brevity over ultimate clarity (for example, having loops to copy inputs using memcpy directly in eval rather than separate input / output copying methods with enums), but i realize this is more of a subjective, philosophical preference rather than one backed by the style guide or a clear set of best practices. Some of my suggestions may partially be driven by this preference, so feel free to acknowledge them and proceed however you feel is best.

The only potential correctness issue I see is the one caught by Shlomi, and I believe it means we will need to maintain two separate copy buffers - one for inputs and one for outputs. Happy to take some time to discuss over gvc.

tensorflow/lite/micro/kernels/if.cc
deqiangc: Thanks Shlomy and Nat! Yes, I think you folks. It looks like we need two intermediate buffer that is NON-overlapping: one for IF input and one for Subgraph Output). I will work on this change.

tensorflow/lite/micro/kernels/if.cc
deqiangc: I also prefer to do this in prepare. Unfortunately, the check here requires getting input/output tensor of other branch subgraphs and I think we don't have a good way to get those. The way to get tensor during Prepare is  GetInputSafe or GetOutput , which requires TfLiteNode as input.  I feel it may be too much to put in new api just for this purpose. Feel free to let me know if there are better ways to get tensor.

tensorflow/lite/micro/kernels/if.cc
deqiangc: done

tensorflow/lite/micro/kernels/if.cc
deqiangc: Done

tensorflow/lite/micro/kernels/if.cc
deqiangc: Thanks for the suggestion!  In the new PR, I rename this to AllocateIntermediateBuffers. I decide to avoid Scratch part because the name of Scratch buffer has been used extensively in other part of code and I feel quite confusing. I also want to call out that these intermediate buffers are in persistent arena since I could not find another way to guarantee that they don't overlap with in/output tensors (Pete and I discussed about this and we could not find a reasonable way to achieve this).

deqiangc:
> 

Thanks for the suggestion, I also agree with the need for two buffers and this new PR reflect that. 

For those functions, I am slightly biased towards saving some code space by using enum and have a common underlying function instead of have copy inside each function.  But please feel free to let me know if we should further discuss this

shlmregev (approved):
Solution makes sense and this fixes the problem with my model.

ScholliYT:
I think I found some edge case where this new approach causes an hard error. Imagine the case where one of the input tensors lies in the read-only memory space (like a kernel for Conv2D). After executing the IF Operation this new implementation tries to write the saved input from intermediate back into the read-only memory space. This causes some kind of hard crash.

I will craft a minimal example later on but this is the log I can provide for now. The Model uses 2 nested IF where the outer one gets a read only Conv2D kernel as input. It crashes after finishing the outer IF and executing this line https://github.com/tensorflow/tflite-micro/blob/746f880a405aeca2db6d005fd9fb8e451f3cea34/tensorflow/lite/micro/kernels/if.cc#L316

image


Does this make sense to you @deqiangc?

ScholliYT:
Hi @deqiangc,

I agree on the point that these "read only" inputs should be mapped as input to the subgraphs instead to save on copy operations. The use case is in fact unusual. I am currently trying to implement a research paper which needs to slice Conv2D kernels based on some `width` (0.0 to 1.0). As TFLM doesn't support dynamic shapes I went with an approach using some `IF` ops to slice with some threshold levels (like 0.25, 0.5...). The different IF branches slice according to the width threshold. More info can be found in #867. 

I know that TFLM is not really made to support these operations 😄 I can provide you with the code for this specific use case (new issue maybe?) if that helps to figure out why these go through the IF instead of being mapped directly.

deqiangc:
@ScholliYT 

Hello, Tom, 

Thanks for the explanation of the background!   It would be useful if you can elaborate how you end up with a read-only tensor as input IF operator. I guess I don't really get the part that using slide_slice would lead to a "read-only" tensor.

PR #854, opened 2022-01-14
Add virtual destructor to micro_context

deqiangc:
Some platforms generate warnings/errors that a class that has
virtual functions should have virtual destructor

BUG=observe in internal presubmit


njeffrie (approved):
How does this relate to TF_LITE_REMOVE_VIRTUAL_DELETE? If it’s different then this looks fine.

deqiangc:
> How does this relate to TF_LITE_REMOVE_VIRTUAL_DELETE? If it’s different then this looks fine.

IIUC, they serve different purpose and both are needed. explicit empty virtual destructor() is needed to make some compiler not to spit out warnings/errors when it sees a class has virtual function, but no virtual destructor.  TF_LITE_REMOVE_VIRTUAL_DELETE removes the default virtual "delete" operator only, which is used by the C++ itanium abi to correctly implement "delete" on destruction of virtual class.

PR #868, opened 2022-01-22
Add a DeallocateTemp to pair with AllocateTemp in SimpleMemoryAllocator.

deqiangc:
Add a DeallocateTemp to pair with AllocateTemp in SimpleMemoryAllocator.

This is a backport of internal cl/423466958

The idea is to pair DeallocateTemp with existing AllocateTemp
to provide a better memory auditing for temporary buffer.

After all existing AllocateTemp have been paired with a DeallocateTemp,
we can then enable an error check when the entire temp memory region
is to be reclaimed, but there are still outstanding temp buffers.

Currently, GetInput api use AllocateTemp under the hood and requires a
separate CL to change the api before we can enforce the assertion.

See go/tflm-pairing-allocatetemp-with-deallocatetemp  for more info

BUG=https://b/209453859


tensorflow/lite/micro/simple_memory_allocator.cc
```diff
@@ -113,10 +114,33 @@ uint8_t* SimpleMemoryAllocator::AllocateTemp(size_t size, size_t alignment) {
     return nullptr;
   }
   temp_ = aligned_result + size;
+  temp_buffer_ptr_check_sum_ =
+      temp_buffer_ptr_check_sum_ ^ (reinterpret_cast<intptr_t>(aligned_result));
```
tingyan19, line 118: Is there a reason that this is not written as "temp_buffer_ptr_check_sum_ ^= reinterpret_cast<intptr_t>(aligned_result);", which should be equivalent and shorter? Same applies to lines 124~125 in SimpleMemoryAllocator::DeallocateTemp(uint8_t* temp_buf) below.

deqiangc: Done. No strong reason other than my tendency of over verbosity sometimes.

tingyan19 (approved):

PR #872, opened 2022-01-25
Update binary size logs

jwithers:
Adds log_sizes.yml which uses tensorflow/lite/micro/tools/metrics/create_size_log_x86.sh to update the log files in data/continuous_builds.

This is set to be run manually right now, but a set of schedule lines is ready to be uncommented.

The commit is set to push directly to main, which might or might not work depending on how the protections are set up for the repo, which I can't see. Again, there is a commented block and instructions that will convert this to creating a PR with the log additions in the same manner as sync.yml.

This was tested in the current configuration in my fork and works, but again might not push properly in this repo depending on the security settings.

BUG=https://issuetracker.google.com/issues/211673155


deqiangc (commented):
There is a tflite-micro/.github/workflows/size_check.yml that is NOT doing much right now; so I think we can overwrite tflite-micro/.github/workflows/size_check.yml with what you have here.

.github/workflows/log_sizes.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: update logs x86
+      uses: docker://ghcr.io/tflm-bot/tflm-ci:0.1
+      with:
+        args: tensorflow/lite/micro/tools/metrics/create_size_log_x86.sh
+    - name: commit logs
+      run: |
+        git config --global user.name 'TFLM-bot'
+        git config --global user.email 'tflm-github-bot@google.com'
+        git commit -am "Automated size log update"
+        git push
```
deqiangc, line 24: Suggest to comment the "git push" and uncomment the below PR generation. TFLM repo does not allow directly push I believe.

.github/workflows/log_sizes.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: update logs x86
+      uses: docker://ghcr.io/tflm-bot/tflm-ci:0.1
+      with:
+        args: tensorflow/lite/micro/tools/metrics/create_size_log_x86.sh
+    - name: commit logs
+      run: |
+        git config --global user.name 'TFLM-bot'
```
deqiangc, line 21: nitpicking: --local

.github/workflows/log_sizes.yml
```diff
@@ -0,0 +1,41 @@
...
+#      uses: peter-evans/create-pull-request@052fc72b4198ba9fbc81b818c6e1859f747d49a8
+#      with:
+#         branch: size-log-update
+#         delete-branch: true
+#         token: ${{ secrets.TFLM_BOT_REPO_TOKEN }}
+#         title: Automated binary size log update
+#         commit-message: Automated binary size log update
+#         committer: TFLM-bot <tflm-github-bot@google.com>
+#         author: TFLM-bot <tflm-github-bot@google.com>
+#         body: "BUG=automated binary size log update\nNO_CHECK_TFLITE_FILES=automated log size update"
```
deqiangc, line 39: nitpicking: may not need NO_CHECK_TFLITE_FILES since they are not sahred.

.github/workflows/log_sizes.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: update logs x86
+      uses: docker://ghcr.io/tflm-bot/tflm-ci:0.1
+      with:
+        args: tensorflow/lite/micro/tools/metrics/create_size_log_x86.sh
+    - name: commit logs
+      run: |
+        git config --global user.name 'TFLM-bot'
+        git config --global user.email 'tflm-github-bot@google.com'
```
deqiangc, line 22: nitpicking: --local

jwithers:
@deqiangc agreed on the not needing size_check, and kinda feel like we could go further because that is pretty generic and replace both this and size_check with log_binary_size.yml, which is more exactly descriptive of the function here.

deqiangc (approved):

PR #874, opened 2022-01-25
Use -Oz for CORE_OPTIMIZATION_LEVEL if armclang is used.

jenselofsson:
For armclang, -Oz is the equivalent of the -Os for gcc. -Os in armclang balances code size and speed, while -Oz optimizes with regards to image size.

More info here:
https://developer.arm.com/documentation/100067/0612/armclang-Command-line-Options/-O?lang=en

BUG=#879


tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -110,7 +110,11 @@ ifneq ($(CO_PROCESSOR),)
   ADDITIONAL_DEFINES += -D$(shell echo $(CO_PROCESSOR) | tr [a-z] [A-Z])
 endif
 
-CORE_OPTIMIZATION_LEVEL := -Os
+ifeq ($(TOOLCHAIN), armclang)
+  CORE_OPTIMIZATION_LEVEL := -Oz
+else ifeq ($(TOOLCHAIN), gcc)
```
njeffrie, line 115: Setting CORE_OPTIMIZATION_LEVEL to -Os should probably not be conditional on ifeq($(TOOLCHAIN), gcc). We use both gcc and clang, and having a catch-all here is probably best.

njeffrie (changes requested):

tensorflow/lite/micro/tools/make/Makefile
jenselofsson: I uploaded a mew commit, was it that you had in mind? I also just realized that I forgot to create and reference an issue. I'll edit my first comment with a link to it.

njeffrie (approved):

PR #885, opened 2022-01-27
Add AllocateTempInputTensor to replace GetInput() and Add AllocateTempOutputTensor to replace GetOutput()

deqiangc:
Add more informative api AllocateTempInputTensor to replace GetInput()
and AllocateTempOutputTensor to replace GetOutput(). These two new API
call out the fact that what GetInput and GetOutput returns are in
temporary buffer and  need to call DeallocateTempTensor() afterwards.

IF operator is the first OP to use these APIs.

BUG=https://b/209453859

### DIFF
--- a/tensorflow/lite/micro/fake_micro_context.cc
+++ b/tensorflow/lite/micro/fake_micro_context.cc
@@ -39,8 +39,17 @@ FakeMicroContext::FakeMicroContext(TfLiteTensor* tensors,
       tensors_(tensors),
       allocator_(allocator) {}
 
-TfLiteTensor* FakeMicroContext::GetTensor(int tensor_index) {
+TfLiteTensor* FakeMicroContext::AllocateTempTfLiteTensor(int tensor_index) {
+  allocated_tensor_count_++;
   return &tensors_[tensor_index];
+}
+
+void FakeMicroContext::DeallocateTempTfLiteTensor(TfLiteTensor* tensor) {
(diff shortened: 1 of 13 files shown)


tensorflow/lite/micro/fake_micro_context.cc
```diff
@@ -39,10 +39,19 @@ FakeMicroContext::FakeMicroContext(TfLiteTensor* tensors,
...
+TfLiteTensor* FakeMicroContext::AllocateTempTfLiteTensor(int tensor_index) {
+  allocated_tensor_count_++;
   return &tensors_[tensor_index];
 }
 
+void FakeMicroContext::DeallocateTempTfLiteTensor(TfLiteTensor* tensor) {
+  allocated_tensor_count_--;
+}
+
+bool FakeMicroContext::IsAllTempTfLiteTensorDeallocated() {
```
tingyan19, line 51: Maybe change method name to AreAllTempTfLiteTensorsDeallocated (using plural), or just AllTempTfLiteTensorsDeallocated?

njeffrie (approved):

tensorflow/lite/micro/fake_micro_context.cc
deqiangc: Just saw this comment. Will make a fixit bug for it. I am not sure of using Are. I might change it to HasAllocatedTempTfLiteTensor(), which avoid is vs are

PR #894, opened 2022-02-01
Replace (almost) all kernel's GetInput/GetOutput with AllocateTempXXx and DeallocateTempXxx micro context api

deqiangc:
Replace (almost) all kernel's GetInput/GetOutput with AllocateTempXXx and DeallocateTempXxx micro context api. 
This is a mostly mechanical change and a continuation of pr #885 to improve both readability and safety of getting tensors.

Almost all kernel are updated with the new api usage except a few, namely, strided_slice, squeeze, transpose, which put the returned value of GetInput/GetOutput in an OpData structure that is persistent; hence it is better to have separate change to modify opdata of those kernels and replace their GetInput/GetOutput usage.  

BUG=https://b/209453859


advaitjain (approved):
I'll trust the CI on this one.

Issue #900, opened 2022-02-01
Duplicated int8 test for tanh with wrong name


tingyan19:
There are two SimpleTestTanhUInt8 tests in tanh_test.cc. They test Int8 (so "UInt8" is a misnomer) and are exactly the same.

PR #901, opened 2022-02-01
Remove duplicated tanh int8 test and correct name

tingyan19:
BUG=#900


deqiangc (changes requested):
One question: not sure why the function (the one that is left after this change) is called SimpleTestTanhUInt8. I am confused b/c  UInt8 indicates the test is for uint8, but the TestTanhQuantized is templated to int8_t type.  

This is unrelated with this change, but wants to raise this observation. If uint8 is right name, maybe take this change to add a comment here.  Otherwise, add a fixit todo bug somewhere. WDYT? 

I am unfamiliar with this code, but removing redundant function seems right to me. So LGTM.

deqiangc (approved):

PR #905, opened 2022-02-02
Fix unreachable code flagged during sync to Google3

rockyrhodes:
BUG=Fix unreachable code issue


deqiangc (dismissed):
Thanks Rocky!  I have another similar fix about to be pushed but together with other changes. But this itself is good to go.

deqiangc (approved):

PR #923, opened 2022-02-11
Add `kTfLiteActRelu0To1` for a new fused activation

jaeyoo:
This PR adds a new ReLU activation clamping into [0, 1] to avoid the inefficiency of the current two steps - Relu6 -> [0, 6] and multiplying 0.166667. This PR is required to pass the internal testing because the source of truth of `tflite-micro` is github.

### DIFF
--- a/tensorflow/lite/c/builtin_op_data.h
+++ b/tensorflow/lite/c/builtin_op_data.h
@@ -71,6 +71,7 @@ typedef enum {
   kTfLiteActTanh,
   kTfLiteActSignBit,
   kTfLiteActSigmoid,
+  kTfLiteActRelu0To1,  // min(max(0, x), 1)
 } TfLiteFusedActivation;
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


tensorflow/lite/c/builtin_op_data.h
```diff
@@ -71,6 +71,7 @@ typedef enum {
   kTfLiteActTanh,
   kTfLiteActSignBit,
   kTfLiteActSigmoid,
+  kTfLiteActRelu0To1,  // min(max(0, x), 1)
 } TfLiteFusedActivation;
```
deqiangc, line 75: is TfLiteFusedActivation persisted in file like tflite file? If yes, then I agree with putting it at the bottom (out of order with other kTfLiteActReluXX) to maintain backward compatibility. 

overall LGTM.

jaeyoo: Yes, in the internal CL, it also adds `kTfLiteActRelu0To` at the bottom for backward compatibility

PR #948, opened 2022-03-02
Add method to reset resource variables.

njeffrie:
Add method to reset resource variables. Call this method from ResetVariableTensors.

BUG=http://b/220940833


tingyan19 (approved):

steven-buytaert:
Who found it a good idea to call free (via FreeSubgraphs) on all the nodes that are still being used for invocation later?

PR #951, opened 2022-03-02
Added support for int8 depthwise_conv on VisionP6

psharath1:
BUG=Depthwise_conv on Vision P6


psharath1:
Seeing build issue
make: execvp: tensorflow/lite/micro/tools/make/ext_libs/xtensa_download.sh: Permission denied

PR #964, opened 2022-03-08
Port operator BROADCAST_TO from lite to micro

deqiangc:
For micro, implement the BROADCAST_TO kernel and test, and add both to the build.
This PR contains several commits for readable history and ease of review following the step in 
https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/porting_reference_ops.md#5-port-the-op-from-lite-to-micro-pr3

BUG=https://b/222716314


tensorflow/lite/micro/kernels/broadcast_to.cc
```diff
@@ -0,0 +1,138 @@
...
+  BroadcastToContext op_context(context, node);
+
+  TF_LITE_ENSURE_MSG(context, (NumDimensions(op_context.input) <= kMaxDims),
+                     "BroadcastTo only supports 1-5D tensor.");
+
+  TF_LITE_ENSURE(context, op_context.shape->type == kTfLiteInt32 ||
+                              op_context.shape->type == kTfLiteInt64);
+  TF_LITE_ENSURE_EQ(context, op_context.input->type, op_context.output->type);
+
+  // Not yet support string type due to the use of memcopy with fixed size.
```
njeffrie, line 108: s/Strings are unsupported due to their variable size?

tensorflow/lite/micro/kernels/broadcast_to_test.cc
```diff
@@ -0,0 +1,214 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
njeffrie, line 1: s/2020/2022

deqiangc: done. thanks for catching

tensorflow/lite/micro/kernels/broadcast_to.cc
```diff
@@ -0,0 +1,138 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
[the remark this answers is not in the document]
deqiangc: done. thanks for pointing out this

tensorflow/lite/micro/kernels/broadcast_to.cc
deqiangc: Yes and this constraint is carried from the lite's implementation: lite does not support it, so no need for micro to support it either I suppose

tensorflow/lite/micro/kernels/broadcast_to.cc
```diff
@@ -0,0 +1,138 @@
...
+constexpr int kInputTensor = 0;
+constexpr int kShapeTensor = 1;
+constexpr int kOutputTensor = 0;
+// A maximum of 5 dimensions.
+constexpr int kMaxDims = 5;
+
+struct BroadcastToContext {
+  BroadcastToContext(TfLiteContext* context, TfLiteNode* node) {
+    micro_context = GetMicroContext(context);
+    input = micro_context->AllocateTempInputTensor(node, kInputTensor);
```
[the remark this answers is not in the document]
deqiangc: I have get rid of this structure/class completely. PTAL.

The trouble is that class constructor is not ideal in handling error. The pattern to address initialization error that I knows of typically requires injection for any things that can fail at initialization. I feel this is a bit loaded for this goal.
This structure is carried from lite's implementation to keep implementation closed to the lite's implementation, but it is not too bad to get rid of it. So that is what I ended up doing.

tensorflow/lite/micro/kernels/broadcast_to.cc
njeffrie: Ack. Perhaps comment can be updated to clear up. I was confused at first.

tensorflow/lite/micro/kernels/broadcast_to.cc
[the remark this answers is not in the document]
njeffrie: Changes look good. It looks like the struct is still here, perhaps missed removing it?

tensorflow/lite/micro/kernels/broadcast_to.cc
deqiangc: Done. Reword the original verbatim comment copy from tflite to something more specific for micro

tensorflow/lite/micro/kernels/broadcast_to.cc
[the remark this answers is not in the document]
deqiangc: Thanks for catching this! Complete the change now.

njeffrie (approved):

PR #977, opened 2022-03-17

mansnils:
Adds int8 and int16x8 specific registrations for CMSIS-NN conv kernel.

Person detect example is updated with int8 specific
registration for conv. It will result in a smaller memory footprint
for kernels that support this. For kernels that not support it the
default registration will be used.

Also 16x8 conv unit tests are updated with specific 16x8
registration. For kernels that do not have this it will default to
default registration.

### DIFF
--- a/tensorflow/lite/micro/examples/person_detection/main_functions.cc
+++ b/tensorflow/lite/micro/examples/person_detection/main_functions.cc
@@ -1,4 +1,4 @@
-/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -75,7 +75,7 @@ void setup() {
   // NOLINTNEXTLINE(runtime-global-variables)
   static tflite::MicroMutableOpResolver<5> micro_op_resolver;
   micro_op_resolver.AddAveragePool2D();
-  micro_op_resolver.AddConv2D();
+  micro_op_resolver.AddConv2D(tflite::Register_CONV_2D_INT8REF());
   micro_op_resolver.AddDepthwiseConv2D();
   micro_op_resolver.AddReshape();
   micro_op_resolver.AddSoftmax();
--- a/tensorflow/lite/micro/examples/person_detection/person_detection_test.cc
+++ b/tensorflow/lite/micro/examples/person_detection/person_detection_test.cc
@@ -1,4 +1,4 @@
-/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -55,7 +55,7 @@ TF_LITE_MICRO_TEST(TestInvoke) {
   // needed by this graph.
   tflite::MicroMutableOpResolver<5> micro_op_resolver;
   micro_op_resolver.AddAveragePool2D();
-  micro_op_resolver.AddConv2D();
+  micro_op_resolver.AddConv2D(tflite::Register_CONV_2D_INT8REF());
   micro_op_resolver.AddDepthwiseConv2D();
   micro_op_resolver.AddReshape();
   micro_op_resolver.AddSoftmax();
--- a/tensorflow/lite/micro/kernels/cmsis_nn/conv.cc
+++ b/tensorflow/lite/micro/kernels/cmsis_nn/conv.cc
@@ -1,4 +1,4 @@
-/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -37,6 +37,11 @@ struct OpData {
   // Index to buffer for optimizations if applicable.
   int buffer_idx;
 };
+
+// TODO(b/169801227): This global struct is needed for the linker to drop unused
+// code (for example, by using Register_FULLY_CONNECTED_INT8 instead of
+// Register_FULLY_CONNECTED).
+TfLiteRegistration conv_registration;
 
 void* Init(TfLiteContext* context, const char* buffer, size_t length) {
   TFLITE_DCHECK(context->AllocatePersistentBuffer != nullptr);
@@ -88,7 +93,7 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
   output_dims.w = output->dims->data[2];
   output_dims.c = output_shape.Dims(3);
 
-  if (input->type == kTfLiteInt8) {
+  if (input->type == kTfLiteInt8 || input->type == kTfLiteInt16) {
     const int num_channels = filter->dims->data[kConvQuantizedDimension];
     data->reference_op_data.per_channel_output_multiplier =
         static_cast<int32_t*>(context->AllocatePersistentBuffer(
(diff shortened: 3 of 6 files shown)


tensorflow/lite/micro/kernels/conv.h
```diff
@@ -87,6 +87,16 @@ inline TfLiteRegistration Register_CONV_2D_INT8REF() {
 }
 #endif
 
+#if defined(CMSIS_NN)
+// Returns a TfLiteRegistration struct for kernel variant that only supports
+// int16 inputs and outputs.
+TfLiteRegistration Register_CONV_2D_INT16X8REF();
```
advaitjain, line 93: That the INT8 variant has REF in its name is because it is a way to explicitly call the reference implementation even when we select OPTIMIZED_KERNEL_DIR=xtensa.

In the case of this PR, what you are looking for are variants called CONV_2D_INT8 and CONV_2D_INT16.

Similar to what we have for fully connected:
https://github.com/tensorflow/tflite-micro/blob/1c6b50ef8b9dac1bb31dfe79679aec781666a86b/tensorflow/lite/micro/kernels/fully_connected.h#L68-L81

I would suggest the following

```cc
#if defined(XTENSA)
// Returns a TfLiteRegistration struct for kernel variant that only supports
// int8 activations and int8 weights and always calls the reference implementation.
TfLiteRegistration Register_CONV_2D_INT8REF();
#else
inline TfLiteRegistration Register_CONV_2D_INT8REF() {
  return Register_CONV_2D();
}

#if defined(CMSIS_NN)
// Returns a TfLiteRegistration struct for kernel variant that only supports
// int8 activations and int8 weights and uses the latency optimized implementations.
TfLiteRegistration Register_CONV_2D_INT8();

// Returns a TfLiteRegistration struct for kernel variant that only supports
// int16 activations and int8 weights and uses the latency optimized implementations.
TfLiteRegistration Register_CONV_2D_INT16();

#else
inline TfLiteRegistration Register_CONV_2D_INT8() {
  return Register_CONV_2D();
}

inline TfLiteRegistration Register_CONV_2D_INT16() {
  return Register_CONV_2D();
}

```

I'm OK with changing person_detect to use the CONV_2D_INT8 for all targets.

tensorflow/lite/micro/kernels/cmsis_nn/conv.cc
```diff
@@ -276,21 +427,6 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
       return EvalQuantizedPerChannel(context, node, params, data, input, filter,
                                      bias, output, nullptr);
       break;
-    case kTfLiteInt16: {
```
advaitjain, line 279: let's not remove int16 from the most generic Eval. We still want Register_CONV_2D() to support all the data types. Instead let's call Eval16x8 from here.

tensorflow/lite/micro/kernels/cmsis_nn/conv.cc
```diff
@@ -38,6 +38,11 @@ struct OpData {
   int buffer_idx;
 };
 
+// TODO(b/169801227): This global struct is needed for the linker to drop unused
+// code (for example, by using Register_FULLY_CONNECTED_INT8 instead of
+// Register_FULLY_CONNECTED).
+TfLiteRegistration conv_registration;
```
advaitjain, line 44: Please update the comment to reflect CONV instead of FC.

tensorflow/lite/micro/kernels/conv_test.cc
```diff
@@ -215,8 +215,8 @@ TF_LITE_MICRO_TEST(SimpleTestQuantized16x8PerChannel) {
           tflite::testing::kBiasData, bias_quantized, scales, zero_points,
           tflite::testing::kOutputShape, tflite::testing::kGoldenData,
           golden_quantized, output_scale, output_zero_point,
-          &tflite::testing::common_conv_params, tflite::Register_CONV_2D(),
-          output_data));
+          &tflite::testing::common_conv_params,
+          tflite::Register_CONV_2D_INT16X8REF(), output_data));
```
advaitjain, line 219: leave this as Register_CONV_2D (once it calls into EvalInt16x8)

tensorflow/lite/micro/kernels/conv_test.cc
```diff
@@ -325,7 +325,7 @@ TF_LITE_MICRO_TEST(SimpleTestQuantized16x8PerChannelRelu6) {
           filter_quantized, tflite::testing::kBiasShape, bias_values,
           bias_quantized, scales, zero_points, tflite::testing::kOutputShape,
           golden_data, golden_quantized, output_scale, output_zero_point,
-          &conv_params, tflite::Register_CONV_2D(), output_data));
+          &conv_params, tflite::Register_CONV_2D_INT16X8REF(), output_data));
```
advaitjain, line 328: leave this as Register_CONV_2D (once it calls into EvalInt16x8)

tensorflow/lite/micro/kernels/conv_test.cc
```diff
@@ -474,13 +474,14 @@ TF_LITE_MICRO_TEST(Kernel1x1Quantized16x8PerChannelRelu6) {
...
-                     golden_quantized, output_scale, output_zero_point,
-                     &conv_params, tflite::Register_CONV_2D(), output_data));
+      kTfLiteOk,
+      tflite::testing::TestConvQuantizedPerChannel(
+          input_shape, input_data, input_quantized, input_scale,
+          input_zero_point, filter_shape, filter_data, filter_quantized,
+          bias_shape, bias_data, bias_quantized, scales, zero_points,
+          output_shape, golden_data, golden_quantized, output_scale,
+          output_zero_point, &conv_params,
+          tflite::Register_CONV_2D_INT16X8REF(), output_data));
```
advaitjain, line 484: leave this as Register_CONV_2D (once it calls into EvalInt16x8)

tensorflow/lite/micro/examples/person_detection/person_detection_test.cc
```diff
@@ -55,7 +55,7 @@ TF_LITE_MICRO_TEST(TestInvoke) {
   // needed by this graph.
   tflite::MicroMutableOpResolver<5> micro_op_resolver;
   micro_op_resolver.AddAveragePool2D();
-  micro_op_resolver.AddConv2D();
+  micro_op_resolver.AddConv2D(tflite::Register_CONV_2D_INT8REF());
```
advaitjain, line 58: change to ``` micro_op_resolver.AddConv2D(tflite::Register_CONV_2D_INT8());```

tensorflow/lite/micro/examples/person_detection/main_functions.cc
```diff
@@ -75,7 +75,7 @@ void setup() {
   // NOLINTNEXTLINE(runtime-global-variables)
   static tflite::MicroMutableOpResolver<5> micro_op_resolver;
   micro_op_resolver.AddAveragePool2D();
-  micro_op_resolver.AddConv2D();
+  micro_op_resolver.AddConv2D(tflite::Register_CONV_2D_INT8REF());
```
advaitjain, line 78: change to ``` micro_op_resolver.AddConv2D(tflite::Register_CONV_2D_INT8());```

advaitjain (changes requested):
Thanks, I like the direction that this is going in.

Can you please add some binary size info with and without this PR to document the fact that the kernel variants are working as expected?

See this PR as an example of what I have in mind:
https://github.com/tensorflow/tensorflow/pull/43682

I'm hoping we can use the STM32F4 target (since sparkfun_edge is no longer in the TFLM repo)

tensorflow/lite/micro/kernels/conv.h
mansnils: Ok got it.

tensorflow/lite/micro/kernels/cmsis_nn/conv.cc
mansnils: Done

tensorflow/lite/micro/kernels/conv_test.cc
mansnils: Done

tensorflow/lite/micro/kernels/conv_test.cc
mansnils: done

tensorflow/lite/micro/examples/person_detection/main_functions.cc
mansnils: done

tensorflow/lite/micro/kernels/cmsis_nn/conv.cc
mansnils: Ok. Perhaps we should change to CONV_2D_INT8 for the rest of the examples as well (maybe in another PR) to keep the memory footprints down?

mansnils:
@advaitjain should we go for this?

advaitjain (approved):

PR #990, opened 2022-03-23
Resolving the issue related to std::unique_ptr::reset

cad-audio:
For unique_ptr, "new" takescare of both memory allocation and
initialization.
But in TFLM, memory allocations happens using AllocatePersistenBuffer and
AllocateTempBuffe API. This doesn't takes care of initialization.
Calling "release" before "reset" to resolve the issue of initialization.

### DIFF
--- a/tensorflow/lite/micro/kernels/xtensa/unidirectional_sequence_lstm.cc
+++ b/tensorflow/lite/micro/kernels/xtensa/unidirectional_sequence_lstm.cc
@@ -670,6 +670,11 @@ TfLiteStatus PrecomputeZeroPointTimesWeightWithBias(
   TF_LITE_ENSURE_EQ(context, weight_shape.DimensionsCount(), 2);
   const int row = weight_shape.Dims(0);
   const int col = weight_shape.Dims(1);
+  /* Releases ownership of its stored pointer, by returning its value
+     and replacing it with a null pointer. */
+  int *temp = output->release();
+  (void)temp;
+
   output->reset(new int32_t[row]);
   if (bias_tensor == nullptr) {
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/kernels/xtensa/unidirectional_sequence_lstm.cc
```diff
@@ -670,6 +670,11 @@ TfLiteStatus PrecomputeZeroPointTimesWeightWithBias(
   TF_LITE_ENSURE_EQ(context, weight_shape.DimensionsCount(), 2);
   const int row = weight_shape.Dims(0);
   const int col = weight_shape.Dims(1);
+  /* Releases ownership of its stored pointer, by returning its value
+     and replacing it with a null pointer. */
+  int *temp = output->release();
```
deqiangc, line 675: I am not entirely sure why this is required.

cad-audio: Hi @deqiangc,

In the case of "new", it allocates memory and calls the constructors to initialize the memory and the pointer will be set to nullptr.
But TFLM uses AllocatePersistentBuffer/AllocateScratchBuffer so the pointer will not be set to nullptr, instead, it uses some random values. Here the behavior of reset is undefined.

To resolve the issue, we are calling release before reset. So it does initializes the pointer to nullptr then reset will work as expected.

Thanks,
Bhanu

tingyan19:
As we agreed upon earlier, the xtensa optimized kernel will try to use similar approaches as implemented in the reference kernel (https://github.com/tensorflow/tflite-micro/pull/1044), where we do not use unique_ptr. Therefore I'm closing the PR.

PR #992, opened 2022-03-24
Add Xtensa kernel wrappers for PAD, SUB and STRIDED SLICE

tingyan19:
BUG=#876,http://b/226586976


tensorflow/lite/micro/kernels/xtensa/pad.cc
```diff
@@ -0,0 +1,273 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+
```
deqiangc, line 2: nit:2022

tensorflow/lite/micro/kernels/xtensa/pad.cc
```diff
@@ -0,0 +1,273 @@
...
+#include "tensorflow/lite/c/common.h"
+#include "tensorflow/lite/kernels/internal/portable_tensor.h"
+#include "tensorflow/lite/kernels/internal/types.h"
+#include "tensorflow/lite/kernels/kernel_util.h"
+#include "tensorflow/lite/kernels/op_macros.h"
+#include "tensorflow/lite/micro/kernels/kernel_util.h"
+#include "tensorflow/lite/micro/kernels/xtensa/xtensa.h"
+
+namespace tflite {
+namespace ops {
```
deqiangc, line 29: I think we are moving to a flat namespace (https://b.corp.google.com/issues/170230530). Unless there is a strong reason, might want to flatten these.

tingyan19: In principle I agree with you. In this specific case, the reason of using non-flattened namespaces is to be consistent with the reference pad.cc, which still uses non-flattened namespaces. If we use flattened namespaces just for the xtensa version, the build will break, since there are calls like tflite::ops::micro::Register_PADV2() in other source files. 

I would recommend leaving it as it is for now for this PR, and flatten all namespaces in a separate effort. What do you think?

deqiangc: Your plan sounds good to me.

tensorflow/lite/micro/kernels/xtensa/pad.cc
tingyan19: Done

tensorflow/lite/micro/kernels/xtensa/pad.cc
```diff
@@ -0,0 +1,273 @@
...
+            tflite::micro::GetTensorData<float>(input), &pad_value,
+            tflite::micro::GetTensorShape(output),
+            tflite::micro::GetTensorData<float>(output));
+      } else {
+        reference_ops::Pad(data->params, tflite::micro::GetTensorShape(input),
+                           tflite::micro::GetTensorData<float>(input),
+                           &pad_value, tflite::micro::GetTensorShape(output),
+                           tflite::micro::GetTensorData<float>(output));
+      }
+    } break;
```
deqiangc, line 174: I don't remember if this style is ok when importing. Just something to watch out for

deqiangc (approved):

PR #999, opened 2022-03-25
Add int8 and float SVDF for non-HiFi Xtensa

rascani:
The Xtensa accelerated SVDF implementation is only supported on HiFi, so
it falls back to the reference implementation on VP6. This was hardcoded
to an int16 implementation in xtensa/svdf.cc, as the HiFi accelerated
implementation only supports int16. This PR enables float and int8
support when falling back on the reference implementation. This is only
impacts non-HiFi Xtensa cores, such as the VP6

BUG=b/225235693


tensorflow/lite/micro/kernels/xtensa/svdf.cc
```diff
@@ -274,8 +267,43 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
+        }
+        case kTfLiteInt8: {
+          EvalInt8SvdfReference(context, node, input, weights_feature,
+                                weights_time, bias, params, activation_state,
+                                output, data);
+          return kTfLiteOk;
+          break;
+        }
+        default:
+          TF_LITE_KERNEL_LOG(context, "Type %s not currently supported.",
```
tingyan19, line 296: Prefer MicroPrintf() to TF_LITE_KERNEL_LOG()

tensorflow/lite/micro/kernels/xtensa/svdf.cc
```diff
@@ -274,8 +267,43 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
+                             TfLiteTypeGetName(weights_time->type));
+          return kTfLiteError;
+      }
+    }
+
+    default:
+      TF_LITE_KERNEL_LOG(context, "Type %s not currently supported.",
```
tingyan19, line 303: Prefer MicroPrintf() to TF_LITE_KERNEL_LOG()

tingyan19 (approved):

PR #1007, opened 2022-03-30

mansnils:
Adds int8 and int16x8 specific registrations for depthwise conv
kernel.
Person detect example is updated with int8 specific
registration for depthwise conv. It will result in a smaller memory footprint
for kernels that support this. For kernels that not support it the
default registration will be used.


tensorflow/lite/micro/kernels/cmsis_nn/depthwise_conv.cc
```diff
@@ -245,6 +250,88 @@ void EvalQuantizedPerChannel(TfLiteContext* context, TfLiteNode* node,
...
 }
 
+void EvalQuantizedPerChannel16x8(TfLiteContext* context, TfLiteNode* node,
+                                 const TfLiteDepthwiseConvParams& params,
+                                 const OpData& data,
+                                 const TfLiteEvalTensor* input,
+                                 const TfLiteEvalTensor* filter,
+                                 const TfLiteEvalTensor* bias,
+                                 TfLiteEvalTensor* output) {
+  cmsis_nn_dw_conv_params dw_conv_params;
```
deqiangc, line 260: line 261 to line 269 seems to be similar with what is in EvalQuantizedPerChannel. Does it make sense to extract this part to a single utility function such as PopulateDwConvParams()?

deqiangc: Actually all the way to line 318 seems to have some similarity

mansnils: Well yes it makes sense, however I get a code size penalty which seems counter productive for this patch. So let's do it as a macro instead, then there is no change in binary size.

deqiangc: Hello, @mansnils, thanks for the update!  When you say a code size penalty, is it correct to understand that code size increase is ON the binary using the specific registration?  Do you have a rough sense of how much?   We really would like to avoid using macro if possible. I know that some existing code still use macro, but we are trying to move away from that if possible.  I wonder whether the code size penalty comes from the common function NOT getting inlined because the compiler is seeing this common function is being called by the two flavor of eval function and the compiler does not know that only one of them will be kept by the linker. So I think this may be a case of using inline inside cpp. For gcc, you might want to use __attribute__((always_inline)) to force compiler to take it.  Thank you for the patience!
If you don't find it work, but penalty is small enough, then I still think it is worthy to keep it as a function for better maintenance.

mansnils: You are right. The always inline attribute really helps here. After that there is only a 32 byte difference for GCC, which is small enough I think.

deqiangc: Thanks for chasing this one down!

deqiangc (approved):

deqiangc (approved):

PR #1013, opened 2022-04-01
Extract buffer allocator interface from SimpleMemoryAllocator

deqiangc:
This PR abstract out the interface of SimpleMemoryAllocator into two
interface class. MicroAllocator interacts with SimpleMemoryAllocator
through these two interface classes.
This decouple will enable us to replace SimpleMemoryAllocator with other
arena buffer manager.

BUG=https://b/226971240


tensorflow/lite/micro/ibuffer_allocator.h
```diff
@@ -0,0 +1,93 @@
...
+
+  // Resizes a buffer that is previously returned by the
+  // AllocateResizableBuffer.
+  virtual TfLiteStatus ResizeBuffer(uint8_t* resizable_buf, size_t size,
+                                    size_t alignment) = 0;
+
+  // Frees up the memory occupied by the resizable buffer.
+  virtual TfLiteStatus DeallocateResizableBuffer(uint8_t* resizable_buf) = 0;
+
+  // Returns a pointer pointing to the start of the overlay memory, which are
```
tingyan19, line 73: "which are" -> "which is"?

tensorflow/lite/micro/micro_allocation_info.h
```diff
@@ -63,7 +63,8 @@ struct GraphAllocationInfo {
 // `Finish`.
 class AllocationInfoBuilder {
  public:
-  AllocationInfoBuilder(const Model* model, SimpleMemoryAllocator* allocator,
+  AllocationInfoBuilder(const Model* model,
+                        INonPersistentBufferAllocator* allocator,
```
tingyan19, line 67: Is it worthwhile calling out the "allocator" to be non-persistent? Like calling it "non_persistent_allocator", or is there a specific reason why we shouldn't do this?

tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -68,7 +69,7 @@ class MicroBuiltinDataAllocator : public BuiltinDataAllocator {
   TF_LITE_REMOVE_VIRTUAL_DELETE
 
  private:
-  SimpleMemoryAllocator* memory_allocator_;
+  IPersistentBufferAllocator* memory_allocator_;
```
tingyan19, line 72: Similar to the previous question - does it make sense to name it "persistent_memory_allocator_"?

tensorflow/lite/micro/ibuffer_allocator.h
deqiangc: Done

tensorflow/lite/micro/micro_allocation_info.h
deqiangc: Good call! I agree. Changed

tensorflow/lite/micro/micro_allocator.cc
deqiangc: Agreed. Changed

tingyan19 (approved):

PR #1018, opened 2022-04-05

mansnils:
Adds three specific registrations for softmax kernel:
- Pure int8.
- Pure int16.
- And one with int8 input and int16 output.

In order to avoid duplicating code CalculateSoftmaxParams is made
public.

### DIFF
--- a/tensorflow/lite/micro/kernels/cmsis_nn/softmax.cc
+++ b/tensorflow/lite/micro/kernels/cmsis_nn/softmax.cc
@@ -1,4 +1,4 @@
-/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


tensorflow/lite/micro/kernels/cmsis_nn/softmax.cc
```diff
@@ -28,58 +28,104 @@ limitations under the License.
...
+          op_data.softmax_params, tflite::micro::GetTensorShape(input),
           tflite::micro::GetTensorData<float>(input),
           tflite::micro::GetTensorShape(output),
           tflite::micro::GetTensorData<float>(output));
       return kTfLiteOk;
     }
-    case kTfLiteInt8:
+    case kTfLiteInt8: {
+      if (output->type == kTfLiteInt8) {
+        arm_softmax_s8(tflite::micro::GetTensorData<int8_t>(input),
```
freddan80, line 101: Would it be possible to call SoftmaxEvalInt8() here?

tensorflow/lite/micro/kernels/cmsis_nn/softmax.cc
```diff
@@ -28,58 +28,104 @@ limitations under the License.
...
+                       op_data.num_rows, op_data.row_size,
+                       op_data.softmax_params.input_multiplier,
+                       op_data.softmax_params.input_left_shift,
+                       op_data.softmax_params.diff_min,
+                       tflite::micro::GetTensorData<int8_t>(output));
+      } else {
+        arm_softmax_s8_s16(tflite::micro::GetTensorData<int8_t>(input),
```
freddan80, line 108: dito for arm_softmax_s8_s16()?

tensorflow/lite/micro/kernels/cmsis_nn/softmax.cc
```diff
@@ -28,58 +28,104 @@ limitations under the License.
...
+      return kTfLiteOk;
+    }
     case kTfLiteInt16: {
-      SoftmaxQuantized(input, output, data);
+      const cmsis_nn_softmax_lut_s16 softmax_params = {
+          .exp_lut = op_data.softmax_params.exp_lut,
+          .one_by_one_lut = op_data.softmax_params.one_over_one_plus_x_lut};
+
+      TFLITE_DCHECK_EQ(
+          arm_softmax_s16(
```
freddan80, line 123: dito for SoftmaxEvalInt16()

tensorflow/lite/micro/kernels/cmsis_nn/softmax.cc
mansnils: I prefer not since it seems binary size will increase then.

tensorflow/lite/micro/kernels/cmsis_nn/softmax.cc
mansnils: same comment here

tensorflow/lite/micro/kernels/cmsis_nn/softmax.cc
mansnils: same

tensorflow/lite/micro/kernels/cmsis_nn/softmax.cc
```diff
@@ -89,17 +135,111 @@ TfLiteStatus SoftmaxEval(TfLiteContext* context, TfLiteNode* node) {
...
+TfLiteRegistration Register_SOFTMAX_INT16() {
+  softmax_registration.init = Init;
+  softmax_registration.free = nullptr;
+  softmax_registration.prepare = Prepare;
+  softmax_registration.invoke = SoftmaxEvalInt16;
+  softmax_registration.profiling_string = nullptr;
+  softmax_registration.builtin_code = 0;
+  softmax_registration.custom_name = nullptr;
+  softmax_registration.version = 0;
+  return softmax_registration;
```
paulinesho, line 242: Is this change a style preference or is there a reason why we shouldn't use brace initialization here?

mansnils: The reason is that the symbol softmax_registration is needed in this context for the linker to drop unused symbols. There is a TODO(b/169801227). So it is needed to get the code size benefits of this quantization specific registration.

paulinesho: @mansnils Thanks for sharing that historical context. We're moving towards a common API for kernel registration in #1084. The good thing is that the new approach also reduces code size (see PR description for numbers) compared to both brace initialization and static initialization. I suggest we make the same change here for `softmax` if #1084 is merged before this.

mansnils: Well yes, I tried with and without the new common kernel registration API for cmsis-nn/fully_conneted.cc. Then I tried with and without the int8 specific registration in person detect example: https://github.com/tensorflow/tflite-micro/blob/10b1223343303b45e6f62a433e25efb5f87958d7/tensorflow/lite/micro/examples/person_detection/main_functions.cc#L78
And the same code size difference is seen. I can't see the internals of TODO(b/169801227) but I guess it can be closed then.

tensorflow/lite/micro/kernels/softmax_common.cc
```diff
@@ -28,11 +28,48 @@ namespace {
...
 const int kInt16LUTArraySize = 513;
 
+}  // namespace
+
 TfLiteStatus CalculateSoftmaxParams(TfLiteContext* context,
                                     const TfLiteTensor* input,
                                     TfLiteTensor* output,
                                     const TfLiteSoftmaxParams* params,
                                     SoftmaxParams* op_data) {
+  // Only allocate LUTs for KTfLiteInt16 data type
```
deqiangc, line 38: Could line 38 to line 71 be extracted as a separate function such as InitializeLutForInt16? This is a bit long and the function name will make it more readable.

mansnils: Done

tensorflow/lite/micro/kernels/softmax_common.cc
```diff
@@ -28,11 +28,58 @@ namespace {
 // Softmax parameter data that persists in user_data
 const int kInt16LUTArraySize = 513;
 
+__attribute__((always_inline)) inline TfLiteStatus InitializeLutForInt16(
```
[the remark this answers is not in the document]
mansnils: Ok but person detect binary increased by 296 bytes when removing the inline line for GCC. No difference for armclang.

Tested like this:
make -j$(nproc) -f tensorflow/lite/micro/tools/make/Makefile TOOLCHAIN=gcc OPTIMIZED_KERNEL_DIR=cmsis_nn TARGET=cortex_m_corstone_300 TARGET_ARCH=cortex-m55 person_detection_test && ll tensorflow/lite/micro/tools/make/gen/cortex_m_corstone_300_cortex-m55_default/bin/

Should we still remove it?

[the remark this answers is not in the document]
deqiangc: Thanks for getting back Mans! I still prefer to remove it. In the mean time, we can create an issue to track this 296 bytes difference.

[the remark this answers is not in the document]
mansnils: Done

deqiangc (approved):

Issue #909, opened 2022-02-04
Add armc6 build in TFLM CI using Github-actions


annietllnd:
We have identified bugs when building TFLM using the armc6 compiler. Thus, we would like to add a way to regression test TFLM using armc6. It is possible to run the relevant tests on our virtual hardware using AWS. By providing Github secrets and a Github actions workflow, the verification can be incorporated into the regular CI. This will guarantee correctness of execution on relevant Cortex-M platforms, and identify performance deviations during validation. A PR will be opened and tagged here.

annietllnd:
See more at the [VHT-AMI](https://github.com/ARM-software/VHT-AMI) repository

annietllnd:
Hi, this is a duplicate of #1021 and since that's been merged, we may close this PR. Thanks Matthias!

PR #1021, opened 2022-04-05
Arm virtual hardware ci

MatthiasHertel80:
Replaces PR#927. Updates for newer Arm Virtual Hardware Client.

BUG=#69


.github/workflows/cortex_m_virtual_hardware.yml
```diff
@@ -0,0 +1,71 @@
...
+          python-version: '3.10'
+      - name: Install AVH Client for Python
+        run: |
+          pip install git+https://github.com/ARM-software/avhclient.git@v0.1.1
+      - name: Execute test suite on Virtual Hardware at AWS
+        run: |
+          avhclient -b aws execute --specfile ./tensorflow/lite/micro/cortex_m_generic/avh.yml
+      - name: Fetch results from VHT
+        run: |
+          cat ./tensorflow/lite/micro/cortex_m_generic/cortex_m_generic.log
```
deqiangc, line 48: If there are errors in running the test in virtual machine,  how are errors manifested? I am not that familiar with github workflow yet, but  I think we need a return error code or something to stop github machinery to treat it as pass instead of just cat (cat may not return error even if there are errors in log).

tensorflow/lite/micro/cortex_m_corstone_300/avh.yml
```diff
@@ -0,0 +1,13 @@
+name: "TensorFlow Lite Micro Corstone-300 ArmClang"
```
deqiangc, line 1: I feel this file might be better placed in .github/cortext_mcorstone_300 folder instead of tensorflow/lite/micro/cortex_m_corstone_300.  What is in tensorflow/lite/micro/cortex_m_corstone_300 are necessary initialization codes for building targets. This file is a configuration and is more about executing task in virtual machine remotely.

.github/workflows/cortex_m_virtual_hardware.yml
deqiangc: Having said that, I am fine with us going with cat first to establish the communication mechanism and have a separate step of failing automatically

deqiangc (commented):
Thanks Matthias! This is great! I love that we will have the capability run those test remotely from tflm repo and increase coverage!

Do you mind adding some more descriptions in the commit? Things such as:

This PR allows the TFLM repo to trigger xxx test remotely in xxx. The goal is to catch failure such as #909 asap. 
This remote execution is through Arm Virtual Hardware Client..
At this PR, the errors are manifiested through....
A separate PR will be added in future to report the failure as a workflow failure.

deqiangc (changes requested):
I forget to click request changes earlier on. Please see comments

.github/workflows/cortex_m_virtual_hardware.yml
MatthiasHertel80: The idea is that the PR should mainly establish the infrastructure. It will actually fail the step executed on the remote instance, as the errorlevel of executions there is passed back to the client and thus to the github runner.
More tests should be added over time, and with those we will need more detailed test reporting and failing mechanisms.

tensorflow/lite/micro/cortex_m_corstone_300/avh.yml
MatthiasHertel80: I agree that the source folder is not ideal. I am a bit afraid that .github with its special purpose might cause issues someday, when Github accidently picks up the yml files and tries to interpret them. What do you think about: tensorflow/lite/micro/tools/arm-virtual-hardware/* as a place where we could place all related files (currently it will be a yml for each AVH workflow)?

.github/workflows/cortex_m_virtual_hardware.yml
deqiangc: ACKED. agree with the first step and priority should be in establish the infra

tensorflow/lite/micro/cortex_m_corstone_300/avh.yml
deqiangc: Sounds good to me in general. What do you think about: tensorflow/lite/micro/tools/github/arm_virtual_hardware/?  Prefer to use underscore _ instead of - to be consistent with the convention inside this folder.

MatthiasHertel80: Done :)

deqiangc (dismissed):

tensorflow/lite/micro/cortex_m_corstone_300/avh.yml
deqiangc: @MatthiasHertel80 I have approved the PR and noticed that it did not get check in b/c formatting issues. Can you please run tensorflow/lite/micro/tools/ci_build/test_code_style.sh --fix_formatting? That should get the format clear. You can also run tensorflow/lite/micro/tools/ci_build/test_code_style.sh afterwards to make sure there are not format issues.

deqiangc (approved):

tensorflow/lite/micro/cortex_m_corstone_300/avh.yml
deqiangc: @MatthiasHertel80 It appears that your branch is out of date and that appears to be the reason that it cannot be merged (or so I guess since this is the only signal on the page).  I tried to update it, but for some reasons, your branch is not updatable by me.  Could you pleas sync to main and push again?

PR #1023, opened 2022-04-05
Minor update to `CONTRIBUTING.md` on cpplint installation.

paulinesho:
`cpplint` can be installed via `pip` as well as getting a local copy.
Improving documentation to include the `pip` alternative.

BUG=


CONTRIBUTING.md
```diff
@@ -166,6 +166,11 @@ Below are some tips that might be useful and improve the development experience.
   is the what we do for the TFLM continuous integration Docker container.
 
 * Get a copy of [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint)
+  or install it:
+
+  ```
+  pip3 install cpplint
```
deqiangc, line 172: I think other documents in this repo typically just refers "pip" instead of pip3.

paulinesho: Actually just one line below this references `pip3`. I can change that to `pip` too if that's preferred.

deqiangc: oh. I see. I personally think pip is better because it is generic (typically just a sym link) while pip3 is explicitly for python3.  However either way is ok as long as we are consistent in the same doc

paulinesho: We do seem to favor pip in this repo so I'll change both references here since I'm touching this file.

PR #1026, opened 2022-04-06
Quantization specific registration for CMSIS-NN fully connected int16

mansnils:
Adds int16 support, int16 specific registration and unit test (ported from TFL) for CMSIS-NN fully connected.

### DIFF
--- a/tensorflow/lite/micro/kernels/cmsis_nn/fully_connected.cc
+++ b/tensorflow/lite/micro/kernels/cmsis_nn/fully_connected.cc
@@ -39,6 +39,10 @@ struct OpData {
 
   // Index to buffer for optimizations if applicable.
   int buffer_idx;
+
+  int32_t batches;
+  int32_t accum_depth;
+  int32_t output_depth;
 };
 
 // TODO(b/169801227): This global struct is needed for the linker to drop unused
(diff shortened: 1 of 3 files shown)


tensorflow/lite/micro/kernels/fully_connected_test.cc
```diff
@@ -316,16 +316,16 @@ TfLiteStatus TestFullyConnectedFloat(
 }
 #endif
 
-template <typename T>
+template <typename T, typename biasT>
```
deqiangc, line 319: can we use <typename dataT, typename weightT, typename biasT>? Basically, instead of int8_t* wieghts_quantized below, use weightT weights_quantized to preserve generality?

tensorflow/lite/micro/kernels/fully_connected_test.cc
```diff
@@ -435,6 +435,36 @@ TF_LITE_MICRO_TEST(SimpleTestQuantizedInt8) {
       kTfLiteOk);
 }
 
+#if defined(CMSIS_NN)
```
deqiangc, line 438: It would be really great if we can also update the reference to support 16x8 and run test for both reference and CMSIS_NN.  Is that possible here?

tensorflow/lite/micro/kernels/fully_connected_test.cc
mansnils: Done

tensorflow/lite/micro/kernels/fully_connected_test.cc
mansnils: Done

tensorflow/lite/micro/kernels/cmsis_nn/fully_connected.cc
```diff
@@ -247,6 +245,70 @@ TfLiteStatus EvalQuantizedInt8(TfLiteContext* context, TfLiteNode* node,
...
+                                const TfLiteEvalTensor* input,
+                                const TfLiteEvalTensor* filter,
+                                const TfLiteEvalTensor* bias,
+                                TfLiteEvalTensor* output) {
+  cmsis_nn_per_tensor_quant_params quant_params;
+  quant_params.multiplier = data.reference_op_data.output_multiplier;
+  quant_params.shift = data.reference_op_data.output_shift;
+
+  cmsis_nn_dims input_dims;
+  input_dims.n = data.batches;
```
deqiangc, line 259: seems line 258 to line 274 are similar with what is in EvalQuantized that they can beneifit from one single function. WDYT?

mansnils: Yes it makes sense

mansnils: Done

deqiangc (approved):

PR #1069, opened 2022-04-21
make TARGET_ARCH and BUILD_TYPE variable could be overrided in target makefile

fanghuaqi:
In the latest source code of tflite-micro, `TARGET_ARCH` and `BUILD_TYPE` could not be override by Makefile of selected **TARGET**, this commit is trying to fix this issue introduced by PR #432

Thanks
Huaqi

### DIFF
--- a/tensorflow/lite/micro/tools/make/Makefile
+++ b/tensorflow/lite/micro/tools/make/Makefile
@@ -235,17 +235,6 @@ endif
 # runtime that can be linked in to other programs.
 MICROLITE_LIB_NAME := libtensorflow-microlite.a
 
-# Where compiled objects are stored.
-GENDIR := $(MAKEFILE_DIR)/gen/$(TARGET)_$(TARGET_ARCH)_$(BUILD_TYPE)/
-CORE_OBJDIR := $(GENDIR)obj/core/
-KERNEL_OBJDIR := $(GENDIR)obj/kernels/
-THIRD_PARTY_KERNEL_OBJDIR := $(GENDIR)obj/third_party_kernels/
-THIRD_PARTY_OBJDIR := $(GENDIR)obj/third_party/
-GENERATED_SRCS_DIR := $(GENDIR)genfiles/
-BINDIR := $(GENDIR)bin/
-LIBDIR := $(GENDIR)lib/
-PRJDIR := $(GENDIR)prj/
-
 # These two must be defined before we include the target specific Makefile.inc
 # because we filter out the examples that are not supported for those targets.
 # See targets/xtensa_xpg_makefile.inc for an example.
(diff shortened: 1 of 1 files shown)


paulinesho:
Is there a scenario where this command doesn't meet all your needs? `make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=riscv32_mcu BUILD_TYPE=release build`? 

It seems that there's going to be some contention with this change due to #422. Currently our CI scripts build with the above convention (specifying TARGET/BUILD_TYPE in command line).

fanghuaqi:
> Is there a scenario where this command doesn't meet all your needs? `make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=riscv32_mcu BUILD_TYPE=release build`?
> 
> It seems that there's going to be some contention with this change due to #422. Currently our CI scripts build with the above convention (specifying TARGET/BUILD_TYPE in command line).

I want to overwrite the TARGET and BUILD_TYPE in target makefile, not via make, the default TARGET in current is x86， which is not suitable for non-x86 target.

fanghuaqi:
Could you @turbotoribio help to review this PR, thanks.

deqiangc:
> > Is there a scenario where this command doesn't meet all your needs? `make -j8 -f tensorflow/lite/micro/tools/make/Makefile TARGET=riscv32_mcu BUILD_TYPE=release build`?
> > It seems that there's going to be some contention with this change due to #422. Currently our CI scripts build with the above convention (specifying TARGET/BUILD_TYPE in command line).
> 
> I want to overwrite the TARGET and BUILD_TYPE in target makefile, not via make, the default TARGET in current is x86， which is not suitable for non-x86 target.

Conceptually, I view BUILD_TYPE/TARGET as part of global and universal env setup.  Makefile based build system does NOT have a good scope concept and in my humble opinion, it is best to keep such global and universal env variables to be at the top level Makefile instead of specifying them at local target Makefile. In this case, it seems that Pausline already suggested an alternative: those flags can be pass in from the command line; so I would prefer that we do not add this capability at this moment

fanghuaqi:
Hi @deqiangc , for my changes in this PR didn't affect `BUILD_TYPE/TARGET` to be global and universal env setup, user can still pass `BUILD_TYPE=xx TARGET=xxx` to overwrite default setting in Makefile, it just make it possible to allow default settings for different target, for eg.

I don't think the default target for `riscv32_mcu` the `TARGET_ARCH` is `x86_64` is correct.

image

paulinesho:
@fanghuaqi We are recommending passing `TARGET_ARCH` in command line instead of overriding it in `{TARGET}_makefile.inc`. This is because these flags are currently used as inputs to some `{TARGET}_makefile.inc` (see #422). We cannot also support overriding it in target-specific makefiles as it causes a circular dependency. Supporting both of these use cases would require refactoring the Makefile which is currently unplanned. Please create a new GitHub issue if your infrastructure cannot support this workflow for some reason, and we can work on a solution there. For now I'll close this PR. 

Thanks for your understanding!

fanghuaqi:
Thank you for the explaination, but I still think this should be fixed, it is not right for `TARGET_ARCH` to be `x86_64` for `riscv_mcu`, and it should be default value in various different target.

And in https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/tools/make/targets/cortex_m_corstone_300_makefile.inc, I didn't see any usage of ` GENDIR and other make variables`.

Issue #1068, opened 2022-04-21
Port elementwise RSQRT int8 support


annietllnd:
Hi! We will soon submit a PR that covers elementwise RSQRT int8 and ABS int8 and int16 (#190). The work is based on an old PR #204 containing both operators, where the corresponding issue does not track the RSQRT part. Thus, I decided to open a new issue for RSQRT to avoid confusion. 

Thanks for your time!
Annie

PR #204, opened 2021-06-21
Port int8 and int16 in ABS kernel to micro

patriklaurell:
@advaitjain this PR ports int8 and int16 support to ABS kernel.

BUG=Fixes #190

Issue #190, opened 2021-06-17
Port operator ABS for int8 and int16 to TFLu


patriklaurell:
@advaitjain This tracks porting of int8 and int16 support in ABS kernel.

It will be delivered as one PR.

annietllnd:
I will shortly open a new PR to track the new changes, as advised in discussion on TFL PR [#50836](https://github.com/tensorflow/tensorflow/pull/50836)

PR #1072, opened 2022-04-21
add int8 and int16 support in elementwise ABS and RSQRT

annietllnd:
Hi! PR #204 was originally intended to be split between both Tensorflow and TFLM repository with shared code in a elementwise_portable header file. But with the new micro context API, the shared code is incompatible with Tensorflow. Thus, we make changes to TFLM only, and handle the porting in Tensorflow as a separate action. As a result, this PR solves #1068 and #190 

That means the PR is meant to do two things. Firstly, it's adding int8 support for the elementwise RSQRT kernel. Secondly, it adds int8 and int16 support for elementwise ABS. As the patch is quite large and some time has passed between "rounds", I'm looking forward to any feedback and questions :)

All the best,
Annie

BUG=#1068 and #190


annietllnd:
@deqiangc Hi Deqiang, all checks have passed so this PR should be ready for review! All the best, Annie

tensorflow/lite/micro/kernels/elementwise.cc
```diff
@@ -53,38 +117,206 @@ TfLiteStatus GenericPrepare(TfLiteContext* context, TfLiteNode* node) {
...
+  int inv_sqrt_shift;
+  GetInvSqrtQuantizedMultiplierExp(value, kReverseShift, &inv_sqrt_multiplier,
+                                   &inv_sqrt_shift);
+  const int32_t data = tflite::MultiplyByQuantizedMultiplier(
+      static_cast<int32_t>(1), inv_sqrt_multiplier, inv_sqrt_shift + kShift);
+  const int32_t output =
+      tflite::MultiplyByQuantizedMultiplier(data, op_data->multiplier,
+                                            op_data->shift - kShift) +
+      op_data->output_offset;
+  return static_cast<T>(std::fmin(std::fmax(output, kMin), kMax));
```
tingyan19, line 275: I'm wondering why are std::fmin() and std::fmax() used here instead of std::min() and std::max(). 

I noticed that the TFLite version uses std::min() and std::max(). 

Also, all parameters (kMin, kMax and output) are integers instead of floating-point numbers.

annietllnd: Hi, thanks for the comment. Originally, I put the fmin/fmax to resolve a unmatching type issue. I will push a solution that adds back std::min/std::max and instead casts the arguments. Let me know if this works. All the best, Annie

tingyan19 (approved):

advaitjain:
@Mergifyio refresh

PR #1076, opened 2022-04-22
Add support for conv 16x8 int32 bias

paulinesho:
Currently, the 16x8 conv op implicitly assumes int64 biases. This
behavior ignores the converter option
`_experimental_full_integer_quantization_bias_type`. The actual bias
size/type can be retrieved from the model via `GetEvalInput()->type`.
Supporting int32 instead of int64 is expected to lower latency on
average.

Verified that bias->type correctly reflects the converter option set
during conversion. Replicated basic unit tests for int32 biases.

BUG=b/230125424


tensorflow/lite/micro/kernels/conv.cc
```diff
@@ -74,16 +74,40 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
+              tflite::micro::GetTensorShape(filter),
+              tflite::micro::GetTensorData<int8_t>(filter),
+              tflite::micro::GetTensorShape(bias),
+              tflite::micro::GetTensorData<std::int64_t>(bias),
+              tflite::micro::GetTensorShape(output),
+              tflite::micro::GetTensorData<int16_t>(output));
+          break;
+        }
+        default:
+          TF_LITE_KERNEL_LOG(context, "Bias type %s (%d) not supported.",
```
tingyan19, line 107: MicroPrintf() is preferred than TF_LITE_KERNEL_LOG()

tensorflow/lite/micro/kernels/conv_test_common.cc
```diff
@@ -220,5 +220,23 @@ TfLiteStatus TestConvQuantizedPerChannel(
       conv_params, registration, output_data);
 }
 
+TfLiteStatus TestConvQuantizedPerChannel(
```
deqiangc, line 223: Could you please add some comments here? something along the line like "Test XXXX with 16 bit activation, 8 bit weight, 32 bit accumuator".  I would add similiar comments on the other two previous functions. 

This is a style suggestion and I tend to be on verbose side. Feel free to push back.

tensorflow/lite/micro/kernels/conv.cc
paulinesho: Fixed TF_LITE_KERNEL_LOG -> MicroPrintf.

tensorflow/lite/micro/kernels/conv_test_common.cc
paulinesho: I agree this isn't very readable as is. Added the suggested comment to all 3 functions.

deqiangc (dismissed):

deqiangc (approved):

PR #1080, opened 2022-04-26
Enable five dimension support in concatenation and export max size in runtime_shape.h to public

deqiangc:
Enable five dimension support in concatenation and export max size in runtime_shape.h to public

There are models that use concatenation op to combine 5 dimension
tensors and then use reshape to cast it to lower dimension.

BUG=https://b/230028089


deqiangc (commented):
@njeffrie 

FYI and comments are welcome!

tingyan19 (approved):

PR #1084, opened 2022-04-27
Enable -Wmissing-field-initializers and update all registration

paulinesho:
In https://github.com/tensorflow/tflite-micro/pull/1082/, we disabled
the compiler error as a workaround for a successful upstream sync from
TFLite. This PR re-enables the error after addressing all errors. A new
`RegisterOp` API is added, and all prior TfLiteRegistration struct
initializations are updated to use the new API. At the minimum, this
ensures that such additions in future will only require an update to
RegisterOp if TFLM doesn't use that field.

BUG=b/230507399

A positive side effect from this change is that code size is also reduced
by about 800 bytes. Similar reduction applies to hexagon and xtensa.

| text   | data | bss   | dec    | hex   | notes                                                 |
|--------|------|-------|--------|-------|-------------------------------------------------------|
| 418124 | 800  | 18860 | 437784 | 6ae18 | cmsis_nn before                                       |
| 417592 | 728  | 18940 | 437260 | 6ac0c | cmsis_nn after                                        |
| 417372 | 728  | 18868 | 436968 | 6aae8 | cmsis_nn after + replaced static init for conv and fc |
| 576639 | 6352 | 19096 | 602087 | 92fe7 | x86 before                                            |
| 575960 | 6208 | 19256 | 601424 | 92d50 | x86 after                                             |


tensorflow/lite/micro/kernels/xtensa/unidirectional_sequence_lstm.cc
```diff
@@ -407,10 +407,6 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
   return op_data;
 }
 
-void Free(TfLiteContext* context, void* buffer) {
```
paulinesho, line 410: @tingyan19 @deqiangc FYI `Free` is removed here as we discussed. Please review.

tensorflow/lite/micro/kernels/kernel_util.h
```diff
@@ -27,6 +27,11 @@ limitations under the License.
 namespace tflite {
 namespace micro {
 
+TfLiteRegistration RegisterOp(
```
deqiangc, line 30: Do you think this is more like RegisterOp (this function does not really register, it just create a registration that the caller will use to register)  or InitializeTfLiteRegistration or CreateTfLiteRegistration?
Realizing how big this PR is, please don't rush to make changes. I am open to discuss on this.

paulinesho: The `RegisterOp` name made sense to me when I opened the PR because it's consistent with callers' convention `Register_<opname>`, i.e all `Register_<opname>` functions that exist before this PR also return a registrations but don't really register anything. I agree with you though all these are not really accurate.

I think `CreateTfLiteRegistration` makes the most sense. But if we're not going to change the callers from `Register_<opname>` to `Create<opname>TfLiteRegistration`, maybe it'll just add another source of confusion. I think there's some value in keeping a consistent name as the caller, so either `RegisterAdd() -> RegisterOp()` or `CreateAddRegistration() -> CreateRegistration()` because conceptually they're the same.

Renaming the new `RegisterOp` is not a lot of effort even though this PR is big, I'm willing to go either way.

deqiangc: ok. Let us stick with with RegisterOp then. Inside Register_opname makes sense because it is directly called by tflm.

deqiangc (approved):

Issue #920, opened 2022-02-09
Porting Reference LSTM Op from Lite to Micro


tingyan19:
This issue is to track efforts to port the reference LSTM kernel from TF Lite to TFLM.

tingyan19:
following instructions in https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/porting_reference_ops.md

christophsiegl:
What is the current status of the implementation? The unidirectionalSequenceLSTM is still missing in all_ops_resolver.cc ?

PR #1091, opened 2022-04-29
Fix LSTM test bug with wrong output dimension

tingyan19:
The output tensor dimensions for the integer LSTM tests were specified incorrectly. This PR fixes this bug, and added guards in the kernel to protect mistakenly specified output dimensions. It also corrects a minor issue that affects the lifetime of a pointer.

BUG=#920, http://b/218742676

### DIFF
--- a/tensorflow/lite/micro/kernels/unidirectional_sequence_lstm.cc
+++ b/tensorflow/lite/micro/kernels/unidirectional_sequence_lstm.cc
@@ -1267,12 +1267,10 @@ TfLiteStatus UnidirectionalSequenceLstmPrepare(TfLiteContext* context,
   TF_LITE_ENSURE_EQ(context, NumElements(cell_state), n_batch * n_cell);
 
   // Check the shape of output tensor against that of input tensor
-  TF_LITE_ENSURE_EQ(context, input->dims->size, output->dims->size);
-  for (int i = 0; i < input->dims->size - 1; ++i) {
-    TF_LITE_ENSURE_EQ(context, input->dims->data[i], output->dims->data[i]);
-  }
-  TF_LITE_ENSURE_EQ(context, output->dims->data[output->dims->size - 1],
-                    n_output);
+  TF_LITE_ENSURE_EQ(context, output->dims->size, 3);
+  TF_LITE_ENSURE_EQ(context, input->dims->data[0], output->dims->data[0]);
+  TF_LITE_ENSURE_EQ(context, input->dims->data[1], output->dims->data[1]);
+  TF_LITE_ENSURE_EQ(context, output->dims->data[2], n_output);
 
   if (is_integer) {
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/kernels/unidirectional_sequence_lstm.cc
```diff
@@ -1266,6 +1266,14 @@ TfLiteStatus UnidirectionalSequenceLstmPrepare(TfLiteContext* context,
   TF_LITE_ENSURE_EQ(context, NumElements(output_state), n_batch * n_output);
   TF_LITE_ENSURE_EQ(context, NumElements(cell_state), n_batch * n_cell);
 
+  // Check the shape of output tensor against that of input tensor
+  TF_LITE_ENSURE_EQ(context, input->dims->size, output->dims->size);
```
deqiangc, line 1270: Could you please elaborate this a bit? I am not entirely following the reason behind this. 

input->dims->size == output-dims->size
and the first N element of input->dims->data and output->dims->data are the same.
Then the last element of output->dims->data is the same as n_output.

However, it seems n_output = input->dims->data[2].
So, is input->dims->size guaranteed to be 3 or something? 
If yes, check do not need loop.
If no,  this guaratee should be called out.

WDYT?

tingyan19: In this implementation it does fix the dims->size to be 3 and directly assign input->dims->data[2] to n_input.

I was trying to follow the logic in the TFLite implementation as close as possible:

  // Resize the output tensors.
  TfLiteIntArray* output_size = TfLiteIntArrayCopy(input->dims);
  output_size->data[input->dims->size - 1] = n_output;
  TF_LITE_ENSURE_OK(context,
                    context->ResizeTensor(context, output, output_size));

Of course we do not do resizing and just check the dimensions. 

I kind of think the loop makes it more generic and extensible in case a higher dimension is added in the future, in comparison with the following:

TF_LITE_ENSURE_EQ(context, input->dims->data[0], output->dims->data[0]);
TF_LITE_ENSURE_EQ(context, input->dims->data[1], output->dims->data[1]);

Please let me know your thoughts on this. Thanks for reviewing it promptly.

deqiangc: I might prefer the simpler approach here just to avoid the confusion for future code readers. But if you really like the more general way, that is fine too (but please add the explanation in the code to address potential confusion like mine).

deqiangc (approved):

tensorflow/lite/micro/kernels/unidirectional_sequence_lstm.cc
tingyan19: Changed to the simpler approach and added a check to make sure the dimension size is 3.

PR #1092, opened 2022-04-30
Add float and hybrid tests for LSTM

tingyan19:
PR #1044 tested integer input LSTM. This PR adds tests for float and hybrid operation modes of the reference kernel. 

We strive to match test cases in the file to https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/unidirectional_sequence_lstm_test.cc. Input and golden vectors are identical although the TFLite version uses a different test framework.

BUG=#920, http://b/218742676


tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test.cc
```diff
@@ -547,218 +2417,409 @@ QuantizationParams SetQuantizationParams(float f_min, float f_max) {
...
     int n_batch, int n_input, int n_cell, int n_output, int sequence_length,
     bool time_major, bool use_cifg, bool use_peephole,
     bool use_projection_weights, bool use_projection_bias, bool use_layer_norm,
-    bool use_8x8_8_implementation, const float ranges[][2],
-    float intermediate_scale[5][2], int intermediate_zp[5][2],
-    TfLiteAffineQuantization intermediate_qparam[5],
+    float cell_clip, float proj_clip,
 
-    const float* input, int8_t* input_quantized,
+    const float* input_original, float* input,
```
deqiangc, line 2444: not strictly required, but I personally favor one def per line.   https://yaqs.corp.google.com/eng/q/5304356331585536.

tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test.cc
```diff
@@ -34,6 +35,1854 @@ constexpr int max_num_input_tensors = 24;
 constexpr int output_tensor_index = max_num_input_tensors;
 constexpr int intermediate_tensor_base = max_num_input_tensors + 1;
 
+// data for no CIFG, no peephole, no projection, and no clipping
+
```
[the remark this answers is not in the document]
tingyan19: Removed.

tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test.cc
```diff
@@ -918,198 +2946,2618 @@ void TestUnidirectionalSequenceLstmInteger(
...
+      tflite::testing::golden_output_no_cifg_peephole_proj_clipping_original,
+      tflite::testing::golden_output_no_cifg_peephole_proj_clipping,
+
+      /*tolerance=*/0.00467,
+      /*input_output_batch_major=*/false,
+      /*weight_type=*/kTfLiteUInt8);
+}
+
+TF_LITE_MICRO_TEST(UndrctnlSqncLstmFloatNoCifgPphlPrjBiasClpTest) {
+  tflite::testing::TestUnidirectionalSequenceLstmFloat(
```
deqiangc, line 5434: just thinking outloud, what is your opinion of defining a TestConfig struct and pass in that test config to the test function? Would that make it better to read? I feel too many input arguments are difficult to read, but also recognize there may be trade-off here

tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test.cc
```diff
@@ -34,6 +35,1854 @@ constexpr int max_num_input_tensors = 24;
...
 
+// data for no CIFG, no peephole, no projection, and no clipping
+
+constexpr int n_batch_no_cifg_no_peephole_no_proj_no_clipping = 1;
+constexpr int n_input_no_cifg_no_peephole_no_proj_no_clipping = 2;
+// n_cell and n_output have the same size when there is no projection.
+constexpr int n_cell_no_cifg_no_peephole_no_proj_no_clipping = 4;
+constexpr int n_output_no_cifg_no_peephole_no_proj_no_clipping = 4;
+constexpr int sequence_length_no_cifg_no_peephole_no_proj_no_clipping = 3;
+
```
deqiangc, line 45: If we define a TestConfig (see later comments), then we don't need to define all these individual constants that are long. We can just TestConfig NoCifgNoPeeholeNoProjNoClippingTestConfig.xxx = xxxx? WDYT?

tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test.cc
tingyan19: I agree with the one def per line rule for variable declarations, but this rule may not apply to function parameters. I tried to separate them into individual lines but clang-format would always group them back into one line.

tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test.cc
tingyan19: Thanks. Great suggestion. I will change this PR to draft and refactor.

tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test.cc
tingyan19: Same as above.

tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test.cc
```diff
@@ -30,446 +30,2762 @@ namespace {
...
-TfLiteAffineQuantization intermediate_qparam_integer_peephole[5];
-
-// Expected outputs, n_batch * sequence_length * n_output
-const int8_t expected_output_integer_peephole[n_batch_integer_peephole *
-                                              sequence_length_integer_peephole *
-                                              n_output_integer_peephole] = {
-    127,  127, -16, -21, 127, 127, 23,   127, 127,
-    -128, 127, 127, 127, 127, 127, -128, 127, 127,
-};
+template <typename T>
```
deqiangc, line 2757: Thanks for the change!  Looks very nice. sorry about nitting on this.  Do you think it is possible to split the structure initializations before this line into a separate cc/h file. On top of my head, new files may look like: 
a unidirectional_sequence_lstm_test_config.h/cc: contains initialization of test structures. 

The benefit of this is that future reader of this test would be able to see the real functional test code faster (right now, have to scroll down 2.7k lines).   

WDYT?

tingyan19: Thanks for the suggestion. Since this included file includes a lot of variable definitions (not just declarations), it is not a typical .h header file. It probably should be a .inc file as a textually included file as mentioned in https://google.github.io/styleguide/cppguide.html#Self_contained_Headers and https://google.github.io/styleguide/cppguide.html#File_Names - please let me know you opinion on this.

deqiangc (approved):
Thanks so much for the changes!

Issue #216, opened 2020-11-16
Remove support for asymmetric uint8 quantization in Tensorflow Lite Micro


advaitjain:
@tensorflow/micro

As described in the [TensorFLow Lite 8 bit quantization specification](https://www.tensorflow.org/lite/performance/quantization_spec):
*"Note: In the past our quantization tooling used per-tensor, asymmetric, uint8 quantization. New tooling, reference kernels, and optimized kernels for 8-bit quantization will use this spec."*

 * [int8 support over uint8 ](https://www.tensorflow.org/lite/performance/quantization_spec#signed_integer_vs_unsigned_integer)
 * [Asymmetric activations and symmetric weights](https://www.tensorflow.org/lite/performance/quantization_spec#symmetric_vs_asymmetric)

More detailed analysis can be read in [this whitepaper](http://go/arxiv/1806.08342).

TfLite Micro will be removing support for the asymmetric uint8 quantization. This will enable TFLM to focus on the fully supported quantization tooling, avoid sending users down the uint8 path and also have the potential for reduced binary size.

Note that this change is only for TfLite Micro and not TfLite.

advaitjain:
Some more context: https://github.com/tensorflow/tensorflow/issues/38285#issuecomment-635533037

PR #1112, opened 2022-05-09
Add Uint8 Dequantize

mbrooksx:
While TFLM still doesn't allow uint8 quantization (https://github.com/tensorflow/tflite-micro/issues/216), there are
use-cases where having uint8 for quantize/dequantize/requantize can
enable legacy models. Restore support for uint8 dequantize.

Tested:
dequantize_test, running TF1.x SSD models on Coral Dev Board Micro

BUG=b/213518048


paulinesho:
Added `BUG=` because CI requires that info. (See https://github.com/tensorflow/tflite-micro/blob/main/CONTRIBUTING.md)

paulinesho:
Looks like your build failed because of code style, not sure why the pre-push hook didn't prevent this. You can fix it by doing `tensorflow/lite/micro/tools/ci_build/test_code_style.sh --fix_formatting` and repushing.

paulinesho (approved):

PR #1113, opened 2022-05-09
Fix Makefile bug where 3P downloads run unconditionally

paulinesho:
If the third_party_downloads directory is cleaned out, the next time any
make command is run, those 3P downloads will be forced to execute, even
if it's just `make clean`. This doesn't make a lot of sense, as 3P
  libraries should only be downloaded when necessary.

BUG=b/232000083


paulinesho:
This is potentially opening an old can of worms (just found b/143904317). Reverting back to draft for now.

advaitjain:
Closing for now. Let's reopen if needed.

PR #1114, opened 2022-05-09
Add GetOptionalTensorData API

paulinesho:
BUG=b/232005813

Bias tensors can be optional for certain ops. The existing GetTensorData
API is modified to assert upon receiving a nullptr, optional tensors
should be retrieved via GetOptionalTensorData now. The new API should
make it explicit that a tensor is optional and so check its return
value. Only reference kernels are updated in this PR.

### DIFF
--- a/tensorflow/lite/micro/kernels/kernel_util.h
+++ b/tensorflow/lite/micro/kernels/kernel_util.h
@@ -45,17 +45,31 @@ const TfLiteEvalTensor* GetEvalInput(const TfLiteContext* context,
 TfLiteEvalTensor* GetEvalOutput(const TfLiteContext* context,
                                 const TfLiteNode* node, int index);
 
-// Returns data for a TfLiteEvalTensor struct.
+// Returns data for a TfLiteEvalTensor struct that are expected to exist.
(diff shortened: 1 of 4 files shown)


tingyan19 (approved):

tensorflow/lite/micro/kernels/kernel_util.h
```diff
@@ -45,19 +45,33 @@ const TfLiteEvalTensor* GetEvalInput(const TfLiteContext* context,
...
 
-// Returns const data for a TfLiteEvalTensor struct.
+// Returns const data for a TfLiteEvalTensor struct that are expected to exist.
 template <typename T>
 const T* GetTensorData(const TfLiteEvalTensor* tensor) {
   TFLITE_DCHECK(tensor != nullptr);
   return reinterpret_cast<const T*>(tensor->data.raw);
 }
 
+// Returns data for a TfLiteEvalTensor struct that could be null.
```
deqiangc, line 62: Maybe add something like following: An null TfLiteEvalEtensor could happens for an optional weight/parameter tensor to an OP.

PR #1116, opened 2022-05-10
Support for Fully Connected on Vision P6

psharath1:
BUG= Added support for Fully Connected operation on vision P6


psharath1 (commented):
Have updated as per the comments

rascani (approved):

PR #1118, opened 2022-05-11
Handle the relative path in generate_cc_arrays.py

deqiangc:
This corresponds to internal cl/447815949.
This changes handles the relative input path to generate_cc_arrays.py

BUG=https://b/199442906


tensorflow/lite/micro/tools/generate_cc_arrays.py
```diff
@@ -90,7 +90,12 @@ def generate_array(input_fname):
 
 
 def get_array_name(input_fname):
-  base_array_name = 'g_' + input_fname.split('.')[0].split('/')[-1]
+  # Normalize potential relative path to remove additional dot.
+  abs_fname = os.path.abspath(input_fname)
+  print(abs_fname)
+  print(abs_fname.split('.')[0])
+  base_array_name = 'g_' + abs_fname.split('.')[-2].split('/')[-1]
+  print(base_array_name)
```
paulinesho, line 98: I think we might want to remove some of these prints?

deqiangc: absolutely!  Thanks for catching this!  This is silly of me. Note that internal cl does not have this issue. I imported a pre-clean up version by mistake.

paulinesho (approved):

PR #1134, opened 2022-05-13
Fix xtensa parallel build errors by checking directory before extracting downloaded files

chris-kuo:
Building xtensa target with -jN fails because the different make processes overwrite each other's downloaded external libraries. The issue can be fixed by checking for existence of external library directory before extracting the files. This enables parallel builds for xtensa targets.

BUG=#1135


chris-kuo:
Could you link this PR with issue #1135? Thank you!

tingyan19 (approved):

PR #1153, opened 2022-05-23
enforcing MicroInterpreter buffer api specifications in Kernal Runner

turbotoribio:
MicroInterpreter specify what kind of buffer api is available but this is only tested in integration test. Changing Kernal Runner enforce same buffer api policy at kernel test level instead of only checking for these buffer api specifications in integration test.

Corresponding Internal Changes - http://cl/453570499

bug=[b/232283593](https://b.corp.google.com/issues/232283593)


tensorflow/lite/micro/kernels/kernel_runner.cc
```diff
@@ -92,4 +98,4 @@ TfLiteStatus KernelRunner::Invoke() {
 }
 
 }  // namespace micro
-}  // namespace tflite
+}  // namespace tflite
```
tingyan19, line 101: Is this change intentional/necessary?

turbotoribio: wait I'm confused which change line 101 ?

tingyan19 (commented):
There are some test failures. I think it might be because the test does not conform to the new restriction you added? Could you take a look at the CI failures and correct them?

turbotoribio:
> There are some test failures. I think it might be because the test does not conform to the new restriction you added? Could you take a look at the CI failures and correct them?

I believe I fixed it !

turbotoribio:
Accidentally closed PR...

advaitjain (approved):

PR #1154, opened 2022-05-23
Improve documentation for testing with Docker

paulinesho:
Added example Docker commands for testing with Dockerfile.micro. We should maybe instead remove this section if we're planning to deprecate Dockerfile.micro. It could be misleading to contributors unfamiliar with `ci.yml` that this is what they need to test locally to replicate CI. 

BUG=Doc improvement


docs/continuous_integration.md
```diff
@@ -30,12 +40,12 @@ shared code. As a result, any changes to this shared code must be made in the
...
 ## Mergify
 We use [Mergify](https://mergify.com/) for our merge queue. [The documentation](https://docs.mergify.com/) is reasonably straight forward.
 ## Config File
 Our [mergify.yml](../.github/mergify.yml) is fairly standard. It triggers on `ci:ready_to_merge` label and requires all branch protection checks to pass before a PR can be added to the queue. When the PR is merged it removes the label.
 ## `ci:run` In Queue
 The one slightly complicated wrinkle in our system is the test suite being run only when the `ci:run` label is applied. As soon as the tests are run, the label is [removed](../.github/workflows/remove-labels.yml).
 
-As the queue processes each PR, it creates a temporary branch and merges in the results of the previous passing PRs my merging main. This merge resets all the test results, so the `ci:run` tag must be [reapplied](../.github/workflows/apply_cirun.yml) in queue.
\ No newline at end of file
+As the queue processes each PR, it creates a temporary branch and merges in the results of the previous passing PRs my merging main. This merge resets all the test results, so the `ci:run` tag must be [reapplied](../.github/workflows/apply_cirun.yml) in queue.
```
tingyan19, line 51: What exactly is the difference between line 41 in the original file and line 51 here? Was it caused by CR/LF differences? Can't really see here because GitHub editor does not seem to highlight anything here.

tingyan19, line 51: What exactly is the difference between line 41 in the original file and line 51 here? Was it caused by CR/LF differences? Can't really see here because GitHub editor does not seem to highlight anything here.

paulinesho: The previous version had no new line at the end of the document. This was automatically added by my editor when I saved.

PR #1156, opened 2022-05-25
size log refactor

jwithers:
This replaces our current nightly profiling log PR machinery with two scripts. log_binary_size.yml now pushes to the profiling-logs branch using the TFLM_BOT_REPO_TOKEN nightly.

log_binary_size_pr.yml pulls the logs out of the profiling-logs branch and creates a PR to merge these to main. Right now it is set to do this every Sunday at 6am PT. There is a commented schedule line to make this the 15th and 30th of every month.

For testing these both run on workflow_dispatch, so you can just run them manually to test functionality.

Note that branch protections much be set on profiling-logs branch to allow the TFLM_BOT_REPO_TOKEN to push to that branch or this will just end in tears.

BUG=https://issuetracker.google.com/issues/230830037


.github/workflows/log_binary_size.yml
```diff
@@ -32,18 +46,10 @@ jobs:
...
-         delete-branch: true
-         token: ${{ secrets.TFLM_BOT_REPO_TOKEN }}
-         title: Automated binary size log update
-         commit-message: Automated binary size log update
-         committer: TFLM-bot <tflm-github-bot@google.com>
-         author: TFLM-bot <tflm-github-bot@google.com>
-         body: "BUG=automated binary size log update"
-         labels: ci:run
-         reviewers: advaitjain
+        git commit -a -m "Automated binary size log update"
```
deqiangc, line 49: after commit, there is no push. but then there is a git checkout $LOG_BRANCH (even though I think we are are already on this branch).  Could you please help elaborate on this sequence?

deqiangc (changes requested):
If you need this in to experiment, pls let me know.

.github/workflows/log_binary_size.yml
deqiangc: Also, is this just committing the updated profile file? 

I am guessing the code in $LOG_BRANCH will stay forever stale, but what is in data/continous_builds/size_profiling will keep updated based on latest code.  Is this a necessary trade-off that we have to live by or could other directories/files in LOG_BRANCH also get updated (without impacting the continuous update of  data/continous_builds/size_profiling of couse)?

.github/workflows/log_binary_size.yml
```diff
@@ -32,18 +46,10 @@ jobs:
...
+        git checkout $LOG_BRANCH
+        git checkout main -- data/continuous_builds/size_profiling/
+        git add .
+        git status
+        git commit -a -m "Automated binary size log update"
+        git push
```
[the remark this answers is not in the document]
jwithers: No, it is pushing on $LOG_BRANCH. The format from the previous line
git checkout <branch> -- <path> 
only checks out the files from <path> into the current branch, despite looking like it is checking out another branch.

With that in mind, you can see what we are doing here is checking out the logs from $LOG_BRANCH into main, running the tests which modify those files, and then checking the modified files into $LOG_BRANCH.

.github/workflows/log_binary_size.yml
jwithers: The lack of a push is intentional, we are making no changes in main. However, to checkout the modified files from main, they must have been committed. It's just how git works. Your changes can't be accessed from another branch until the files are committed. 
I see how this is confusing. Give me a few minutes and I will do a little explanation of what is going on here with steps, cause yeah, if you don't get what is happening this looks off and isn't intuitive to everyday use at all.

jwithers:
The goal of these workflows is to run some tests against the code in main and then store the logs, which are cumulative. This is complicated by our branch protections, which do not allow us to check code into main directly, so instead we have to store our logs in $LOG_BRANCH.

The rough operation of the code goes as follows.
1. checkout main to get the latest codebase
2. pull our latest logs from $LOG_BRANCH into main
3. perform tests and emit log files
4. checkout $LOG_BRANCH
5. pull newly updated logs from main into $LOG_BRANCH 
6. push $LOG_BRANCH to repo

A couple of small items can make this code look a bit confusing unless you do fiddly things with git frequently.
* `git checkout --track <remote>/<branch>` creates a local tracking branch from a remote branch reference and changes to that branch.
* `git checkout <branch> -- <path>` checks out files from `<branch>` into the current branch. It does not switch branches.
* In order to check out files from a branch you are not currently on, those files must have first been committed to that branch. Changed files that aren't committed can't be moved between branches.

.github/workflows/log_binary_size.yml
jwithers: As for updating, we can definitely update anything we want in the $LOG_BRANCH. Only the files being affected by the test being run here will be pulled over. This is specifically as generic as possible so we can use this with other test suites that accumulate housekeeping files.

deqiangc:
> The goal of these workflows is to run some tests against the code in main and then store the logs, which are cumulative. This is complicated by our branch protections, which do not allow us to check code into main directly, so instead we have to store our logs in $LOG_BRANCH.
> 
> The rough operation of the code goes as follows.
> 
> 1. checkout main to get the latest codebase
> 2. pull our latest logs from $LOG_BRANCH into main
> 3. perform tests and emit log files
> 4. checkout $LOG_BRANCH
> 5. pull newly updated logs from main into $LOG_BRANCH
> 6. push $LOG_BRANCH to repo
> 
> A couple of small items can make this code look a bit confusing unless you do fiddly things with git frequently.
> 
> * `git checkout --track <remote>/<branch>` creates a local tracking branch from a remote branch reference and changes to that branch.
> * `git checkout <branch> -- <path>` checks out files from `<branch>` into the current branch. It does not switch branches.
> * In order to check out files from a branch you are not currently on, those files must have first been committed to that branch. Changed files that aren't committed can't be moved between branches.

Excellent explaination!  I approve this PR. 
I think let us try to have another follow PR to include your write up in the yml file after we verify the PR is verified works

deqiangc (approved):

PR #1164, opened 2022-05-27
Manual changes needed to fix the automated TF --> TFLM sync.

advaitjain:
* import changes from upstream TF:
     * https://github.com/tensorflow/tensorflow/commit/2f824e0ce28b2d533ac4edb0d01c38f8758b39a4
     * https://github.com/tensorflow/tensorflow/commit/856fd18fcf0b33908d8fec243953d3fa97bf49fa

 * make the corresponding changes in the TFLM code, following:
     * http://cl/449226814
     * http://cl/449192972

BUG=http://b/233257155 and http://b/234097792
NO_CHECK_TFLITE_FILES=manually imported files from upstream TF due to an API change.


tensorflow/lite/micro/kernels/reduce_common.cc
```diff
@@ -111,11 +117,23 @@ TfLiteStatus PrepareMeanOrSumHelper(TfLiteContext* context, TfLiteNode* node,
     op_data->output_scale = output->params.scale;
   }
 
+  context->RequestScratchBufferInArena(context, output_size * sizeof(int32_t),
+                                       &op_data->temp_index_idx);
+  context->RequestScratchBufferInArena(context, output_size * sizeof(int32_t),
+                                       &op_data->temp_buffer_idx);
```
deqiangc, line 123: Maybe I am misunderstanding something.  input can be either type int8 or int16,  but output is fixed to be int32_t?  Maybe worthy of documenting the reason of using sizeof(int32_t) here

advaitjain: Since the change to the reference implementation of reduce is going to be rolled-back in the near future, let's merge this as is.

Issue #1166, opened 2022-05-30
Hard coded registers are used for Arm targets


mansnils:
Hard coded registers are currently used for cycle count (GetCurrentTimeTicks()) and should be avoided.

PR #1167, opened 2022-05-30
Remove hard coded DWT registers for Arm targets

mansnils:
Arm Cortex-M55 uses the more realistic PMU counters.
Adds Arm Cortex-M85 for cortex_m_generic target.

BUG=https://github.com/tensorflow/tflite-micro/issues/1166

### DIFF
--- a/tensorflow/lite/micro/cortex_m_corstone_300/system_setup.cc
+++ b/tensorflow/lite/micro/cortex_m_corstone_300/system_setup.cc
@@ -1,4 +1,4 @@
-/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


mansnils:
Hi @advaitjain and @deqiangc 
I see that test_project_generation.sh is using cortex_m_generic_makefile.inc with argument project_generation. There is a comment that none of the parameters matter since we are not going to build any code: https://github.com/tensorflow/tflite-micro/blob/7d48cbd96c7fc657b5aa0b39a6e9a238b82d2ee7/tensorflow/lite/micro/tools/make/targets/cortex_m_generic_makefile.inc#L95
But later in test_project_generation.sh code is actually built with make -j8 BUILD_TYPE=cmsis_nn: https://github.com/tensorflow/tflite-micro/blob/7d48cbd96c7fc657b5aa0b39a6e9a238b82d2ee7/tensorflow/lite/micro/tools/ci_build/test_project_generation.sh#L86

This fails because now the parameters matters in this PR. In order to avoid hardcoding TARGET_ARCH need to be set (to something else than project_generation) in cortex_m_generic_makefile.inc, which will include correct header file in cortex_m_generic/micro_time.c.

So there are different ways to approach this. Perhaps the simplest way is to add something like this in the generated Makefile:
-I./third_party/cmsis/Device/ARM/ARMCM4/Include
-DCMSIS_DEVICE_ARM_CORTEX_M_XX_HEADER_FILE=\"ARMCM4.h\"

WDYT?

deqiangc (commented):
I really like the change in system_setup.cc and micro_time.cc and I think they are much more clean.  I would like a second opinion from Advait on the impact on project generation and target specific inc (e.g cortext_m_generic_makefile.inc). I slightly dislike the fact that cortext_m_generic_makefile.inc would now have more control branches (aka, more if else) for project generation, but think it is should be ok.

mansnils:
@advaitjain - ok with you related project generation?

Issue #1151, opened 2022-05-23
signed integer over/underflow and GetCurrentTimeTicks()


v0lker:
`tensorflow/lite/micro/micro_time.h:GetCurrentTimeTicks()` and friends use signed integers, with which some arithmetic operations trigger potentially undefined behaviour (for example those done in `MicroProfiler`).

counter values are also generally unsigned, so it makes more sense semantically.

it seems quite straightforward to change all of them to unsigned types. please let me know whether you agree and whether you'd like a patch.

PR #1169, opened 2022-05-30
use unsigned data types for counter values (#1151)

v0lker:
1. semantically it makes more sense

2. arithmetic over/underflow is not defined on signed ints, whereas it
   is on unsigned ones, so this removes a potential source of bugs.

BUG=#1151


deqiangc (approved):
LGTM

v0lker:
@deqiangc , thank you for the review. unfortunately, CI only ran tests afterwards, and i subsequently found out had there were some formatting issues i needed to address. could i have another review / ack, please?

tensorflow/lite/micro/arc_custom/micro_time.cc
```diff
@@ -25,18 +25,18 @@ limitations under the License.
...
   else
-    return std::numeric_limits<int32_t>::max();
+    return std::numeric_limits<uint32_t>::max();
 }
 
-int32_t GetCurrentTimeTicks() {
+uint32_t GetCurrentTimeTicks() {
   const unsigned ticks_real = _timer_default_read();
-  const int32_t ticks_cast = static_cast<int32_t>(ticks_real & 0x7fffffff);
+  const uint32_t ticks_cast = static_cast<uint32_t>(ticks_real & 0x7fffffff);
```
deqiangc, line 39: I miss this before. There is no need to do & 0x7ffffff when you use uint32_t.  That & is only needed when casting to int32

v0lker: i didn't give it the appropriate amount of consideration - i thought it'd be better to just keep the values, as it might be expected somewhere. looking at some random ARC documentation i found, it seems that there are two counters, one 32b and one 64b.

so, yes, i agree, i should've removed the mask (and will do so now).

i think it does not break anything (other than limiting the length of the time intervals that can be measured, lengths that are probably not relevant for tflite), but it seems pointless and also confusing.

v0lker: pushed the change just now. please let me know whether you would suggest any other changes..

deqiangc (approved):

PR #1187, opened 2022-06-08
Issue Creation Test

jwithers:
This is an intermediate test for Generate Issue When CI Builds Fail tasking.

Specifically, create-issue.yml:
* Triggers when the label `ci:issue` is added to a PR.
* Uses the JasonEtco/create-an-issue action to produce an issue with text from the template located in .github/ci-error-template.md. This issue is assigned to the PR author and has a link back to the PR .
* Removes the `ci:issue` tag.
 
Quick note on create-an-issue: this is mit licensed so when we start worrying about chain issues we can bring the commit we are using into our repo if we so choose.

The reason this intermediate PR needs to happen is to test some cases against forks and our actual setup. Specifically, does issue creation work at assigning to a fork PR from a user who isn't a previous committer. The github docs as I read them are ambiguous on that point and it would be pointlessly time consuming to test that in mocks as opposed to just test it here. Plus this checks against our actual tokens. 

**This code will not affect our normal CI at all.** It only activates when the `ci:issue` tag is applied. It also contains a couple of test elements for dumping the github context and pr user to console that are only for testing and will not be in the final production code.

BUG=https://issuetracker.google.com/issues/229828377


advaitjain (approved):
approving to get the testing process moving along.

PR #1192, opened 2022-06-08
Update softmax_int8_int16.cc

psharath1:
BUG= Softmax Fix. Reverted changes to the file which introduced failures

### DIFF
--- a/tensorflow/lite/micro/kernels/xtensa/softmax_int8_int16.cc
+++ b/tensorflow/lite/micro/kernels/xtensa/softmax_int8_int16.cc
@@ -108,11 +108,11 @@ void* XtensaInitSoftmax(TfLiteContext* context, const char* buffer,
 TfLiteStatus XtensaPrepareSoftmax(TfLiteContext* context, TfLiteNode* node) {
 #if defined(HIFI4) || defined(HIFI4_INTERNAL) || defined(HIFI5)
   return PrepareHifi(context, node);
-#elif defined(VISION_P6)
-  TF_LITE_ENSURE_OK(context, SoftmaxPrepareVision(context, node));
-  return kTfLiteOk;
 #else
   TF_LITE_ENSURE_OK(context, SoftmaxPrepare(context, node));
+#if defined(VISION_P6)
+  TF_LITE_ENSURE_OK(context, SoftmaxPrepareVision(context, node));
+#endif
   return kTfLiteOk;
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/kernels/xtensa/softmax_int8_int16.cc
```diff
@@ -108,11 +108,11 @@ void* XtensaInitSoftmax(TfLiteContext* context, const char* buffer,
 TfLiteStatus XtensaPrepareSoftmax(TfLiteContext* context, TfLiteNode* node) {
 #if defined(HIFI4) || defined(HIFI4_INTERNAL) || defined(HIFI5)
   return PrepareHifi(context, node);
-#elif defined(VISION_P6)
-  TF_LITE_ENSURE_OK(context, SoftmaxPrepareVision(context, node));
-  return kTfLiteOk;
 #else
```
rascani, line 111: I would suggest creating some consistency between the prepare functions. PrepareHifi calls SoftmaxPrepare, but SoftmaxPrepareVision does not. Consider having either both call SoftmaxPrepare or neither.

tensorflow/lite/micro/kernels/xtensa/softmax_int8_int16.cc
```diff
@@ -108,11 +108,11 @@ void* XtensaInitSoftmax(TfLiteContext* context, const char* buffer,
...
   TF_LITE_ENSURE_OK(context, SoftmaxPrepare(context, node));
+#if defined(VISION_P6)
+  TF_LITE_ENSURE_OK(context, SoftmaxPrepareVision(context, node));
```
rascani, line 114: This can just be `return SoftmaxPrepareVision(context, node);`

rascani (approved):

tensorflow/lite/micro/kernels/xtensa/softmax_int8_int16.cc
psharath1: For Vision P6 we have always followed this method of calling regular prepare followed by prepareVision for all ported kernels.  will need to check with HiFi team

tensorflow/lite/micro/kernels/xtensa/softmax_int8_int16.cc
psharath1: Will look at incorporating this for the next submissions

tingyan19 (approved):

PR #1194, opened 2022-06-09
Mark all subgraph inputs as used in the memory allocator

CNugteren:
This fixes a bug introduced in https://github.com/tensorflow/tflite-micro/pull/947 when there are multiple subgraphs and one of the non-main subgraphs (e.g. through a while loop) has inputs that aren't used. The memory planner did not mark these inputs as used, but the while-loop implementation does make copies from one subgraph to another. Without this fix it would cause writes to other memory addresses, resulting in incorrect results if unlucky.

This bug could be triggered when running the following Keras model (converted to TFLite) with TFLite Micro:
```python

class SimpleWhileLoop(tf.keras.layers.Layer):
    def __init__(self):
        super(SimpleWhileLoop, self).__init__()

    def call(self, inputs):

        def body(counter, data):
            return counter * 2, data * 2

        def condition(counter, data):
            return counter < 10

        i = tf.constant(1, dtype=tf.int32)
        i, result = tf.while_loop(condition, body, [i, inputs])
        return result


def get_model() -> tf.keras.Model:
    input_layer = tf.keras.layers.Input(shape=(3, 4), batch_size=1)
    x = SimpleWhileLoop()(input_layer)
    return tf.keras.Model(input_layer, x, name="while_test")
```
Before this PR, at the start of the while loop (in `micro/kernels/while.cc:85`) the initial counter value (1) would be copied to the condition subgraph, but then right after all the 12 data values would be copied as well, although they are unused. Debugging showed that the address of the loop counter was in the middle of the planned range of the data values, thus causing the counter value to be overwritten by the data. With this PR that no longer happens, although I can't be 100% sure that the bug is solved: it could be luck with the way the memory plan turned out to be. However, conceptually this fix makes sense to me so I'm sending this PR.


tensorflow/lite/micro/micro_allocation_info.cc
```diff
@@ -240,6 +240,7 @@ TfLiteStatus AllocationInfoBuilder::MarkAllocationLifetimes(
     const int tensor_index = subgraph->inputs()->Get(i);
     AllocationInfo* current = &subgraph_allocation_info[tensor_index];
     UpdateFirstCreated(current, allocation_scope_count_);
+    UpdateLastUsed(current, allocation_scope_count_);
```
deqiangc, line 243: Thanks for the PR and the detailed description, Cedric! 
Looking at the example in the description, that "unused" tensor for the WHILE OP would be marked with last used after the two subgraphs have been traversed (Specifically at line 276).  The proposed change here does not hurt but should not help as well because line 276 should kick in and override the last_used. 

Note that there is a TestMultiSubgraphWithIf in micro_allocation_info_test.cc that test some similar scenarios that you describe. Maybe can craft a similar unit test case?

The memory planner code is tricky and I could be missing something here; so feel free to let us know what you think.

CNugteren: Thank you for the review and apologies for the late reply. I've now slightly modified the existing test model (I thought copy-pasting 99% of the test would not be good practice) to no longer use 2 inputs for the `MUL` operator in the else-branch:
image


Then, I undid my change again and it resulted in this assert:
```
allocation_info[8].last_used == 4 failed at tensorflow/lite/micro/micro_allocation_info_test.cc:118 (-1 vs 4)
```
In other words, without my modification, this input to the then-graph is considered unused. However (at least in the `while.cc` implementation) it is used, even if none of the ops in the graph use it (like in the Python example above).

Note that the reason the values are now 4 instead of 5 is that they are no longer used in the `MUL` operation.

So I am now more confident that my fix is indeed a proper fix. Let me know what you think.

Tombana:
I think the same fix has been applied a month ago here: https://github.com/tensorflow/tflite-micro/commit/7925e839c75a1d0b165150b128ba8ec88dcec031

PR #1195, opened 2022-06-09
Add support for offline planning in case of multiple subgraphs

CNugteren:
In https://github.com/tensorflow/tflite-micro/pull/947 support for multiple subgraphs was added to the memory planner. However, in case memory planning information was provided offline this would not work since it would only read the planning data for the first subgraph. This PR fixes that.

The fix itself is simple and local to the function, assuming all offline planned tensors are just concatenated, e.g.: `[s0t0, s0t1, s1t0, s2t1, s1t2, s3t0, s3t1, s3t2, s3t3]`. The documentation is updated to reflect this.

The metadata format also had a subgraph ID according to the documentation, but that was never read. I tried to use that but that wouldn't work either, because it would only hold one subgraph ID and we can only have one meta-data with a given name. So I'm not sure how this was intended to be used originally but I added a note to the documentation and the code to clarify that it is unused at the moment.

BUG=see PR description


advaitjain:
Thanks for the detailed info and apologies for the long delay. I will take a look in the next two weeks.

tensorflow/lite/micro/docs/memory_management.md
```diff
@@ -94,6 +94,11 @@ integers of the following format:
 | ... | ... |
```
mansnils, line 94: I think before there were support for multiple subgraphs the intention was something like this.

| Offset | Value |
|-|-|
| 0 | Offline allocation format version |
| 1 | Subgraph index to which this allocation applies |   <--- SUBGRAPH INDEX 0
| 2 | Number offsets following: n |                       <--- NUMBER OF OFFSETS FOR SUBGRAPH 0
| 3 | Byte offset of tensor #0 or -1 to allocate at runtime |
| 4 | Byte offset of tensor #1 or -1 to allocate at runtime |
| ... | ... |
| 3+(n-1) | Byte offset of tensor #(n-1) or -1 to allocate at runtime |

| Offset | Value |
|-|-|
| 0 | Offline allocation format version |
| 1 | Subgraph index to which this allocation applies |    <--- SUBGRAPH INDEX 1
| 2 | Number offsets following: n |                        <--- NUMBER OF OFFSETS FOR SUBGRAPH 1
| 3 | Byte offset of tensor #0 or -1 to allocate at runtime |
| 4 | Byte offset of tensor #1 or -1 to allocate at runtime |
| ... | ... |
| 3+(n-1) | Byte offset of tensor #(n-1) or -1 to allocate at runtime |

Etc adding more of these...

However now it seems this is the current behavior and I think the documentation should be updated to something like this:

| Offset | Value |
|-|-|
| 0 | Offline allocation format version |
| 1 | Number of subgraphs: n |
| 2 | Number tensors |
| 3 | Byte offset for subgraph 0 of tensor #0 or -1 to allocate at runtime |
| 4 | Byte offset for subgraph 0 of tensor #1 or -1 to allocate at runtime |
| ... | ... |
| 3 + (num_tensors_for_subgraph_0 - 1) | Byte offset for subgraph 0 of tensor #(num_tensors_for_subgraph_0 - 1) or -1 to allocate at runtime |
| ... | Byte offset for subgraph 1 of tensor #0 or -1 to allocate at runtime |
| ... | Byte offset for subgraph 1 of tensor #1 or -1 to allocate at runtime |
| ... | ... |
| ... | Byte offset for subgraph 1 of tensor #(num_tensors_for_subgraph_1 - 1) or -1 to allocate at run
| ... | ... |
| ... | Byte offset for subgraph (n - 1) of tensor #0 or -1 to allocate at runtime |
| ... | Byte offset for subgraph (n - 1) of tensor #1 or -1 to allocate at runtime |
| ... | ... |
| ... | Byte offset for subgraph (n - 1) of tensor #(num_tensors_for_subgraph_x - 1) or -1 to allocate at run

mansnils (changes requested):
Thanks for this. The fix looks good to me. Although the documentation seems to be not in sync with the current behavior. See other comment.
A unit test would be great as well.

tensorflow/lite/micro/docs/memory_management.md
CNugteren: OK thank you for the explanation. I've updated the documentation slightly. I didn't put the full table in as you put it here, I think that will only be confusing, especially for the most common case with only a single subgraph. Let me know if the current documentation is still missing something.

mansnils (dismissed):

advaitjain (approved):

Issue #1190, opened 2022-06-08
Add support for the DIV op


sasapetr:
This issue is to track efforts to port the reference DIV kernel from TF Lite to TFLM.  Some work has already been done in https://github.com/tensorflow/tensorflow/issues/45431, but it seems that it was not fully ported.

 https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/porting_reference_ops.md explains the steps to follow.

JustinOng:
It seems like the DIV kernel was fully merged into TFLM, but it was removed in tensorflow/tensorflow#48198. I can't find any other reference to `184059002` to explain why it was removed either.

PR #1196, opened 2022-06-10
This is a copy without modification of the reference kernel for DIV from lite

sasapetr:
This is a copy of the DIV kernel from lite to micro.

It corresponds to step 5.1 in the porting reference ops documentation.

This is part of issue #1190 (internal bug b/235190494).

### DIFF
--- a/tensorflow/lite/micro/kernels/div.h
+++ b/tensorflow/lite/micro/kernels/div.h
@@ -1,8 +1,11 @@
 /* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
+
     http://www.apache.org/licenses/LICENSE-2.0
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/tensorflow/lite/micro/kernels/div.h
+++ b/tensorflow/lite/micro/kernels/div.h
@@ -1,0 +1,244 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


tingyan19 (changes requested):
Please take a look at https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/porting_reference_ops.md#5-port-the-op-from-lite-to-micro-pr3 - this "copy without modification" part is supposed to be the first commit of a much larger PR. Please complete all 7 steps in this section and then request for review.

sasapetr:
Thanks Ting, I'll do that.  Just to note: while the documentation mentions doing everything as a single PR, the examples in the documentation have done this in multiple PR.  In particular, step 5 is broken into the following PRs:

https://github.com/tensorflow/tensorflow/pull/45457 (this is equivalent to this PR afaics, step 5.1)
https://github.com/tensorflow/tensorflow/pull/45646 (this is removing the lite-specific code, step 5.2)
https://github.com/tensorflow/tensorflow/pull/45647 (this is the porting part, step 5.3 only)

It might be good to fix the documentation and make it clear that the examples therein do not reflect the way we should structure the PRs.

tingyan19 (dismissed):

tensorflow/lite/micro/kernels/div_test.cc
```diff
@@ -0,0 +1,377 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
tingyan19, line 1: Are the following tests from TFLite ported over? 

TEST(IntegerDivOpTest, NoActivation)
TEST(IntegerDivOpTest, ActivationRELU_N1_TO_1)
TEST(IntegerDivOpTest, VariousInputShapes)
TEST(IntegerDivOpTest, WithBroadcast)

sasapetr: These are not ported as TFLM only supports float and int8 types (step 5.2 in the documentation says "Remove input and output types other than int8 and float32."), and these tests are testing int32.  Or maybe I'm missing something?

tensorflow/lite/micro/kernels/div.h
```diff
@@ -0,0 +1,204 @@
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
```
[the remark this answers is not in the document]
sasapetr: Div.h has been synced from upstream in https://github.com/tensorflow/tflite-micro/pull/1200, I've switched to using it in commit c600091bd85102aa569e4bce5a5e6ee0970fbc7f.

Issue #1199, opened 2022-06-13
GreedyMemoryPlanner tests hangs for target cortex_m_corstone_300


annietllnd:
Some GreedyMemoryPlanner test cases hang when run on the Corstone-300 on Cortex-M55, when optimization level != O2 (optimizing for size). The hang seems be a HardFault caused by an unaligned memory access issue, causing a hang in the [reverse sort algorithm](https://github.com/tensorflow/tflite-micro/blob/f868a900f1f788fd0a55b5655b47fc352f9a0239/tensorflow/lite/micro/memory_planner/greedy_memory_planner.cc#L43). It is unclear why this happens, as I have not been able to find any signs of corrupted memory.  The affected tests are:

TestGreedyBasics
TestGreedyMedium
TestPersonDetectionModel
TestOverlapCase

Recreate by following the same methodology as in #847, but for the greedy memory planner test instead 

Apply this patch which removes the fix for this issue:
https://github.com/jenselofsson/tflite-micro/commit/f70fbce65f1bdb230f62389fd1e7c305ddfbd8de

$ make -f tensorflow/lite/micro/tools/make/Makefile clean
$ make -f tensorflow/lite/micro/tools/make/Makefile TARGET=cortex_m_corstone_300 TARGET_ARCH=cortex-m55 test_greedy_memory_planner_test

annietllnd:
Hi, this issue falls under #274

annietllnd:
#1207 did not fix this, sorry for the confusion! However #1228 will, so when it's merged we can close this, too :relaxed:

Issue #847, opened 2022-01-10
DetectionPostprocessFloatRegularNMS hangs for target cortex_m_corstone_300 on Cortex-M55


jenselofsson:
EDITED 2022-01-27 for accuracy

The hang occurs for the testcase [DetectionPostprocessFloatRegularNMS](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/kernels/detection_postprocess.cc#L118) in the [Init-function](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/kernels/detection_postprocess.cc#L118) when trying to access element at "nms_score_threshold" in the flexbuffers map m.

It also seems to happen when trying to access the following elements in the same map:
nms_iou_threshold
y_scale
x_scale
h_scale
w_scale

The hang occurs in the [ReadScalar-function](https://github.com/google/flatbuffers/blob/master/include/flatbuffers/base.h#L406) in the flatbuffer library, when p is dereferenced. The hang occurs due to a HardFault due to unaligned access. This makes me think that the issue-report might belong more in the flatbuffers-issue tracker.

Steps to reproduce:
Apply this patch which removes the fix for this issue:
https://github.com/jenselofsson/tflite-micro/commit/f70fbce65f1bdb230f62389fd1e7c305ddfbd8de

$ make -f tensorflow/lite/micro/tools/make/Makefile TARGET=cortex_m_corstone_300 TARGET_ARCH=cortex-m55 test_kernel_detection_postprocess_test

jenselofsson:
I seemed to have spoken too soon, I had some local changes laying about which gave a different result. The new results are that GCC works fine, but hangs for armclang for at least -O0, -O1, -O2, -O3.

annietllnd:
[...]

PR #1207, opened 2022-06-15
fix alignment bug in DetectionPostProcess for target based on Arm Corstone-300 software

annietllnd:
Hi, this PR resolves the hang in the PPD kernel, caused by a HardFault: unaligned access. See explanation in issue referenced below. Also note that a similar issue (#1199), for which we are yet to find the root cause, has been opened and is referenced in the .inc file for the target. With this patch also comes some updates to the documentation and generation of the test data.
/Annie
BUG=#847


tensorflow/lite/micro/tools/make/targets/cortex_m_corstone_300_makefile.inc
```diff
@@ -16,7 +16,7 @@
 # ARM Cortex M makefile targeted for a FVP based on Arm Corstone-300 software.
 # For more info see: tensorflow/lite/micro/cortex_m_corstone_300/README.md
 
-# TODO(#193): Remove this.
+# TODO(#1199): Remove this
 CORE_OPTIMIZATION_LEVEL := -O2
```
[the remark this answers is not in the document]
annietllnd: Hi, this comment has been addressed. We have disabled some other problematic tests (to be resolved by #274), and added a fix for the network tester test which had a similar behavior.

mansnils (approved):

tensorflow/lite/micro/tools/make/targets/cortex_m_corstone_300_makefile.inc
```diff
@@ -197,9 +194,13 @@ INCLUDES += \
   -I$(CMSIS_PATH)/Device/ARM/$(ARM_CPU)/Include \
   -I$(CMSIS_PATH)/CMSIS/Core/Include
 
-# TODO(#47070): Examine why some tests fail here.
+# TODO(#274): Examine why some tests fail here.
 EXCLUDED_TESTS := \
```
deqiangc, line 198: I am a bit concerned of removing all these tests.  How much benefit do it gain by removing -O2?  I am a bit leaning on being safe since greedy_memory_planner are one important piece of the code and if it does not pass the test and the failure are genuine, debugging it on target would be a tremendous pain.  I totally agree on the need chasing down why this fails though and have create internal bugs to highlight this as well.

mansnils: One benefit of removing -O2 is that if not doing it more of these errors may accumulate in other unit tests, so having -O2 may hide these errors.

deqiangc: ok. It is fair to expose issues rather than leaving it festing.

deqiangc (approved):

PR #1214, opened 2022-06-17
Codeowners branch

jwithers:
Adding self to all codeowners categories for testing per conversation with @advaitjain on 17 june 22

BUG=https://issuetracker.google.com/issues/236272037


advaitjain:
Closing because we don't need to go down this debugging path right now: https://github.com/tensorflow/tflite-micro/pull/1072/#issuecomment-1159148640

PR #1220, opened 2022-06-21
log png generation fix

jwithers:
The problem here was that the logs were being pruned to 60 entries for image generation from the top of the file, when later entries were being appended to the end. So after 60 entries the png images stopped changing. Have used pandas slicing and reindexing to fix the problem. Also on advice of advait have added dates to the image titles so we can see the date ranges being covered.

If these files end up never being pruned and get huge a few decades from now, likely want to do this reduction using tail and temporary files on the filesystem which would be more memory efficient. We aren't likely to have to worry about that in our working lifetimes. 

BUG=https://issuetracker.google.com/issues/235876344

### DIFF
--- a/.github/ci-error-template.md
+++ b/.github/ci-error-template.md
@@ -1,0 +1,7 @@
+---
+title: Failed CI Test
+labels: bug
+---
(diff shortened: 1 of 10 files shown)


deqiangc (approved):

deqiangc (changes requested):
Please fix the format check failure. See presubmit test.

deqiangc (approved):

PR #1227, opened 2022-06-27
Relax 0 size constraint on intermediate tensors to a warning

deqiangc:
Once customer reports a non-zero size intermediate tensors.
The model is not public available; so this PR is for them
to move ahead and find out more.

BUG=request from customer


advaitjain:
Closing since we don't seem to require this change.

PR #1228, opened 2022-06-28
Fix alignment bugs in greedy memory planner and micro tests

mansnils:
This is needed for target based on Arm Corstone-300 software.

BUG=https://github.com/tensorflow/tflite-micro/issues/274


tensorflow/lite/micro/micro_context_test.cc
```diff
@@ -48,7 +49,7 @@ tflite::MicroContext CreateMicroContext() {
 // Test structure for external context payload.
 struct TestExternalContextPayloadData {
   // Opaque blob
-  uint8_t blob_data[128];
+  alignas(4) uint8_t blob_data[128];
```
deqiangc, line 52: Could you please elaborate why these align directive are needed? For example, the blob_data seems to rely on byte access only; so I am not sure we should align blindly. Same comment goes with micro_graph_placement_buffer.  It would be great to understand what access cause the requirement of align directive.  For example, I am OK with some flatbuffer align to certain boundary because we understand that flatbuffer code use word access when accessing certain uint32 data member (because flatbuffer align the data to native boundary, which is quite reasonable).

rockyrhodes (approved):

PR #1229, opened 2022-06-28
Port the SUM operator from lite

### LINKED ISSUE #1219: Add support for the SUM op
This issue tracks the effort to move the SUM op (which is a shortcut for reduce sum operation) from lite to TFLM.  Some reduce ops already exist (mean and max), so we will add SUM to the same files.

This corresponds to internal bug b/236681001.

sasapetr:
- Add micro implementation for the reduce sum op.
- Fix the error message for supported types in reduce ops
- Refactor reduce_test: replace TestMeanFloatInput4D with the more general version.
- Refactor reduce_test: replace TestMeanOpQuantized with the more general version.
- Add eval helper function for the reduce sum op.
- Use ReduceGeneric for reduce sum implementation.
- Add tests for reduce sum.

Fixes #1219 

BUG=https://github.com/tensorflow/tflite-micro/issues/1219
Internal bug: b/236681001


tensorflow/lite/micro/kernels/reduce_common.cc
```diff
@@ -308,4 +308,68 @@ TfLiteStatus EvalMaxHelper(TfLiteContext* context, TfLiteNode* node,
...
+              tflite::micro::GetTensorData<int16_t>(input), op_data->input_zp,
+              op_data->input_scale, input->dims->data, input->dims->size,
+              tflite::micro::GetTensorData<int16_t>(output), op_data->output_zp,
+              op_data->output_scale, output->dims->data, output->dims->size,
+              tflite::micro::GetTensorData<int>(axis), num_axis,
+              params->keep_dims, temp_index, resolved_axis, temp_buffer,
+              /*compute_sum=*/true));
+    } break;
+    default:
+      TF_LITE_ENSURE_MSG(context, false,
```
paulinesho, line 368: Could you change this to using `MicroPrintf` and returning an error instead? See the default case in `EvalMaxHelper`.

tensorflow/lite/micro/kernels/reduce_test.cc
```diff
@@ -91,34 +95,6 @@ TfLiteStatus ValidateReduceGoldens(TfLiteTensor* tensors, int tensors_size,
   return kTfLiteOk;
 }
 
-void TestMeanFloatInput4D(int* input_dims_data, const float* input_data,
```
paulinesho, line 94: Really like the refactorization done here. Thanks!

tensorflow/lite/micro/kernels/reduce_test.cc
```diff
@@ -606,12 +549,351 @@ TF_LITE_MICRO_TEST(MeanInt84DWithoutKeepDimsWithPrecision) {
...
+  tflite::testing::TestReduceOpQuantized<int8_t>(
       kInputShape4D, kInputData4D, input_data_quant, input_scale,
       input_zero_point, tflite::testing::kAxisShape4D,
       tflite::testing::kAxisData4D, kOutputShape4D, kGoldenData4D,
       output_data_quant, expected_output_data_quant, output_scale,
-      output_zero_point, &params);
+      output_zero_point, tflite::Register_MEAN(), &params, 1.0);
+}
+
+TF_LITE_MICRO_TEST(SumFloatFlatten2ReduceDims) {
```
paulinesho, line 560: At first glance these tests don't seem to be directly ported from TFLite. Please let me know if I'm wrong and need to take a closer look. Otherwise, could you elaborate why TFLite tests were dropped and new ones were added? Please also elaborate how these expected outputs were obtained.

sasapetr: This is actually due to github repo not being synced with google3, see b/236636177 for details.  I added the ported tests here in expectation of this being fixed before this PR got merged, but there's been no progress there.  Let me know how you'd like to proceed.

tensorflow/lite/micro/kernels/reduce_common.cc
sasapetr: Done in commit db07f0ff96e5476de178bec59e463ac1b437ec6c

tensorflow/lite/micro/kernels/reduce_test.cc
paulinesho: The linked bug refers the 3 corresponding tests for Mean, not Sum. Is this the first time these 3 SumFlatten tests are added? If so, did you generate the golden data?

I'm still confused what the test porting strategy here is. For example, I see that we have `SumFloat(2/3/4)DKeepDims` here, but there are 12 `ConstFloatSumOpTest` tests in TFLite. Are those relevant? The other additions for Int8 and Int16 are fine.

sasapetr: I'm not trying to port the tests from tensorflow/lite/kernels/reduce_test.cc, but rather add the equivalents of the mean op tests found in tensorflow/lite/micro/kernels/reduce_test.cc (this file).  I'm doing this in order to be consistent with the tests in this file, and thus my new tests should correspond 1:1 to the tests for the mean op in this file.  For example, mean has MeanFloat2DKeepDims so I'm adding SumFloat2DKeepDims, mean has MeanInt82DKeepDims so I'm adding SumInt82DKeepDims, etc and I'm reusing the same inputs as the mean op tests.  If you look at the mean op tests, they also don't correspond 1:1 to the tests from Lite.

Trying to keep the mean and sum op tests 1:1 *within this file* is also why I'm adding the three extra tests (SumFloatFlatten2ReduceDims, SumFloatFlatten2NonReduceDims, and SumFloatFlatten2MiddleDims) -- I expected that the Mean* version of these tests would be synced to the github repo so after that sync I'd be consistent.  This however never happened, and now I see that Advait actually rolled back these tests in cr/456546439 so perhaps it makes sense that I remove those tests as well?  Let me know what you prefer.

Also, if you prefer that I port the tests from lite directly rather than stay consistent with the tests within this file I'm fine with that, just let me know.  It's unclear to me what the guidelines here are.

paulinesho: That makes sense. Let's keep all the tests that you've already created since you've already gone through the trouble of adding them. In addition, can we do another pass of looking at TFLite's `ConstFloatSumOpTest` and see what can be ported? That seems like the biggest gap to me right now. Irrelevant tests can be omitted. We should be good to go after this.

sasapetr: Ported three more tests from lite.  The rest either don't make sense here or are already covered by existing tests, PTAL.

paulinesho: Thanks for doing another pass.

paulinesho (approved):

paulinesho:
@sasapetr This PR didn't pass the presubmits because the op doesn't exist in the xtensa optimized reduce kernel. That's okay, but we should comment out the new tests in `reduce_test.cc` with `#if !defined(XTENSA)`. Another issue is that the PR description is missing the `BUG=` keyword. Please add that and point it to the GitHub issue and buganizer ticket. Thanks!

sasapetr:
Added the Xtensa workaround and added the bug information.  Note that the PR was already linked to the Github issue as I did have the 'Fixes #1219' in the description -- can you clarify what BUG= does in addition to this?  Btw, might be worth calling these things out in the documentation, just to avoid delays and make the whole process smoother.

paulinesho (approved):

Issue #1148, opened 2022-05-20
LSTM Sigmoid and Tanh to LUT based implementation instead of the gemmlowp implementation


lukmanr-cadence:
Hi @tingyan19 , @advaitjain , 

TFLM has sigmoid and tanh ops which call LUT based reference_integer_ops::Logistic and reference_integer_ops::Tanh.  This implementation is better than gemmlowp implementations in terms of both precision and cycle performance.  
But LSTM function calls micro_tensor_utils::ApplySigmoid and micro_tensor_utils::ApplyTanh, which calls gemmlowp::logistic and gemmlowp::tanh. 
In both reference kernel and LSTM implementations, calling the same implementation for sigmoid and tanh is preferred. 
LUT based implementation (reference_integer_ops::Logistic and reference_integer_ops::Tanh)  is preferred since these implementations are better in terms of precision and cycle performance.  

Thanks,
Lukman

PR #1253, opened 2022-07-14
Cadence:LUT based impln for sigmoid and tanh

lukmanr-cadence:
LUT based reference_integer_ops implementation for sigmoid and tanh are
called instead of gemmlowp implementation.
BUG= #1148


tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test_config.cc
```diff
@@ -231,7 +231,7 @@ LstmIntegerTestConfig lstm_integer_no_peephole_config = {
                 kLstmSeqLengthIntegerNoPeephole *
                 kLstmNumOutputIntegerNoPeephole]){
             127, 127, -108, -67, 127, 127, -128, 127, 127, -128, 127, 127, 127,
-            127, 127, -128, 127, 127},
+            127, 13, -128, 127, 127},
```
advaitjain, line 234: this is quite a large difference in the overall output compared to the previous implementation. Can you confirm that this is expected and not a bug?

tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test_config.cc
```diff
@@ -471,9 +471,9 @@ LstmIntegerTestConfig lstm_integer_peephole_config = {
     .output = lstm_output_peephole,
     .expected_output =
         (int8_t[kLstmNumBatchIntegerPeephole * kLstmSeqLengthIntegerPeephole *
-                kLstmNumOutputIntegerPeephole]){127, 127, -16, -21, 127, 127,
-                                                23, 127, 127, -128, 127, 127,
-                                                127, 127, 127, -128, 127, 127},
+                kLstmNumOutputIntegerPeephole]){127,127, -16, -21, 127, 127,
+                                                -128, 127, 127, -128, 127, 127,
+                                                127, 127, 105, -128, 127, 127},
```
advaitjain, line 476: this is quite a large difference in the overall output compared to the previous implementation. Can you confirm that this is expected and not a bug?

advaitjain (changes requested):

tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test_config.cc
lukmanr-cadence: Yes. This difference is expected. Activation functions (sigmoid and tanh) are called in 5 places in LSTM. 
Input to Activations functions are constant precision(Q12 format) for first 4 calls. So, here for all potential inputs, LUT based activation functions will produce the outputs with greater accuracy. But last activation function (tanh) is called for the calculation for the cell state vector tensor. Here input precision for the  tanh function will be depend on the scale factor of cell state vector tensor. 

For the unit test case, input to this tanh function is Q15 format. Inputs needs to be down shifted to Q12  before calling LUT based tanh function since LUT based activation runs only for the Q12 input. But original gemmlowp implementation is called  directly on Q15 inputs. Although the inputs are downshifted, the LUT based activations produces higher accuracy in majority of outputs. Maximum absolute difference for cell state for the original implementation (gemmlowp)and LUT based implementation is just 6 for the first iteration. But this difference is continues to grow over the following iterations and the final difference is quite large for this unit test.
 
Below are input and outputs for the last activation function (tanh for the cell state vector calculation)  for various implementations.
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/lukmanr/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/lukmanr/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">



tanh   Input | Reference   calculated from actual float tanh implementation | gemmlowp | LUT based   reference_integer_ops | gemmlowp (abs   diff w.r.t ref) | LUT based   reference_integer_ops (abs diff w.r.t ref)
-- | -- | -- | -- | -- | --
-14654 | -13749 | -13744 | -13750 | 5 | 1
8937 | 8721 | 8720 | 8721 | 1 | 0
5772 | 5713 | 5708 | 5709 | 5 | 4
6717 | 6624 | 6620 | 6620 | 4 | 4
-11848 | -11357 | -11356 | -11357 | 1 | 0
7851 | 7704 | 7700 | 7701 | 4 | 3
1512 | 1510 | 1512 | 1511 | 2 | 1
12938 | 12305 | 12304 | 12304 | 1 | 1



</body>

</html>

advaitjain:
Thanks for the detailed explanation. I'm comfortable with merging this change and then taking more time to investigate what it means for differences between the TfLite and TFLM reference implementations that this PR is introducing.

advaitjain (approved):

lukmanr-cadence:
Thank you, Advait and Team !

PR #1254, opened 2022-07-15
Fix Python interpreter destructor bug and disable invoke exception

paulinesho:
1) BUG=http://b/239225101

In the current Python interpreter, we do Py_DECREF(model_) in the body destructor of
InterpreterWrapper, which will be called first before member variables' destructors. This causes
an issue because destructing the MicroInterpreter (namely FreeSubgraphs) relies on some parts of
model_. This means that when the destructor of MicroInterpreter is called, the model may have
already been released, causing a segfault.

We can bypass this by not using a unique pointer for MicroInterpreter, and simply delete
interpreter_ before we call Py_DECREF(model_).

2) BUG=http://b/238459185

Some error codes are expected, such as cycles_until_run in the circular buffer op. The interpreter
wrapper shouldn't throw any exception in Python when invoke returns an error.


advaitjain (approved):
The PR descriptions seems to be a concatenation of two changes. Other than that, LGTM

advaitjain (changes requested):
undoing my LGTM to prevent our CI machinery from merging with the current PR description.

paulinesho:
> undoing my LGTM to prevent our CI machinery from merging with the current PR description.

Thanks for the quick attention. I actually did sneak in the second small change into this PR. :|

advaitjain (approved):

PR #1268, opened 2022-07-21
Fix linter issues in a few files.

sasapetr:
BUG=cleanup


advaitjain (approved):
I pushed a change to revert one file whose source of truth is upstream TF and we can now merge this PR.

PR #1275, opened 2022-07-22
enable reduce_sum kernel tests for Xtensa.

advaitjain:
PR #1229 ported over reduce sum but did not add a reference fallback to `kernels/xtensa/reduce.cc`. As a result #1229 put the kernel tests behind a `#if !defined(XTENSA)` guard.

Instead of excluding targets in the kernel tests, what we would like to do is add reference code fallbacks to the optimized kernel sources so that we can guarantee that the optimized kernels have the same functionality as the reference kernels (even if not all the code paths are truly optimized).

With PR #1274 (specifically [this commit](https://github.com/tensorflow/tflite-micro/pull/1274/commits/d36c9dd598dcbf352f2c60463fd0d4153703a1cd)), we added reference fallbacks to `xtensa/reduce.cc` but missed the fact that the kernel test cases had been ifdef'd out.

The current PR enables all the kernels test cases for all targets.

BUG=http://b/236681001

### DIFF
--- a/tensorflow/lite/micro/kernels/reduce_test.cc
+++ b/tensorflow/lite/micro/kernels/reduce_test.cc
@@ -37,11 +37,8 @@ static const int32_t kAxisData2D[] = {1};
 static const int kOutputElements2D = 2;
 static int kOutputShape2D[] = {2, 1, 2};
 static const float kGoldenData2D[] = {2.5, 6.5};
-#if !defined(XTENSA)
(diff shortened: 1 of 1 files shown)


advaitjain:
Tagging @sasapetr

paulinesho (approved):
Thanks for fixing this @advaitjain! I'll add this to the FAQ for our porting guide as well.

PR #1276, opened 2022-07-22
Fix the xtensa hifi5 quantize tests.

tanmaydas82:
BUG=239981923


advaitjain (changes requested):
looks great!

let's rearrange the cases within the switch block to have a consistent ordering of uint8, int8, int16, int32.

I think that will make it easier to read.

advaitjain (dismissed):

advaitjain (approved):

PR #1290, opened 2022-07-30
Microprofiler can now output logs for benchmarks in the csv format.

tanmaydas82:
BUG=240626703


tensorflow/lite/micro/micro_profiler.cc
```diff
@@ -67,4 +67,58 @@ void MicroProfiler::LogCsv() const {
...
+      break;
+    }
+    MicroPrintf("%s,%d", ticksPerTag.tag, ticksPerTag.ticks);
+  }
+  MicroPrintf("total,%d", totalTicks);
+
+#endif
+}
+
+int MicroProfiler::FindExistingOrNextPosition(const char* tagName) {
```
paulinesho, line 96: Could we have some comments on what this function does?

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -61,6 +61,10 @@ class MicroProfiler {
   // Separated Value) form.
   void LogCsv() const;
 
+  // Prints the profiling information of each of the tags in CSV (Comma
+  // Separated Value) form.
```
advaitjain, line 65: We should expand this description.

Something like:
```
// Prints  total ticks for each unique tag in CSV format.
//
// Output will have one row for each unique tag along with the total ticks summed across all events with that particular tag.
```

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -72,6 +76,14 @@ class MicroProfiler {
   uint32_t end_ticks_[kMaxEvents];
   int num_events_ = 0;
 
+  struct ticks_per_tag {
```
advaitjain, line 79: rename: TicksPerTag

https://google.github.io/styleguide/cppguide.html#Type_Names

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -72,6 +76,14 @@ class MicroProfiler {
...
+    const char* tag;
+    uint32_t ticks;
+  };
+  ticks_per_tag total_ticks_per_tag[kMaxEvents];
```
advaitjain, line 83: add a comment that in practice the number of unique tags will be less than the number of events but it is theoretically possible for each event to be unique and hence we allow for the TicksPerTag  array to have kMaxEvents entrires.

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -72,6 +76,14 @@ class MicroProfiler {
...
+
+  int FindExistingOrNextPosition(const char* tagName);
```
advaitjain, line 85: nit: add a comment about what this function does.

advaitjain, line 85: tagName -> tag_name

tensorflow/lite/micro/micro_profiler.cc
```diff
@@ -67,4 +67,58 @@ void MicroProfiler::LogCsv() const {
 #endif
 }
 
+void MicroProfiler::LogTicksPerTagCsv() {
+#if !defined(TF_LITE_STRIP_ERROR_STRINGS)
+  MicroPrintf(
+      "\"Unique Ops in the Graph\",\"Total Ticks (all instances of the Op)\"");
```
advaitjain, line 73: nit: technically may not be Ops. It could be any event that we profile with the MicroProfiler.

Unique Tag, Total ticks across all events with that tag.

tensorflow/lite/micro/micro_profiler.cc
```diff
@@ -67,4 +67,58 @@ void MicroProfiler::LogCsv() const {
...
+  int totalTicks = 0;
```
advaitjain, line 74: total_ticks

tensorflow/lite/micro/micro_profiler.cc
```diff
@@ -67,4 +67,58 @@ void MicroProfiler::LogCsv() const {
...
+  }
+
+  for (int i = 0; i < num_events_; ++i) {
+    ticks_per_tag ticksPerTag = total_ticks_per_tag[i];
+    if (ticksPerTag.tag == nullptr) {
+      break;
+    }
+    MicroPrintf("%s,%d", ticksPerTag.tag, ticksPerTag.ticks);
+  }
+  MicroPrintf("total,%d", totalTicks);
```
advaitjain, line 91: this seems to not strictly be a part of the csv format. I would say skip from the current function.

advaitjain (changes requested):
high level comment is that it might be worth implementing a generic Map class that is embedded friendly and then use that to perform the computation that we are doing to get to the total ticks per tag.

tensorflow/lite/micro/micro_profiler.h
tanmaydas82: Thank you for the link! Really helpful.

tensorflow/lite/micro/micro_profiler.cc
tanmaydas82: Thank you for the suggestion Pauline!  Adding a descriptive comment.

tensorflow/lite/micro/micro_profiler.h
tanmaydas82: Done. Thanks for the suggestion!

tensorflow/lite/micro/micro_profiler.h
tanmaydas82: Done. Thank you for the suggestion!

tensorflow/lite/micro/micro_profiler.h
tanmaydas82: Done. Thank you for the suggestion!

tensorflow/lite/micro/micro_profiler.h
tanmaydas82: Done. Thank you for the suggestion!

tensorflow/lite/micro/micro_profiler.cc
tanmaydas82: Got it. I was thinking to print the total count as well as I have seen that this is required for the calculation but if you think that is not required, I can remove it. I can probably change it to "total number of ticks" or something like that to make it more meaningful.

tensorflow/lite/micro/micro_profiler.cc
tanmaydas82: Done. Thank you for the suggestion!

tensorflow/lite/micro/micro_profiler.cc
```diff
@@ -67,4 +67,65 @@ void MicroProfiler::LogCsv() const {
...
+    if (each_tag_entry.tag == nullptr) {
+      break;
+    }
+    MicroPrintf("%s,%d", each_tag_entry.tag, each_tag_entry.ticks);
+  }
+  MicroPrintf("total,%d", total_ticks);
+#endif
+}
+
+/**
```
advaitjain, line 95: super nit: let's use // for the comment style for consistency. 

https://google.github.io/styleguide/cppguide.html#Comment_Style

tensorflow/lite/micro/micro_profiler.cc
```diff
@@ -67,4 +67,65 @@ void MicroProfiler::LogCsv() const {
...
+ */
+int MicroProfiler::FindExistingOrNextPosition(const char* tag_name) {
+  int pos = 0;
+  for (; pos < num_events_; pos++) {
+    TicksPerTag each_tag_entry = total_ticks_per_tag[pos];
+    if (each_tag_entry.tag == nullptr) {
+      return pos;
+    } else {
+      const char* currentTagName_t = each_tag_entry.tag;
+      const char* newTagName_t = tag_name;
```
advaitjain, line 111: current_tag_name
new_tag_name

tensorflow/lite/micro/micro_profiler.cc
```diff
@@ -67,4 +67,65 @@ void MicroProfiler::LogCsv() const {
...
+      bool matched = true;
+      while ((*currentTagName_t != '\0') && (*newTagName_t != '\0')) {
+        if (((*currentTagName_t) == (*newTagName_t))) {
+          ++currentTagName_t;
+          ++newTagName_t;
+        } else {
+          matched = false;
+          break;
+        }
```
advaitjain, line 120: Let's use strcmp instead of this custom loop.

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -72,6 +77,17 @@ class MicroProfiler {
...
+  struct TicksPerTag {
+    const char* tag;
+    uint32_t ticks;
+  };
+  // In practice, the number of tags will be much lower than the number of
+  // events. But it is theoretically possible that each event to be unique and
+  // hence we allow total_ticks_per_tag to have kMaxEvents entries.
+  TicksPerTag total_ticks_per_tag[kMaxEvents];
+
+  int FindExistingOrNextPosition(const char* tagName);
```
advaitjain, line 89: tag_name: https://google.github.io/styleguide/cppguide.html#Variable_Names

advaitjain (changes requested):

tensorflow/lite/micro/micro_profiler.cc
tanmaydas82: Thank you for the suggestion.

tensorflow/lite/micro/micro_profiler.cc
tanmaydas82: done

tensorflow/lite/micro/micro_profiler.cc
tanmaydas82: I was not aware of if strcmp is available in tflm Thank you for the suggestion.

tensorflow/lite/micro/micro_profiler.h
tanmaydas82: Done. Missed the header file last time.

tensorflow/lite/micro/micro_profiler.cc
```diff
@@ -67,4 +68,54 @@ void MicroProfiler::LogCsv() const {
...
+#if !defined(TF_LITE_STRIP_ERROR_STRINGS)
+  MicroPrintf(
+      "\"Unique Tag\",\"Total ticks across all events with that tag.\"");
+  int total_ticks = 0;
+  for (int i = 0; i < num_events_; ++i) {
+    uint32_t ticks = end_ticks_[i] - start_ticks_[i];
+    if (tags_[i] == nullptr) {
+      MicroPrintf("Event does not have a tag name.");
+      continue;
+    }
```
advaitjain, line 81: can this happen? seems like this should be a DCHECK and not an error that we want to print and continue on.

tensorflow/lite/micro/micro_profiler.cc
```diff
@@ -67,4 +68,54 @@ void MicroProfiler::LogCsv() const {
...
+    int position = FindExistingOrNextPosition(tags_[i]);
+    if (position == -1) {
+      MicroPrintf("Number of tags exceeds the memory allocation.");
+      return;
+    }
```
advaitjain, line 86: again, this should not happen, right? We explicitly se the total number of unique tags to be equal to the total number of permitted events. 

DCHECK seems more appropriate.

advaitjain (changes requested):

tensorflow/lite/micro/micro_profiler.cc
tanmaydas82: Added a DCHECK. Thank you for the suggestion.

tensorflow/lite/micro/micro_profiler.cc
tanmaydas82: Right. I was probably overthinking. Added DCHECK.

advaitjain (approved):

PR #1301, opened 2022-08-02
syncing with google3 change

turbotoribio:
Seanet model has issues on future pixel phone which are believed to be related to reseting state when feature/model are toggled on and off. Added functionality to reset to overlap_add and framer ops in model which were the ops where different outputs were being produced on toggle due to no reset being available before. 

Corresponding g3 cl - [cl/464501489](https://critique.corp.google.com/464501489)

BUG=[b/220940833](https://b.corp.google.com/issues/220940833)


advaitjain (approved):

steven-buytaert:
Hi, with these changes, the free function is called on all registrations/nodes at the end of the AllocateTensors method. In other words, any memory allocated in the init function is released again. What if this allocated memory is required during invocation? Is it part of the contract that memory allocated during init is no longer available during invocation?

PR #1327, opened 2022-08-08
CMSIS-NN: Add offset check for int16 kernels

hmogensen-arm:
Add check that input and output tensors with int16 have offset zero
BUG=Fixes https://github.com/tensorflow/tflite-micro/issues/1364


tensorflow/lite/micro/kernels/pooling_common.cc
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2021-2022 The TensorFlow Authors. All Rights Reserved.
```
advaitjain, line 1: Please replace 2021 with 2022.

```
Copyright 2022
```

tensorflow/lite/micro/kernels/svdf_common.cc
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2020, 2022 The TensorFlow Authors. All Rights Reserved.
```
advaitjain, line 1: make this `Copyright 2022`

tensorflow/lite/micro/kernels/svdf_common.cc
```diff
@@ -487,6 +487,10 @@ TfLiteStatus PrepareSvdf(TfLiteContext* context, TfLiteNode* node) {
             &(data->scratch_output_tensor_index));
     TF_LITE_ENSURE_OK(context, scratch_output_status);
   } else {
+    if (input->type == kTfLiteInt16) {
```
advaitjain, line 490: SVDF does not support int16 activations. I would recommend leaving this file unchanged.

advaitjain (changes requested):
Thanks for the PR! Can you please add a BUG=<link to issue> to the PR description?

tensorflow/lite/micro/kernels/svdf_common.cc
advaitjain: could you revert the changes in svdf_common? or if you think this is needed, please let me know why.

advaitjain (changes requested):
for future, if a force-push can be avoided that would be great. It helps make sure we can review incremental changes.

What that means is that try not to `git rebase` but instead use `git merge`.

See this for more info: https://github.com/tensorflow/tflite-micro/blob/main/CONTRIBUTING.md#during-the-pr-review

tensorflow/lite/micro/kernels/svdf_common.cc
hmogensen-arm: My apologies, I was changing the wrong file. I will upload a new PR once it has passed the internal review

advaitjain (approved):

PR #1340, opened 2022-08-11
Micro conformer: concatenate should support bool.

imcgraw:
BUG=http://b/238904420


advaitjain (changes requested):
Could you also port over the [bool contact kernel tests from lite](https://github.com/tensorflow/tensorflow/blob/40059647dfb7947d02a8f09dc7fde408f51bd7f8/tensorflow/lite/kernels/concatenation_test.cc#L450-L470).

advaitjain (approved):

PR #1345, opened 2022-08-12
Micro conformer: support int32 in Add op.

imcgraw:
BUG=b/238904420

### DIFF
--- a/tensorflow/lite/micro/kernels/add.cc
+++ b/tensorflow/lite/micro/kernels/add.cc
@@ -30,28 +30,58 @@ limitations under the License.
 
 namespace tflite {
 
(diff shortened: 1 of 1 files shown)


advaitjain (changes requested):
could you port over the tests as well: https://github.com/tensorflow/tensorflow/blob/e268f00c0f2bfc166e3b9b53e4193f3df288e799/tensorflow/lite/kernels/add_test.cc#L256-L315

TFLM does not have any notion of TYPED_TEST_SUITE (since we don't and can't use gtest). If you could have the test cases be for int32 only that would be sufficient.

advaitjain:
Closing since this appears to not be necessary

PR #1353, opened 2022-08-15
Fix typo causing integration test failure.

njeffrie:
Hopefully takes something off the fixit list. I guess the GH test failure notifications did their job :)

BUG=fixes https://github.com/tensorflow/tflite-micro/issues/1352


advaitjain (approved):
Yes, definitely gets one item off the fixit list. Thanks!

Tagging @rkuester

PR #1363, opened 2022-08-18
Fix bug for multiple subgraphs in allocators

Tombana:
This fixes two bugs in the recording micro allocator and one in the regular micro allocator, all related to models with multiple subgraphs.

BUG=http://b/243820068


tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -510,13 +510,14 @@ TfLiteStatus MicroAllocator::FinishModelAllocation(
   }
 
   // Allocate scratch buffer metadata and buffers for variable tensors.
+  TF_LITE_ENSURE_STATUS(AllocateScratchBufferHandles(
+      scratch_buffer_handles, scratch_buffer_request_count_));
+
```
Tombana, line 515: `scratch_buffer_request_count_` is the number of scratch buffers for all subgraphs together, so it should not be inside the subgraph loop.

tensorflow/lite/micro/recording_micro_allocator.cc
```diff
@@ -163,10 +163,12 @@ TfLiteStatus RecordingMicroAllocator::AllocateNodeAndRegistrations(
 
   TfLiteStatus status =
       MicroAllocator::AllocateNodeAndRegistrations(model, subgraph_allocations);
+
+  RecordAllocationUsage(allocations,
+                        recorded_node_and_registration_array_data_);
```
Tombana, line 168: This was called in a subgraph loop, so it would count the same allocations `num_subgraphs` times. The same holds for line 198 below.

tensorflow/lite/micro/recording_micro_allocator.cc
```diff
@@ -176,8 +178,12 @@ TfLiteStatus RecordingMicroAllocator::AllocateNodeAndRegistrations(
...
     // decrementing by 1 and adding the actual number of nodes used in the
     // graph:
-    recorded_node_and_registration_array_data_.count +=
-        model->subgraphs()->Get(subgraph_idx)->operators()->size() - 1;
+    if (model->subgraphs()->Get(subgraph_idx)->operators()) {
+      recorded_node_and_registration_array_data_.count +=
+          model->subgraphs()->Get(subgraph_idx)->operators()->size() - 1;
+    } else {
+      recorded_node_and_registration_array_data_.count -= 1;
+    }
```
Tombana, line 186: It is possible for a subgraph to have no operators. An example is an `IF` node where only the true-branch has a body and the else-branch is empty. In that case `operators()` returns a nullpointer and this would segfault.

advaitjain (changes requested):
Thanks for the PR, the fixes look reasonable to me.

To get this merged, we would want to add test cases that fail without the fixes but passes with them. Do you have a small example model that we can use for this purpose?

Tombana:
> To get this merged, we would want to add test cases that fail without the fixes but passes with them. Do you have a small example model that we can use for this purpose?

I've looked through the tests in `micro_allocator_test.cc`, and there are multi-subgraph models available in `tflite::testing::` which are usable for this. I wrote a simple test, see below. However, there's no easy interface to `MicroAllocator` to test if `AllocateScratchBufferHandles` was called more than once. A second call will simply allocate more memory and overwrite the `scratch_buffer_handles` pointer. See the `// TODO` below.

I see the following options:
- assert on the exact amount of bytes allocated by the allocator. This might differ on 32-bit vs 64-bit platforms though.
- run the entire test in a loop, once without scratch buffers and once with scratch buffers. Then the total amount of bytes allocated should increase by `sizeof(ScratchBufferHandle) * num_scratch_buffers` and *not* times the number of subgraphs.

Do you have any suggestions and/or preferences?

```cpp
TF_LITE_MICRO_TEST(TestMultisubgraphNumScratchAllocations) {
  // Any model with multiple subgraphs:
  const tflite::Model* model =
      tflite::testing::GetSimpleModelWithSubgraphsAndIf();

  constexpr size_t arena_size = 1024;
  uint8_t arena[arena_size];
  tflite::MicroAllocator* allocator = tflite::MicroAllocator::Create(
      arena, arena_size, tflite::GetMicroErrorReporter());
  TF_LITE_MICRO_EXPECT(nullptr != allocator);

  tflite::SubgraphAllocations* subgraph_allocations =
      allocator->StartModelAllocation(model);
  TF_LITE_MICRO_EXPECT(nullptr != subgraph_allocations);

  // Request two scratch buffers
  int buffer_idx1;
  TF_LITE_MICRO_EXPECT_EQ(
      kTfLiteOk, allocator->RequestScratchBufferInArena(20, 0, &buffer_idx1));
  int buffer_idx2;
  TF_LITE_MICRO_EXPECT_EQ(
      kTfLiteOk, allocator->RequestScratchBufferInArena(30, 0, &buffer_idx2));

  tflite::ScratchBufferHandle* scratch_buffer_handles = nullptr;
  TF_LITE_MICRO_EXPECT_EQ(
      kTfLiteOk, allocator->FinishModelAllocation(model, subgraph_allocations,
                                                  &scratch_buffer_handles));

  // TODO: Check that AllocateScratchBufferHandles was only called once
}
```

advaitjain (changes requested):
Thanks for adding the test, it looks good! I pushed a commit with some minor edits.

Can I also request a test for the recording_micro_allocator changes?

Tombana:
@advaitjain I've added a test for the recording allocator and checked that it failed without the changes from this PR.

I did not manage to add a test for the `subgraph->operators() == nullptr` case, it seemed that this did not happen for the models in the tflite::testing infrastructure.

advaitjain (approved):

advaitjain:
> @advaitjain I've added a test for the recording allocator and checked that it failed without the changes from this PR.
> 
> I did not manage to add a test for the `subgraph->operators() == nullptr` case, it seemed that this did not happen for the models in the tflite::testing infrastructure.

Thanks! Let's get this PR merged. I have created an internal bug (http://b/243820068) to add a test case for the case where `subgraph->operators() == nullptr` so that it is on our task list. Feel free to send a PR for that if you'd like.

Issue #1336, opened 2022-08-11
No interface class for MicroProfiler


mansnils:
This feature https://github.com/tensorflow/tflite-micro/commit/4a8320a72772dba3190570b9b9163b0db86bc37f may signicantly increase stack usage or memory footprint for projects even not using the feature. A profiler class that inherits from MicroProfiler will increase it's size even though it can not access the array since it is private to MicroProfiler.
One option to remedy this could be to add a new Profiler inheriting from MicroProfiler with the additonal CSV log feature (calling it ExtendedMicroProfiler), and place for more features. However even better would be to have an abstract base class for MicroProfiler, with BeginEvent/EndEvent that MicroProfiler inherits from. Projects would then have the option to only override this interface profiler class. 
@advaitjain @tanmaydas82 What do you think?

advaitjain:
Having an interface class sounds good to me. We can start with what you have in #1376 and see if the interface needs to be made richer in the future.

PR #1376, opened 2022-08-24
Provide an interface class for MicroProfiler

mansnils:
BUG=Fixes https://github.com/tensorflow/tflite-micro/issues/1336


advaitjain (dismissed):

advaitjain (approved):

advaitjain (dismissed):

advaitjain:
@mansnils, this change will be a bit more involved if we try and replace all uses of MicroProfiler with MicroProfilerInterface (which would be good to do). I have pushed a few commits but there is more to be fixed. Let me know if you want to take a crack at the remaining changes.

mansnils:
@advaitjain, I think now all uses of MicroProfiler that can be replaced have been done so.

advaitjain (approved):

advaitjain (approved):

advaitjain (approved):

Issue #1382, opened 2022-08-28
Fix project generation for cortex_m_generic target


ddavis-2015:
@tensorflow/micro 

The project_generation TARGET_ARCH is missing from the cortex_m_generic makefile.  This causes the arduino-examples repo to fail to sync from the tflite-micro repo.

PR #1383, opened 2022-08-28
Project generation fix

ddavis-2015:
@tensorflow/micro 

Fix for project_generation TARGET_ARCH missing from cortex_m_generic makefile.  This should resolve the current issue with the arduino-examples repo not syncing from the tflite-micro repo.

bug=#1382


advaitjain (changes requested):
We removed `TARGET_ARCH=project_generation` with https://github.com/tensorflow/tflite-micro/pull/1167 and I don't remember the full context.

tagging @mansnils for more info there.

@ddavis-2015, will your issue be fixxed if you use `TARGET_ARCH=cortex-m4` for the sync in the Arduino repository?

advaitjain:
> @advaitjain No. Not having project_generation means that the micro_speech example is not part of project generation and therefore the kissfft download wont happen.

Got it, you are referring to this check: https://github.com/tensorflow/tflite-micro/blob/659618e0a5221d9046e1bcb2f828b46568846499/tensorflow/lite/micro/tools/make/targets/cortex_m_generic_makefile.inc#L201-L209

@mansnils, any concerns with the current PR being merged?

mansnils:
> > @advaitjain No. Not having project_generation means that the micro_speech example is not part of project generation and therefore the kissfft download wont happen.
> 
> Got it, you are referring to this check:
> 
> https://github.com/tensorflow/tflite-micro/blob/659618e0a5221d9046e1bcb2f828b46568846499/tensorflow/lite/micro/tools/make/targets/cortex_m_generic_makefile.inc#L201-L209
> 
> @mansnils, any concerns with the current PR being merged?

We removed it because the TARGET=cortex_m_generic build in tensorflow/lite/micro/tools/ci_build/test_project_generation.sh would otherwise fail. Reason is that different flags need to be set depending on TARGET_ARCH=cortex-m${ARM_CPU}. This affects PMU/DWT counters, i.e. that correct header file is included in tensorflow/lite/micro/cortex_m_generic/micro_time.cc.
Because of this we also had to select an Arm Cortex-M CPU in tensorflow/lite/micro/tools/ci_build/test_project_generation.sh https://github.com/tensorflow/tflite-micro/blob/b671143b0839d3f955dd31097121e240a0231b6d/tensorflow/lite/micro/tools/ci_build/test_project_generation.sh#L71, which means it will no longer break if adding back TARGET_ARCH=project_generation.
If adding back TARGET_ARCH=project_generation I suspect it could fail in a similar way as before in some other context.
Could an alternative be to introduce a new make variable to handle this micro_speech download logic?

advaitjain:
I have applied the `ci:run` label and let's see what errors we get. I would think that we should be able to figure out a solution that works for all use cases.

advaitjain (approved):
all the CI passed so we're happy.

mansnils:
Just for my understanding, why is TARGET_ARCH=project_generation not needed in test_project_generation.sh, should test_project_generation.sh not reflect a real use case? And same question from different angle, if TARGET_ARCH=project_generation is not needed in test_project_generation.sh, why is it needed elsewhere?
But as long as it works I am happy..

advaitjain:
> Just for my understanding, why is TARGET_ARCH=project_generation not needed in test_project_generation.sh, should test_project_generation.sh not reflect a real use case? And same question from different angle, if TARGET_ARCH=project_generation is not needed in test_project_generation.sh, why is it needed elsewhere? But as long as it works I am happy..

You are right. #1396 adds `TARGET=cortex_m_generic TARGET_ARCH=project_generation` to the CI (and makes some additional changes to enable that).

PR #1387, opened 2022-08-30
Changing the gen directory from tensorflow/lite/micro/make/gen to gen/.

tanmaydas82:
BUG=https://b.corp.google.com/issues/244204467

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 *google*
 *__pycache__*
 venv
+gen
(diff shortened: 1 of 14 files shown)


advaitjain (changes requested):
let's add `gen` to [the top-level .gitignore](https://github.com/tensorflow/tflite-micro/blob/main/.gitignore) and confirm that we won't inadvertently commit that into the repo.

and remove `gen` from https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/tools/make/.gitignore

advaitjain (approved):

PR #1389, opened 2022-08-30
Add reset function into the python wrapper

rewu93:
Some models such as LSTM are stateful, which requires state reset after each inference. This PR adds the model reset function, which has already been implemented in C++, into the TFLM python interpreter. 

BUG=http://b/244330968

### DIFF
--- a/tensorflow/lite/micro/python/interpreter/tests/interpreter_test.py
+++ b/tensorflow/lite/micro/python/interpreter/tests/interpreter_test.py
@@ -73,10 +73,6 @@ class ConvModelTests(test_util.TensorFlowTestCase):
       # TODO: Remove tolerance when the bug is fixed.
       self.assertAllLessEqual((tflite_output - tflm_output), 1)
 
-      # Reset the model state
-      tflm_interpreter.reset()
-      tflite_interpreter.reset_all_variables()
-
   def testModelFromFileAndBufferEqual(self):
     model_data = generate_test_models.generate_conv_model(True, self.filename)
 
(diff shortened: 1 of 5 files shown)


tensorflow/lite/micro/python/interpreter/tests/interpreter_test.py
```diff
@@ -73,6 +73,10 @@ def testCompareWithTFLite(self):
       # TODO: Remove tolerance when the bug is fixed.
       self.assertAllLessEqual((tflite_output - tflm_output), 1)
 
+      # Reset the model state
+      tflm_interpreter.reset()
+      tflite_interpreter.reset_all_variables()
```
paulinesho, line 78: I'd say don't add reset here because the conv model is stateless so resetting here doesn't do anything and will just cause confusion. 

`tflm_interpreter.reset()` and `tflite_interpreter.reset_all_variables()` are also not equivalent.

rewu93: Got it. Should we have `TfLiteStatus ResetVariableTensors()` included as well? But it seems that we are going to remove it in the future: https://github.com/tensorflow/tflite-micro/blob/639ae6378a5847a610e32926459c632c0b68fc17/tensorflow/lite/micro/micro_interpreter.h#L125

advaitjain (approved):
adding in Reset to the Python API sounds good.

We will be removing ResetVariableTensors completely: http://b/244457206

PR #1396, opened 2022-09-01
Test TARGET_ARCH=project_generation as part of our CI.

advaitjain:
After some back and forth, we have settled on keeping a `TARGET_ARCH=project_generation` as part of `TARGET=cortex_m_generic`. This PR adds testing of that functionality to our CI.

In order to reconcile the changes from https://github.com/tensorflow/tflite-micro/pull/1167 with `TARGET=cortex_m_generic TARGET_ARCH=project_generation` we use `if defined(PROJECT_GENERATION)` to sub out functions that were creating trouble in #1167 and add `-DPROJECT_GENERATION` to the Makefile that is used as part of test_project_generation.sh

BUG=see discussion in https://github.com/tensorflow/tflite-micro/pull/1383 for more context.


suleshahid (approved):
LGTM.

Issue #1397, opened 2022-09-01
gh pr create failing with TFLM pre-push hook


advaitjain:
`gh pr create` with the pre-push hook described in https://github.com/tensorflow/tflite-micro/blob/main/CONTRIBUTING.md#initial-setup is failing with the following error message:
```
Connection to github.com closed by remote host.
```

Not having the pre-push hook (e.g. `mv .git/hooks/pre-push .git/hooks/pre-push.bak`) makes the issue go away.

advaitjain:
Issue seems to be resolved by following https://stackoverflow.com/q/60833006

I did have to restart sshd with:
```
sudo systemctl restart ssh
```

PR #1398, opened 2022-09-01
Remove test target that does not exist.

advaitjain:
Chanced upon this while debugging why git push with a pre-push hook as described in https://github.com/tensorflow/tflite-micro/blob/main/CONTRIBUTING.md#initial-setup was failing.

Root cause of the push failing was fixed with https://stackoverflow.com/q/60833006 but this cleanup is still useful.

BUG=#1397


tanmaydas82 (approved):
LGTM!

PR #1399, opened 2022-09-02
Fix path to gen folder

mansnils:
BUG=https://github.com/tensorflow/tflite-micro/runs/8149595332?check_suite_focus=true


advaitjain (approved):
Thanks for the fix!

PR #1404, opened 2022-09-08
Fix the issue described in http://b/245773171

advaitjain:
This PR fixes the issue for one particular target, but there are still others with the exact same error.

The mako template is already correct as can be seen: 
https://github.com/tensorflow/tflite-micro/blob/87d2db744611937f5dfb818ef639b84370cffe44/tensorflow/lite/micro/integration_tests/templates/integration_tests_cc.mako#L52-L54

What this means is that many of the integration_tests.cc need to be regenerated with the updated mako template.

Currently, we do not have an easy way to regenerate the integration_tests if all we have is a change to the mako template.  Instead, we need to perform the full step of going from the model and regenerating all the tflite and golden input and output csvs. That is a problem and should be fixed.

Additionally, the integration_test targets are currently cc_binary instead of cc_test which means that we do not have any CI coverage for them. We should revisit that decision as well.

BUG=http://b/245773171


tanmaydas82 (approved):
LGTM. Thank you for the quick fix!

PR #1424, opened 2022-09-19
label refactor

jwithers:
see bug for details

BUG=https://issuetracker.google.com/issues/246664234


.github/workflows/apply_cirun.yml
```diff
@@ -16,7 +16,9 @@ jobs:
     runs-on: ubuntu-latest
     if: contains(github.event.pull_request.labels.*.name, 'ci:ready_to_merge')
     steps:
-      - uses: actions/github-script@v5
+      - name: add ci run
```
vamsimanchala, line 19: Please update the file comments to reflect the latest behavior. Maybe something like- "1. ci:ready_to_merge is added to a PR, by the mergify[bot]", would do?

jwithers: The comments as they exist are accurate. ci:ready_to_merge is added to a pr manually or by the butlerbot, not by mergify.

jwithers:
@advaitjain There is a structure we have a lot of in here: 
`(github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ci:run'))`

Since that is coming after a on: pull_request in almost all cases the github.event.name is redundant. Can you think of a reason we did this? If not, I would like to go through and get rid of all these except in the few cases where there is different behavior between the workflow_run and pull_request triggered versions of the same workflow.

advaitjain (approved):

Issue #1426, opened 2022-09-19
Speech training notebook fails because of Colab TF 1.x deprecation


petewarden:
Google Colabs removed support for the `%tensorflow 1.x` directive in August 2022. The [speech training notebook](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/examples/micro_speech/train/train_micro_speech_model.ipynb) still relies on 1.x functionality, and isn't easy to adapt.

I'll provide a workaround PR shortly as a suggested fix.

PR #1427, opened 2022-09-19
Fix TF 1.x deprecation speech training notebook

petewarden:
Google Colabs removed support for the `%tensorflow 1.x` directive in August 2022. This notebook still relies on 1.x functionality, and isn't easy to adapt, as noted in #1426, so I've suggested a change that switches to the 1.15 CPU version of the framework. This doesn't take advantage of the GPU, so it's a bit slower, but only increases the time to about two hours instead of one in my experiments, so I think it's the least-worst option.


petewarden:
Retrying the CLA

PR #1428, opened 2022-09-19
Add the example to train/evaluate quantized LSTM model on MNIST dataset using python

rewu93:
This PR provides an example of 
* Convert the trained keras LSTM model into the quantized .tflite model (int8 fully quantized). 
* Run inference on the quantized model using the TFLM python interpreter. 

It also extends the test code for quantized models. 

BUG=http://b/245386704


tensorflow/lite/micro/examples/mnist_lstm/evaluate_test.py
```diff
@@ -26,21 +26,27 @@
...
+  np.random.seed(42)  #Seed the random number generator
+
+  def testInputErrHandling(self):
+    wrong_size_image_path = os.path.join(PREFIX_PATH, "samples/resized9.png")
+    with self.assertRaisesWithPredicateMatch(ValueError,
+                                             "Invalid input image shape"):
+      evaluate.predict_image(self.tflm_interpreter, wrong_size_image_path)
 
   def testCompareWithTFLite(self):
     tflite_interpreter = tf.lite.Interpreter(model_path=self.model_path)
```
advaitjain, line 45: Let's start enforcing reference kernels with TfLite when we compare the TFLM interpreter against TfLite?

@paulinesho

advaitjain (commented):
looks good! One question (somewhat unrelated to this PR) about using the TfLite interpreter with reference kernels.

tensorflow/lite/micro/examples/mnist_lstm/evaluate_test.py
paulinesho: Agreed, so we can remove the +/- 1 tolerance.

rewu93: @paulinesho  the +/-1 tolerance is for the python interpreter test right? Bit exactness is not pursued here. The test logic is to compare TFLite float with TFLM float first. Then TFLM quant-dequant result is compared with TFLM float.  Lastly, the prediction accuracy of the quant model. Everything is compared in floats.

rewu93: The overall testing logic should be: TF/keras (source of truth) -> TFLITE (float) -> TFLM (float) ->  TFLM (quant). Reference kernels should be used when comparing across frameworks. Correct me if I am wrong

advaitjain: right, thanks for the explanation @rewu93. I still think we should be using the tflite interpreter with the reference kernels everywhere in the TFLM repo as a matter of best practice. I'm going to approve this PR and if you could file a bug and follow-up with a different PR that makes such a change across the board, that would be excellent.

advaitjain (approved):

PR #1440, opened 2022-09-23
Update the apply_cirun workflow to only trigger on a synchronize event.

advaitjain:
With https://github.com/tensorflow/tflite-micro/pull/1424, we are triggering all the workflows directly from the `ci:ready_to_merge_label`.

As a result, we no longer need to apply ci:run from a pull_request:labeled event.

Additionally, with the apply_cirun workflow in its current form, when a reviewer adds the `ci:ready_to_merge` label the workflow ends up removing it which interferes with the merge queue.

The correct behavior is to only automatically apply ci:run on a synchronize event (when the event comes from mergify).

BUG=http://b/246664234


advaitjain:
@jwithers: changes look correct to you?
@vamsimanchala: happy to discuss more if that helps with the review.

vamsimanchala (approved):
LGTM

PR #1443, opened 2022-09-24
Add virtual destructor to fix build error from `-Wnon-virtual-dtor`.

advaitjain:
Without the current change, (and if we are building with `-Wnon-virtual-dtor`) we will get the following build error:

```
./tensorflow/lite/micro/micro_profiler_interface.h:24:7: error: 'class tflite::MicroProfilerInterface' has virtual functions and accessible non-virtual destructor [-Werror=non-virtual-dtor]
   24 | class MicroProfilerInterface {
      |       ^~~~~~~~~~~~~~~~~~~~~~
```

The `-Wnon-virtual-dtor` warning is applied to internal Google builds but not to the TFLM GitHub builds (prior to this PR) which is why the issue was not caught by the presubmit checks for #1376.

The current change updates the build flags for both bazel and make to add `-Wnon-virtual-dtor` and have the OSS and internal builds be more consistent.

Manually verified that with the new warning, we get the same error on Github as we do internally. And hence we have confidence that the fix here (which passes all the GitHub presubmits) will also pass the internal checks.

BUG=#1336

### DIFF
--- a/tensorflow/lite/micro/build_def.bzl
+++ b/tensorflow/lite/micro/build_def.bzl
@@ -2,6 +2,7 @@ def micro_copts():
     return [
         "-Wall",
         "-Werror",
+        "-Wnon-virtual-dtor",
         "-DFLATBUFFERS_LOCALE_INDEPENDENT=0",
     ]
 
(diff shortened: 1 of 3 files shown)


vamsimanchala (approved):
LGTM

PR #1450, opened 2022-09-28
python: fix problems with extension module build

### LINKED ISSUE #1283: Cannot run tflm python interpreter test with bazel
I'm gettting some numpy errors when I run the bazel test tensorflow/lite/micro/tools/python_interpreter/tests:interpreter_test.
(https://github.com/tensorflow/tflite-micro/pull/1147#issue-1242402865)

Steps:
```
1. git clone tfl-micro
2.  bazel test tensorflow/lite/micro/tools/python_interpreter/tests:interpreter_test
```

OUTPUT:

```
In file included from tensorflow/lite/micro/python/interpreter/src/numpy_utils.cc:16:
./tensorflow/lite/micro/python/interpreter/src/numpy_utils.h:20:10: fatal error: numpy/ndarraytypes.h: No such file or directory
   20 | #include <numpy/ndarraytypes.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~
```

I'm using ...
```
Python 3.10.5
Numpy 1.23.1

```
I realize that I can get the numpy include directory from numpy.get_include() but I have never used bazel and cannot seem to find the right place to add that piece. I also presume that something else needs to be done than just pointing gcc to that include directory.

Any help would be appreciated.

I am trying to get this test running to ultimately debug some different outputs I'm seeing when I run a TFLite model in Python vs that model converted to TFLM on a Raspberry Pi Pico with the same input. I was hoping that I could see if I used this to test the tflmicro model to see if something is wrong with the implementation that's ported over for the Pico.

rkuester:
Fix problems with the Python extension module build:

- Explicitly build against downloaded NumPy headers instead of relying on the build OS implicitly via the compiler's built-in paths, fixing #1283.
- Use tagged, verified downloads of pybind11_bazel and pybind to silence noise in the bazel output and improve tamper resistance, reproducibility, and caching.

The underlying issue was that the extension module depends on header files from numpy but that dependency was not captured in the BUILD rules.

Upstream Tensorflow's solution is too complicated to import. They manage dependencies on Python runtimes and libraries via much more extensive system of [custom rules](https://github.com/tensorflow/tensorflow/tree/master/third_party/py), designed to meet their many additional requirements.

Background:
- [pybind/pybind11](https://github.com/pybind/pybind11) is a C++ library that helps in creating our extension's interface to CPython.
- [pybind/pybind11_bazel](https://github.com/pybind/pybind11_bazel) provides Bazel rules for building pybind-based extensions, but doesn't cover the unusual case of using C headers provided in another Python module (in our case, NumPy).
- Bazel's own [bazelbuild/rules_python](https://github.com/bazelbuild/rules_python) provides rules for fetching Python dependencies from PyPI for use when *running* Python code via Bazel, but doesn't directly help with the unusual case of *building* code using headers provided by Python packages from PyPI (again, in our case, NumPy).
- The standard way, in Bazel, to build against third-party code is to create an [external repository](https://bazel.build/docs/external) by writing and calling [repository rules](https://bazel.build/extending/repo).

This PR supplies the C headers from NumPy to the build of our extension by adding a repository (via the new repository rule `tflm_py_cc_headers`) that reuses the PyPI package downloaded by rules_python and wraps it in a `cc_library` target on which the build of the extension can depend.

BUG=fixes #1283

### DIFF
--- a/tensorflow/lite/micro/python/py_cc_headers.bzl
+++ b/tensorflow/lite/micro/python/py_cc_headers.bzl
@@ -1,0 +1,109 @@
+"Repository rule for creating a repository with the C headers of a Python package."
+
+# This extends the standard @rules_python rules to expose the C headers
+# contained in some Python packages like NumPy. It extends @rules_python rather
+# than duplicating the download mechanism, and to ensure the versions of Python
+# packages used throughout the WWORKSPACE are consistent. Exposing the headers
+# could conceivably become a build-in feature of @rules_python at some point.
+
+def _find_headers(ctx, py_library, include_prefix):
+    # Find the path to the headers within the @rules_python repository for the Python
+    # package `py_library`.
+
+    # WARNING: To get a flesystem path via ctx.path(), its argument must be a
+    # label to a non-generated file. ctx.path() does not work on non-file
+    # A standard technique for finding the path to a repository (see,
+    # e.g., rules_go) is to use the repository's BUILD file; however, the exact
+    # name of the build file is an implementation detail of @rules.python.
+    build = py_library.relative(":BUILD.bazel")
+    path = ctx.path(build).dirname
+
+    # Append the include_prefix, if any. get_child() needs one component at a
+    # time.
+    if include_prefix:
+        for c in include_prefix.split("/"):
+            path = path.get_child(c)
+
+    return path
+
+def _tflm_py_cc_headers_impl(ctx):
+    # Create a repository with the directory tree:
+    #     repository/
+    #               |- _include --> @rules_python//package/[include_prefix]
+    #               \_ BUILD
+
+    # Symlink to the headers within the @rules_python repository
+    src = _find_headers(ctx, ctx.attr.py_library, ctx.attr.include_prefix)
+    destdir = "_include"
+    ctx.symlink(src, destdir)
+
+    # Write a BUILD file publishing a cc_library() target for the headers.
+    BUILD = """\
+cc_library(
+    name = "%s",
+    hdrs = glob(["%s/**/*.h"], allow_empty=False),
+    includes = ["%s"],
+    visibility = ["@//tensorflow/lite/micro:__subpackages__"],
+)
+""" % (ctx.attr.name, destdir, destdir)
+    ctx.file("BUILD", content = BUILD, executable = False)
+
+tflm_py_cc_headers = repository_rule(
+    implementation = _tflm_py_cc_headers_impl,
+    local = True,
+    attrs = {
+        "py_library": attr.label(mandatory = True),
+        "include_prefix": attr.string(mandatory = False),
+    },
+)
+"""Repository rule for creating an external repository `name` with the C
+headers from a Python package published as a `py_library` target by
+@rules_python. The top-level Bazel package in the repository provides a
+cc_library target named `name` (the default target of the package) for use
+as a dependency in targets building against the headers.
+
+Use the optional `include_prefix` to base headers at a subdirectory of the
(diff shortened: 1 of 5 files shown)


rkuester:
I've updated the description with a theory of operation, linking to some relevant documentation.

tensorflow/lite/micro/python/py_cc_headers.bzl
```diff
@@ -0,0 +1,59 @@
...
+    if prefix:
+        for s in prefix.split("/"):
+            p = p.get_child(s)
+    return p
+
+def _tflm_py_cc_headers_impl(ctx):
+    # Symlink from the directory containing the headers, in the repository
+    # created by @python_rules pip_install(), to _destdir within this new
+    # repository.
+    any_file = py_bzl_package(ctx.attr.py_package) + ":BUILD.bazel"
```
advaitjain, line 27: let;s add a comment about what any_file is being used for. BUILd.bazel is created when numpy is downloaded as part of the bazel build.

tensorflow/lite/micro/python/py_cc_headers.bzl
```diff
@@ -0,0 +1,59 @@
...
+    bzl_package_path = ctx.path(Label(any_file)).dirname
+    src = _make_include_path(bzl_package_path, ctx.attr.py_package, ctx.attr.prefix)
```
advaitjain, line 29: give an example in the comments about what this line is doing?

tensorflow/lite/micro/python/py_cc_headers.bzl
```diff
@@ -0,0 +1,59 @@
+"Repository rule for creating a repository with the C headers of a python package."
+
```
rkuester, line 2: Explain the end state of the created repository.

tensorflow/lite/micro/python/py_cc_headers.bzl
```diff
@@ -0,0 +1,59 @@
...
+        "prefix": attr.string(mandatory = True),
+    },
+)
+"""Repository rule for creating a new repository `name` with a top-level
+Bazel package containing the C headers from the Python distribution package
+named `py_package`. The top-level Bazel package provides a target ":cc_library"
+to be used as a dependency in targets building against the headers.
+
+For example:
+
```
advaitjain, line 50: let's use this location for a full illustration of the steps:

1. how to create a repo with the repository_rule defined here
2. what the directory structure in .bazel/cache/<>/ looks like
3. how this new repository is used from a BUILD target that depends on it.

tensorflow/lite/micro/python/py_cc_headers.bzl
rkuester: Comment added within repository rule implementation.

tensorflow/lite/micro/python/py_cc_headers.bzl
rkuester: Comment added.

tensorflow/lite/micro/python/py_cc_headers.bzl
rkuester: Implementation refactored to be more clear, and comments added.

tensorflow/lite/micro/python/py_cc_headers.bzl
rkuester: Extensive comment added. #2 covered by a comment up in the implementation, as the rule user (the audience of this particular comment) shouldn't care.

advaitjain (approved):

PR #1451, opened 2022-09-28
fix all broken integration test in TFLM directory

turbotoribio:
see bug - https://b.corp.google.com/issues/245773171 for context, but essentially a few of the integration test need to be regenerated so that they match the mako template and run.  Since a simple system doesn't exist without remaking all the existing files in those directories I added the change that were needed manually and and not by running the micro/integration_tests/generate_per_layer_tests.py script. 

BUG= https://b.corp.google.com/issues/245773171


advaitjain (changes requested):
can you add to the PR description that the changes were made manually (and not by running the `micro/integration_tests/generate_per_layer_tests.py` script?

also, the PR description is missing BUG=

advaitjain (approved):

PR #1456, opened 2022-10-05
Run the test_project_generation from the parent dir of tflite-micro

tanmaydas82:
BUG=https://b.corp.google.com/issues/249442305

### DIFF
--- a/tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py
+++ b/tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py
@@ -44,11 +44,10 @@ def _get_dirs(file_list):
   return dirs
 
 
-def _get_file_list(key, makefile_options, args):
+def _get_file_list(key, makefile_options, tensorflow_root):
   params_list = [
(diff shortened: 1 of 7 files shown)


tensorflow/lite/micro/tools/project_generation/create_tflm_tree.py
```diff
@@ -233,8 +257,10 @@ def main():
   # no longer be needed once all the downloads are switched over to bash
   # scripts.
   params_list = [
-      "make", "-f", "tensorflow/lite/micro/tools/make/Makefile",
-      "third_party_downloads"
+      "make", "-f",
```
advaitjain, line 260: per our discssion, we will try and use `TENSORFLOW_ROOT` and `EXTERNAL_DIR` from `ar.makefile_options`.

tanmaydas82: Done! Thanks for the suggestions.

advaitjain (approved):

PR #1460, opened 2022-10-09
Micro profiler additions

### LINKED ISSUE #1454: Expand functionality of MicroProfiler
@tensorflow/micro 

Proposed additions to the MicroProfiler API to support models that consume streaming data (such as the micro_speech example).  These additions allow the accumulation of event data across multiple inference operations, along with extended reporting of the per tag event data.

```
  // Logging for models that process streaming data and want the log
  // output of the cumulative stats only at the end of the data stream.
  // Typical usage would be:
  // 1) Invoke
  // 2) UpdateTotalTicksPerTag
  // 3) ClearEvents
  // 4) Repeat from 1 until data stream exhausted
  // 5) LogTicksPerTagCsvExtended
  // 6) ClearTotalTicksPerTag
  // 7) Repeat from 1 with new data stream

  // For each unique tag, prints in CSV format:
  //   tag, total ticks, event count, average ticks/tag, percent of total time
  // Output will have one row for each unique tag along with the
  // total ticks summed across all events with that particular tag.
  void LogTicksPerTagCsvExtended();

  // Updates the ticks and event count for each unique tag in the current
  // list of events.
  void UpdateTotalTicksPerTag();

  // Clear the ticks and event count for each unique tag for which an
  // event was generated.
  void ClearTotalTicksPerTag();
```

TicksPerTag expanded with `total_events` field:

```
 private:
  struct TicksPerTag {
    const char* tag;
    uint32_t ticks;
    uint32_t total_events;
  };
```

ddavis-2015:
@tensorflow/micro 

Additional methods for MicroProfiler.

bug=fixes #1454


ddavis-2015:
This PR is withdrawn in favor of PR https://github.com/tensorflow/tflite-micro/pull/1648

PR #1462, opened 2022-10-10
add support for MicroResourceVariables from python wrapper

suleshahid:
Adding support for MRVs from python interpreter.

* Also had to change interpreter function call to using the MicroAllocator, as MRV is only able to be created with an Allocator object.
* The python function call remains the same, with an optional additional argument for the number of MRVs.
* Future feature request is to calculate the number of MRVs automatically and original python call could be reverted so user does not have to worry about this.

BUG=[251832638](https://b.corp.google.com/issues/251832638)


tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
```diff
@@ -203,7 +203,7 @@ InterpreterWrapper::~InterpreterWrapper() {
 
 InterpreterWrapper::InterpreterWrapper(
     PyObject* model_data, const std::vector<std::string>& registerers_by_name,
-    size_t arena_size) {
+    size_t arena_size, const int num_resource_variables) {
```
paulinesho, line 206: `const` has no effect here so it's encouraged to leave it out. (see [go/totw/109](https://goto.google.com/totw/109))

paulinesho: Can we create some basic tests that passes in a non-zero `num_resource_variables` from Python? LGTM otherwise.

suleshahid: > `const` has no effect here so it's encouraged to leave it out. (see [go/totw/109](https://goto.google.com/totw/109))

If I am not misunderstanding from the go link, this should be fine because it is for the definition of the function, so it will still provide functionality in that the variable will be a constant in scope of the function definition. I should however, remove it from interpreter_wrapper.h line 29, as that is the function declaration in which case it is unnecessary?

suleshahid: > Can we create some basic tests that passes in a non-zero `num_resource_variables` from Python? LGTM otherwise.

Added a simple additional test in the interpreter to make sure nothing breaks with calling with MRV parameters. Also, by using the test for verifying fromBytes and fromFile are equal, it ensures that change to use allocator and MRV still provides correct execution.

paulinesho (approved):

PR #1463, opened 2022-10-11
Python interpreter test using a hand-made flatbuffer model

rewu93:
## Issue
Previous test on the python interpreter used an end to end CNN, which introduces complexity and randomness. 

## Solution
Replace the CNN model to a handmade ADD model to test the python interpreter. This makes the test 100% deterministic and controllable. This also serves an example of creating the flatbuffer from scratch using python. 

BUG=http://b/248061370

### DIFF
--- a/tensorflow/lite/micro/python/interpreter/tests/test_model.py
+++ b/tensorflow/lite/micro/python/interpreter/tests/test_model.py
@@ -1,0 +1,296 @@
+# Copyright 2022 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Create a sample model for testing purposes.
+
+A simple quantized int8 model with single ADD operation.
+"""
+
+import flatbuffers
+import numpy as np
+
+from tflite_micro.tensorflow.lite.python import schema_py_generated as schema_fb
+
+TFLITE_SCHEMA_VERSION = 3
+
+
+def _generate_quantization_parameters(builder, rmin, rmax, qmin, qmax):
+  """Tensor-wise quantization parameter generation
+  """
+
+  def caclulate_scale(rmin, rmax, qmin, qmax):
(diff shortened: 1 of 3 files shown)


tensorflow/lite/micro/python/interpreter/tests/test_model.py
```diff
@@ -0,0 +1,296 @@
...
+  schema_fb.QuantizationParametersAddMin(builder, quant_min_offset)
+  schema_fb.QuantizationParametersAddMax(builder, quant_max_offset)
+  schema_fb.QuantizationParametersAddScale(builder, quant_scale_offset)
+  schema_fb.QuantizationParametersAddZeroPoint(builder,
+                                               quant_zero_point_offset)
+  quantization_offset = schema_fb.QuantizationParametersEnd(builder)
+  return quantization_offset
+
+
+def build_mock_flatbuffer_model():
```
paulinesho, line 83: I'm sorry for the late review, this PR fell off my radar. 

A few comments around maintainability:
1. This is a hefty function to create a simple model with add. Can anything here be parameterized and abstracted away so that the next time we want to create another simple model with a different op, we don't have to replicate this or even attempt to understand the details here? For example, maybe we can break this long function into multiple helper functions?
2. I can see this utility being reused over and over for perhaps different or other use cases. Perhaps it's a good time for `tensorflow/lite/micro/python/utils`?
3. Can we have clearer function names and filenames? For example, if this file or this function's main export is a simple add model, we could make it super specific like `build_mock_add_flatbuffer_model()` so that it sets a good example for future additions.

tensorflow/lite/micro/python/interpreter/tests/test_model.py
```diff
@@ -0,0 +1,296 @@
...
+import flatbuffers
+import numpy as np
+
+from tflite_micro.tensorflow.lite.python import schema_py_generated as schema_fb
+
+TFLITE_SCHEMA_VERSION = 3
+
+
+def _generate_quantization_parameters(builder, rmin, rmax, qmin, qmax):
+  """Tensor-wise quantization parameter generation
```
paulinesho, line 29: It's not too hard to guess in this case, but I think some comments on what the input arguments are would be worthwhile.

tensorflow/lite/micro/python/interpreter/tests/test_model.py
```diff
@@ -0,0 +1,296 @@
...
+  schema_fb.ModelAddBuffers(builder, buffers_offset)
+  schema_fb.ModelAddSignatureDefs(builder, signature_defs_offset)
+  model_offset = schema_fb.ModelEnd(builder)
+  builder.Finish(model_offset)
+  model = builder.Output()
+
+  return model
+
+
+def load_model_from_flatbuffer(flatbuffer_model):
```
paulinesho, line 286: Is this also a util helper that we should hide? i.e. append `_`?

PR #1468, opened 2022-10-12
Download CMSIS-NN separately and update header paths

mansnils:
Adding new download of CMSIS-NN since it is moving to its own repository.
Also updates some lint errors.

BUG=CMSIS-NN is moving to its own repo


tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
```diff
@@ -64,6 +52,6 @@ THIRD_PARTY_CC_HDRS += \
 # code without any modifications.
 INCLUDES += \
   -I$(CMSIS_PATH) \
+  -I$(CMSIS_NN_PATH) \
```
advaitjain, line 55: let's remove this and instead add `-I(DOWNLOADS_DIR)` to [this list](https://github.com/tensorflow/tflite-micro/blob/bea0931154d8ec38afc06a64ae1244c38a6f848f/tensorflow/lite/micro/tools/make/Makefile#L81-L87).

That will allow us to have the include paths be `#include "cmsis_nn/Include/..."` which would preferable.

tensorflow/lite/micro/kernels/cmsis_nn/add.cc
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 
 #include "tensorflow/lite/kernels/internal/reference/add.h"
 
-#include "CMSIS/NN/Include/arm_nnfunctions.h"
+#include "Include/arm_nnfunctions.h"
```
advaitjain, line 18: I added a comment in cmsis_nn.inc to change the -I so that the include path starts with `cmsis_nn`.

tensorflow/lite/micro/kernels/cmsis_nn/add.cc
```diff
@@ -103,19 +103,20 @@ TfLiteStatus CalculateOpData(TfLiteContext* context, TfLiteAddParams* params,
...
-  op_params.input1_multiplier = data->input1_multiplier;
-  op_params.input1_shift = data->input1_shift;
-  op_params.input2_offset = data->input2_offset;
-  op_params.input2_multiplier = data->input2_multiplier;
-  op_params.input2_shift = data->input2_shift;
-  op_params.output_offset = data->output_offset;
-  op_params.output_multiplier = data->output_multiplier;
-  op_params.output_shift = data->output_shift;
+void UpdateOpParams(tflite::ArithmeticParams* const op_params,
+                    const OpData* data) {
```
advaitjain, line 107: is there a reason to change from reference to pointer in this PR?

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
```diff
@@ -19,41 +19,29 @@ ifeq ($(CMSIS_PATH), $(CMSIS_DEFAULT_DOWNLOAD_PATH))
...
+ifeq ($(CMSIS_NN_PATH), $(CMSIS_NN_DEFAULT_DOWNLOAD_PATH))
+  DOWNLOAD_RESULT := $(shell $(MAKEFILE_DIR)/ext_libs/cmsis_nn_download.sh ${DOWNLOADS_DIR} ${TENSORFLOW_ROOT})
+  ifneq ($(DOWNLOAD_RESULT), SUCCESS)
+    $(error Something went wrong with the CMSIS-NN download: $(DOWNLOAD_RESULT))
+  endif
+endif
 
 ifeq (,$(CMSIS_NN_LIBS))
-  THIRD_PARTY_KERNEL_CC_SRCS += $(call recursive_find,$(CMSIS_PATH)/CMSIS/NN/Source,*.c)
+  THIRD_PARTY_KERNEL_CC_SRCS += $(call recursive_find,$(CMSIS_NN_PATH)/Source,*.c)
```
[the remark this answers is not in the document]
mansnils: Done

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
mansnils: I prefer to keep it like before. If using CMSIS_NN_PATH option, i.e. an external repo the path may not be cmsis_nn/Include. Ok?

tensorflow/lite/micro/kernels/cmsis_nn/add.cc
mansnils: See my comment above

tensorflow/lite/micro/kernels/cmsis_nn/add.cc
mansnils: Reason was to fix lint errors.
https://github.com/tensorflow/tflite-micro/blob/main/CONTRIBUTING.md#before-submitting-your-pr

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
felix-johnny: Since in TFLM we provide an [option](https://github.com/tensorflow/tflite-micro/pull/1468/files#diff-11d69e56729e5b7cdd1845a3d72024600ac4323334f49d5dc55028273fc7c859R28) to a user to use their own version of CMSIS-NN, having 'cmsis_nn/' would force them to use a downloaded folder name that is set by the glue file. That might not be possible for users. 

  For that, I would prefer use of 'Include/*.h' as well. 
   I understand that using 'cmsis_nn/Include' might be to avoid ambiguities, but since the NN folder from the existing 'CMSIS/NN' will be removed, there shouldn't be any conflict of the NN includes between the existing CMSIS_5 repository and the new CMSIS-NN repository.

mansnils:
@advaitjain and @felix-johnny My apologies! I accidentally overwrote my topic branch for this PR. I recreated the branch and reopened the PR but it may be harder to review it now.

mansnils:
@advaitjain ready_to_merge label was applied but it was not merged, does it not need approval?

advaitjain (approved):

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
advaitjain: Chatted with @mansnils and conclusion was that we'll keep as is. It might mean some changes to the Arduino sync but it should not be too much work.

PR #1472, opened 2022-10-13
Add 4-bit conv kernel support

paulinesho:
Add 4-bit conv kernel support

It is the open source equivalent of and already reviewed and approved http://cl/480909268.

BUG=b/248328557


tensorflow/lite/micro/memory_helpers.cc
```diff
@@ -90,6 +90,9 @@ TfLiteStatus TfLiteTypeSizeOf(TfLiteType type, size_t* size) {
     case kTfLiteComplex128:
       *size = sizeof(double) * 2;
       break;
+    case kTfLiteInt4:
```
advaitjain, line 93: let's add a comment describing why describing `kTfLiteInt4` has size of 1 byte is what we want.

advaitjain (commented):
Does this PR correspond to http://cl/480909268? If so, happy to approve as-is and address comments in follow-up PR.

If not, this would be a good PR to add a kernel test for 4 bit conv.

advaitjain:
> Does this PR correspond to http://cl/480909268? If so, happy to approve as-is and address comments in follow-up PR.
> 

Let's update the PR description to reflect that it is the open source equivalent of and already reviewed and approved http://cl/480909268

paulinesho:
> > Does this PR correspond to http://cl/480909268? If so, happy to approve as-is and address comments in follow-up PR.
> 
> Let's update the PR description to reflect that it is the open source equivalent of and already reviewed and approved http://cl/480909268

It's the equivalent. Sure.

advaitjain (approved):

PR #1481, opened 2022-10-19
refactor to use workflow_call strategy

jwithers:
See bug link for more details. This is expected to replace #1471 after talking with @advaitjain about the benefits.

BUG=https://issuetracker.google.com/issues/246664234


.github/workflows/tests_entry.yml
```diff
@@ -0,0 +1,95 @@
...
+          github-token: ${{ secrets.TFLM_BOT_REPO_TOKEN }}
+          script: |
+            github.rest.issues.removeLabel({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              name: 'ci:run'
+            })
+  
+  call-ci:
```
advaitjain, line 69: should we add `needs: ci-ready-to-merge` to all the `call-` jobs?

https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#defining-prerequisite-jobs

jwithers: Oh yes. We definitely should. Good catch, thanks.

.github/workflows/tests_entry.yml
```diff
@@ -0,0 +1,95 @@
...
+  no-labels:
+    runs-on: ubuntu-latest
+    if: ${{ !(contains(github.event.pull_request.labels.*.name, 'ci:run') ||
+        contains(github.event.pull_request.labels.*.name, 'ci:ready_to_merge')) }}
+    steps:
+      - run: exit 1
+  some-labels:
+    runs-on: ubuntu-latest
+    if: ${{ (contains(github.event.pull_request.labels.*.name, 'ci:run') ||
+        contains(github.event.pull_request.labels.*.name, 'ci:ready_to_merge')) }}
```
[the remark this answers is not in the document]
jwithers: Eh, turns out this isn't as clean as we would like. Putting the `if` condition at the job level means the job is skipped entirely if not met. Any jobs after that will skip if they have a `needs` on that job. So while we can separate each label conceptually this way, we still need a little bit of ugly if logic in the steps. Still doing it, but it isn't as pretty as your code above would have been if the skip behavior would have been different.
And using the `if: always()` dodge to make sure the following job runs even if the previous job skips won't work because that will trap our exits and make sure our tests will always run no matter what.

.github/workflows/tests_entry.yml
```diff
@@ -0,0 +1,95 @@
...
+          github-token: ${{ secrets.TFLM_BOT_REPO_TOKEN }}
+          script: |
+            github.rest.issues.removeLabel({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              name: 'ci:ready_to_merge'
+            })
+      - name: fail-on-bad-synch
+        if: ${{ (contains(github.event.pull_request.labels.*.name, 'ci:ready_to_merge') &&
```
advaitjain: this should not be dependent on `ci:ready_to_merge`. It should fail if there is a sync event not from mergify (independent of what label exists).

does that sound right to you?
```
if !env.SYNC_BY_MERGIFY
run: exit 1
```

.github/workflows/tests_entry.yml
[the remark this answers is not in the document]
advaitjain: understood. perhaps we should make some environment variables for the different expressions?

https://docs.github.com/en/actions/learn-github-actions/expressions#example-setting-an-environment-variable

```yaml
env:
  SYNC_BY_MERGIFY: ${{ (github.event.action == 'synchronize') && !(github.event.sender.login == 'mergify[bot]') }}
```

.github/workflows/tests_entry.yml
```diff
@@ -0,0 +1,95 @@
...
+        contains(github.event.pull_request.labels.*.name, 'ci:ready_to_merge')) }}
+    steps:
+      - run: exit 1
+  some-labels:
+    runs-on: ubuntu-latest
+    if: ${{ (contains(github.event.pull_request.labels.*.name, 'ci:run') ||
+        contains(github.event.pull_request.labels.*.name, 'ci:ready_to_merge')) }}
+    steps:
+      - name: remove-ready-to-merge
+        if: ${{ (contains(github.event.pull_request.labels.*.name, 'ci:ready_to_merge') &&
```
advaitjain, line 39: we could have this be:

```yaml
if: ${{ (contains(github.event.pull_request.labels.*.name, 'ci:ready_to_merge') && !env.SYNC_BY_MERGIFY
```

advaitjain (changes requested):

.github/workflows/tests_entry.yml
jwithers: So for all of these take a look at what I just pushed. It was just a matter of moving the logic into the steps instead of at the job level. While that basically reverted the logic line to what we had in the original version, it still provides separation of the tags. I don't think doing more here would help a lot with readability or conceptual clarity.

.github/workflows/tests_entry.yml
jwithers: Made all prep jobs sequential and added last in tag logic chain `ci-run` to all the `call-` jobs.

.github/workflows/run_xtensa.yml
```diff
@@ -0,0 +1,30 @@
+name: Run-Extensa
```
advaitjain, line 1: Run-Xtensa

.github/workflows/run_xtensa.yml
```diff
@@ -0,0 +1,30 @@
...
+  schedule:
+    # 10am UTC is 3am or 4am PT depending on daylight savings.
+    - cron: '0 10 * * *'
+
+  # Allow manually triggering of the workflow.
+  workflow_dispatch: {}
+
+jobs:
+  call-extensa-dispatched:
+    uses: ./.github/workflows/extensa.yml
```
advaitjain, line 13: nit: xtensa not extensa

Issue #1482, opened 2022-10-19
CMSIS-NN optimized pooling kernel missing int16 support


annietllnd:
As per the title, the int16 support should be added to the glue

PR #1483, opened 2022-10-19
Add int16 support for CMSIS-NN optimized pooling kernel

annietllnd:
BUG=#1482


tensorflow/lite/micro/kernels/pooling.h
```diff
@@ -71,14 +71,28 @@ void MaxPoolingEvalQuantized(TfLiteContext* context, TfLiteNode* node,
 TfLiteRegistration Register_AVERAGE_POOL_2D_INT8();
 
 TfLiteRegistration Register_MAX_POOL_2D_INT8();
+
+TfLiteRegistration Register_AVERAGE_POOL_2D_INT16();
```
advaitjain, line 75: Very minor comment:

Let's group by the kernel instead of the quantization.

So have max_pool int8 and int16 next to each other.

And then average_pool int8 and int16 in a separate block of code.

annietllnd: Good point, done

mansnils:
Merging this PR with https://github.com/tensorflow/tflite-micro/pull/1591 and hence closing this one.

PR #1491, opened 2022-10-21
Fix casting to avoid precision loss when mixing double and floats

UlrikHjort:
BUG=#1492


UlrikHjort:
@advaitjain  Do you have a hint what it is "Tests Entry Point / no-labels (pull_request_target)" fails on ?

advaitjain:
> @advaitjain Do you have a hint what it is "Tests Entry Point / no-labels (pull_request_target)" fails on ?

If you can bring your PR up to date with main that should fix the issue.

advaitjain (changes requested):
In general, doing the computation in double precision isn't really necessary and mostly exists to be able to share code with TfLite.

Is there a specific example of a where the changes from this PR are necessary?

UlrikHjort:
> In general, doing the computation in double precision isn't really necessary and mostly exists to be able to share code with TfLite.
> 
> Is there a specific example of a where the changes from this PR are necessary?

The details for the rationale for this change is described in PR  #1500 from @SebastianBoblest :

 "We tested this code with the CMSIS-NN backend of tflite-micro and see slight differences in the computed multipliers with computations performed in float precision and double precision. Also we experienced numerical differences between tensorflow lite and tflite-micro for some models with the CMSIS-NN backend and suspected that these casting issues might be one reason for this."
 
 @vdkhoi

advaitjain (approved):

PR #1498, opened 2022-10-26
move pr has BUG= check into tests_entry.yml

jwithers:
Does what it says on the tin. Conceptually these tests belong in tests_entry.yml and this fixes an inconsistent triggering issue when labels are applied by bots.

BUG=https://issuetracker.google.com/issues/255646154


.github/workflows/tests_entry.yml
```diff
@@ -69,31 +70,43 @@ jobs:
...
+    steps:
+      - name: Check for BUG=
+        if: ${{ !contains(github.event.pull_request.body, 'BUG=') }}
+        run: |
+          echo "PR description requires a BUG= line with issue number."
+          echo "See https://testing.googleblog.com/2017/09/code-health-providing-context-with.html for additional context"
+          exit 1
   
   call-ci:
-    needs: ci-run
```
advaitjain, line 74: let's keep `needs:ci-run` here and below.

Its handy to have the `call-*` jobs run in parallel with `pr-has-bug`

jwithers: I think I am misunderstanding something. So if pr has bug fails the merge/checks fail and we have to rerun the suite anyway. If that is the case, we want them to fail fast to save time. So we would want this check, which is only a few seconds, to run before getting into the crunchy stuff. What am I seeing wrong there?

jwithers: Understand use case after call with @advaitjain and reverting to `needs:ci-run`

advaitjain (approved):

PR #1508, opened 2022-10-31
Add README for Python interpreter

paulinesho:
BUG=b/256650563


paulinesho:
> LGTM. One comment is that most python users may not be familiar with Bazel. People (like old me) may get confused why python code need to be compiled. However this will not be an issue after we get the PyPi package. In the future we should consider create a Colab tutorial to walk through the essential features, which fit the appetite of python users.

Thanks for reviewing! Agree that a colab would be nice in the future.

rewu93 (approved):

PR #1531, opened 2022-11-07
Remove noasan, nomsan, noubsan tags for tflm_runtime

paulinesho:
BUG=b/236154360

These tags are no longer required now that b/236154360 is fixed.

### DIFF
--- a/tensorflow/lite/micro/python/interpreter/tests/BUILD
+++ b/tensorflow/lite/micro/python/interpreter/tests/BUILD
@@ -8,10 +8,7 @@ py_test(
     main = "interpreter_test.py",
     python_version = "PY3",
     tags = [
-        "noasan",
-        "nomsan",  # Python doesn't like these symbols from interpreter_wrapper_pybind.so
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/python/interpreter/tests/BUILD
```diff
@@ -7,12 +7,6 @@ py_test(
     srcs = ["interpreter_test.py"],
     main = "interpreter_test.py",
     python_version = "PY3",
-    tags = [
-        "noasan",
-        "nomsan",  # Python doesn't like these symbols from interpreter_wrapper_pybind.so
-        "notap",  # TODO(b/247808903)
```
advaitjain, line 13: I think we want to keep `notap`: http://b/247808903

advaitjain (approved):

paulinesho:
Unfortunately we can't quite remove these on bazel yet. The issue is slightly different, so created a new bug b/258036822 and abandoning this PR.

PR #1537, opened 2022-11-08
Revert "Fix casting to avoid precision loss when mixing double and floats"

advaitjain:
Reverts tensorflow/tflite-micro#1491

@mansnils from the CMSIS-NN team and I just chatted. This PR and https://github.com/tensorflow/tflite-micro/pull/1501 are actually introducing a difference between Lite and Micro. If we do need these changes they would have to be made in both the TFLM and the TfLite kernels.

Before going down that path (which would involve a reasonable amount of time and effort from everyone involved), we would like to better understand what the issue reported in https://github.com/tensorflow/tflite-micro/issues/1492 is and if that can be reproduced by the CMSIS-NN team as well.

BUG=https://github.com/tensorflow/tflite-micro/issues/1492


suleshahid (approved):
LGTM

Issue #1337, opened 2022-08-11
tensorflow/lite/micro/tools/ci_build/test_x86.sh does not compile with gcc 12.1


mansnils:
gcc-12.1 warns that size of metadata->name()->c_str() (it sees as 0) is smaller than strlen(kOfflineMemAllocMetadata) (which is 23):
https://github.com/tensorflow/tflite-micro/blob/eacc122d917605288e694220a5074c17750292f7/tensorflow/lite/micro/micro_allocation_info.cc#L326

Error:
tensorflow/lite/micro/micro_allocation_info.cc: In member function ‘TfLiteStatus tflite::AllocationInfoBuilder::GetOfflinePlannedOffsets(const int32_t**)’:
tensorflow/lite/micro/micro_allocation_info.cc:326:18: error: ‘int strncmp(const char*, const char*, size_t)’ specified bound 23 exceeds source size 0 [-Werror=stringop-overread]

advaitjain:
With https://github.com/tensorflow/tflite-micro/pull/1477 we are seeing an error with gcc-12 again.

https://github.com/tensorflow/tflite-micro/pull/1544 is suppressing the warning for now, but it would be good to see what better fix would be.

We'll also update CI to use gcc-12.

advaitjain:
Closing this issue since it should be fixed with #1550. Let's reopen if it is still not fixed.

PR #1544, opened 2022-11-10
Do not consider warning related to string overread as error.

tanmaydas82:
BUG=https://b.corp.google.com/issues/258039184 and #1337

### DIFF
--- a/tensorflow/lite/micro/tools/make/Makefile
+++ b/tensorflow/lite/micro/tools/make/Makefile
@@ -150,10 +150,9 @@ COMMON_FLAGS := \
   $(CC_WARNINGS) \
   $(ADDITIONAL_DEFINES)
 
-ifneq ($(TARGET), xtensa)
-  ifneq ($(TARGET), hexagon)
-    COMMON_FLAGS += -Wno-stringop-overread
-  endif
+# no-stringop-overread is not a valid flag for some toolchains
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -150,6 +150,12 @@ COMMON_FLAGS := \
   $(CC_WARNINGS) \
   $(ADDITIONAL_DEFINES)
 
+ifneq ($(TARGET), xtensa)
+  ifneq ($(TARGET), hexagon)
```
advaitjain, line 154: let's make this:
```make
# no-stringop-overread is not a valid flag for some toolchains
ifeq ($(findstring $(TARGET),xtensa hexagon),)
  COMMON_FLAGS += -Wno-stringop-overread
endif
```

that will make it easier to add more targets, if needed.

advaitjain (changes requested):

tensorflow/lite/micro/tools/make/Makefile
tanmaydas82: Great suggestion! Learned it. Thank you!

advaitjain (approved):

advaitjain:
@mergifyio refresh

PR #1554, opened 2022-11-15
Allow 3D output shape for FC layer.

cad-audio:
3D output shape support for Fully_Connected layer under xtensa.

The fix to the reference kernels was made with https://github.com/tensorflow/tensorflow/pull/54223 (and the corresponding import from TF Lite to TFLM). TFLM currently does not have a test case for this (corresponding to what was added with https://github.com/tensorflow/tensorflow/pull/54223/),

BUG=https://github.com/tensorflow/tensorflow/issues/53501


tensorflow/lite/micro/kernels/xtensa/fully_connected.cc
```diff
@@ -28,6 +28,8 @@ limitations under the License.
 #include "tensorflow/lite/micro/kernels/xtensa/xtensa_fully_connected.h"
 #include "tensorflow/lite/micro/micro_log.h"
 
+#define FC_FIX_3D_DATA
+
```
advaitjain, line 32: is this really necessary? can we simply change the default without introducing an ifdef?

advaitjain (changes requested):
one clarifying question and then this PR should be ready to go.

tensorflow/lite/micro/kernels/xtensa/fully_connected.cc
vp-cad: Yes, this #define is not necessary and the change can be default behavior.

advaitjain (approved):

advaitjain (approved):

PR #1558, opened 2022-11-15
Increase arena size in network example

mansnils:
BUG=Tensor arena is too small in network example


advaitjain (approved):

advaitjain:
@mergifyio refresh

PR #1568, opened 2022-11-16
Provide access to profiler events without logging them.

afoxley:
Bug=None


afoxley:
Not needed now that we can use custom profiler

Issue #1571, opened 2022-11-17
armclang does not support option -Wno-stringop-overread


annietllnd:
Hi, the changed added for #1544 omits the Arm(R) Cortex(R)-M-based builds, which adds the flag and as a result introduces an error it's not valid for armclang. I'm going to submit a PR which adds the Cortex-M builds to to list which is exempted from the added flag. 

For reference, this topic is also mentioned in #1337 

Thanks,
Annie

t4i5m6:
Hi, I met this problem too when I tried to make test_hello_world_test  with my Macbook with M2 chip.

error: unknown warning option '-Wno-stringop-overread' [-Werror,-Wunknown-warning-option]
make: *** [tensorflow/lite/micro/tools/make/Makefile:651: gen/osx_arm64_default/obj/core/tensorflow/lite/micro/examples/hello_world/hello_world_test.o] Error 1

Thanks,
Tim

PR #1572, opened 2022-11-17
exempt Arm(R) Cortex(R)-M-based builds from string overread warning

annietllnd:
BUG=#1571

Using the [filter](https://www.gnu.org/software/make/manual/html_node/Text-Functions.html) function instead, we are able to filter out all builds where this issue has been seen


advaitjain (approved):

advaitjain:
Just confirming, is this still needed after https://github.com/tensorflow/tflite-micro/pull/1550?

advaitjain:
> Just confirming, is this still needed after #1550?

What I should have asked is that if we remove `Wno-stringop-overread` are we ok? Since #1550 fixes the warning, we can go ahead and revert #1544

annietllnd:
Hi Advait, yes that's also an option. If you're okay with reverting #1544, then let's do that. Thanks!

advaitjain:
> Hi Advait, yes that's also an option. If you're okay with reverting #1544, then let's do that. Thanks!

Reverting with https://github.com/tensorflow/tflite-micro/pull/1577

PR #1575, opened 2022-11-18
Set a default last_used value for micro_allocation_info

vamsimanchala:
In the TFLM memory planning there is an implicit assumption currently that all tensors (other than the model inputs and outputs) must be both inputs and outputs of an OP.

In the case of the model from the bug, the else subgraph of the IF doesn't have any OPs which is breaking that assumption and hence the buffer for that tensor was not getting planned.

This fix will ensure that tensors that are outputs from an OP but not inputs to any other OP also have a reasonable lifetime.

BUG=b/257084942


advaitjain (approved):
Can you change the bug to be a hyperlink?

vamsimanchala:
> Can you change the bug to be a hyperlink?

Done.

PR #1576, opened 2022-11-18
Forcing reference kernel usage for TFLite interpreter during testing 

rewu93:
BUG=http://b/259572815

Currently, the default kernel usage of the TFlite interpreter is optimized ones. This creates test inconsistencies when comparing TFLM results (reference kernel) with TFL's. Inference results comparison is logically valid if both are using reference kernels.


tensorflow/lite/micro/python/interpreter/tests/interpreter_test.py
```diff
@@ -143,7 +143,10 @@ def testCompareWithTFLite(self):
     tflm_interpreter = tflm_runtime.Interpreter.from_bytes(model_data)
 
     # TFLite interpreter
-    tflite_interpreter = tf.lite.Interpreter(model_content=model_data)
+    tflite_interpreter = tf.lite.Interpreter(
```
paulinesho, line 146: Does removing tolerance in line 175 now work?

rewu93: Yes. Removed. Again the real fix will be #1463

tensorflow/lite/micro/examples/hello_world/train/train_hello_world_model.py
```diff
@@ -503,7 +503,9 @@ def predict_tflite(tflite_model: bytes, x_test: np.ndarray) -> np.ndarray:
   x_test_ = x_test_.astype(np.float32)
 
   # Initialize the TFLite interpreter
-  interpreter = tf.lite.Interpreter(model_content=tflite_model)
+  interpreter = tf.lite.Interpreter(model_content=tflite_model,
+                                    experimental_op_resolver_type=tf.lite.
+                                    experimental.OpResolverType.BUILTIN_REF)
```
advaitjain, line 508: the formatting is odd (yapf does that sometimes).

Let's try adding a \:
```py
  interpreter = tf.lite.Interpreter(model_content=tflite_model,
                                    experimental_op_resolver_type=\
                                        tf.lite.experimental.OpResolverType.BUILTIN_REF)
```

advaitjain: and if that works, perhaps make the same change in the other places as well?

rewu93: Done

advaitjain (approved):

PR #1580, opened 2022-11-22
Fix string overread warning with gcc-12.

advaitjain:
With #1550, we are still seeing the string overread warning with gcc-12. Manually tested with gcc-12 that the build fails without this fix and passes with the fix from this PR.

BUG=http://b/258039184, #1337 and #1571


rewu93 (approved):

advaitjain:
@mansnils: I hope this works for you.

PR #1583, opened 2022-11-22
remove unnecessary apt install of numpy that appear to be causing http://b/260016282

advaitjain:
Manually tested by merging this change in my fork of tflite-micro and confirmed that the bazel job goes past the install stage with this change (and fails without it).

We will need to merge this change before it can have an effect on the tflite-micro repo and that will be done by ignoring the branch protection rules (administrator only).

Also, removed some extra trailing whitespaces.

BUG=http://b/260016282


advaitjain:
This is blocking everything so I am going to merge without waiting for review.

PR #1584, opened 2022-11-22
Update to Renode 1.13.2

advaitjain:
With help from @PiotrZierhoffer, this PR updates TFLM to use the Renode release from Oct 3 (renode-1.13.2.linux-portable).

Manually confirmed that I can run the following command locally (http://b/257332395):
```
make -f tensorflow/lite/micro/tools/make/Makefile TARGET=bluepill test -j8
```

BUG=http://b/260027053 and http://b/257332395

PR #1586, opened 2022-11-22
Run renode from within docker container

advaitjain:
Workaround to keep renode working as part of CI without needing to upgrade(which is the subject of #1584).

BUG=http://b/260027053


advaitjain:
Merging without branch protection because we need to merge this PR and then we can have all tests pass for #1580

PR #1598, opened 2022-11-29
Enable TAP for python interpreter test

rewu93:
BUG=http://b/247808903

Enable TAP for python interpreter test. Previous decision on disabling is due to the randomness of model parameters introduced by the training process. This issue will be addressed by b/248061370. This PR only relaxes the tolerance for test coverage considerations. 

This PR reverts https://github.com/tensorflow/tflite-micro/pull/1436.

### DIFF
--- a/tensorflow/lite/micro/python/interpreter/tests/interpreter_test.py
+++ b/tensorflow/lite/micro/python/interpreter/tests/interpreter_test.py
@@ -59,15 +59,13 @@ class ConvModelTests(test_util.TensorFlowTestCase):
                      1)
     self.assertEqual(
         input_details["quantization_parameters"]["quantized_dimension"], 0)
-    # TODO(b/248061370): use assertEqual after having fixed flatbuffer
+    # TODO(b/247808903): use assertEqual after having fixed flatbuffer
     self.assertAlmostEqual(
-        input_details["quantization_parameters"]["scales"][0],
-        0.003,
-        delta=0.1)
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/python/interpreter/tests/interpreter_test.py
```diff
@@ -59,15 +59,13 @@ def testInput(self):
...
     self.assertAlmostEqual(
-        input_details["quantization_parameters"]["scales"][0],
-        0.003,
-        delta=0.1)
+        input_details["quantization_parameters"]["scales"][0], 0.003, delta=1)
     self.assertAlmostEqual(
         input_details["quantization_parameters"]["zero_points"][0],
         -128,
-        delta=100)
+        delta=256)
```
paulinesho, line 68: How valuable are these tests when the delta is set to something like 256? What if we check only the types here to make sure that all arrays are properly set up, and leave numbers test to b/247808903?

rewu93: Yes you are right! It is not testing anything since the ranges are 1 and 256. It is essentially doing what you said "check only the types here to make sure that all arrays are properly set up, and leave numbers test to b/247808903". I dont want to delete or comment out them since we need to add them back later. What do you think? I can make the comment more specific here

paulinesho (approved):

PR #1599, opened 2022-11-30
Allow applications to register custom PAD kernels

davidgao:
This allows models with unsupported data types to run.

BUG=http://b/260762559


tensorflow/lite/micro/kernels/pad.h
```diff
@@ -0,0 +1,34 @@
...
+#define TENSORFLOW_LITE_MICRO_KERNELS_PAD_H_
+
+#include "third_party/tensorflow/lite/c/common.h"
+
+namespace tflite {
+namespace ops {
+namespace micro {
+namespace pad {
+
+TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node);
```
advaitjain, line 26: nit: can you make this a flat tflite namepspace (http://b/170230530)?

The function would have to become `PadPrepare`, similar to https://github.com/tensorflow/tflite-micro/blob/b4279b999d34ed71a57f53ce47162f8a1f7f133b/tensorflow/lite/micro/kernels/conv.h#L77

advaitjain (changes requested):
one small nit comment.

tensorflow/lite/micro/kernels/pad.h
davidgao: Done, cleaned up the xtensa version as well to keep them aligned.

advaitjain (approved):

PR #1601, opened 2022-11-30
Update generate micro mutable script

mansnils:
* Support multiple models as input and update documentation to reflect this.
* Resolve broken Makefile generate model test.
* Handle custom operators correctly.

Co-authored-by: Christoffer Krabbe christoffer.krabbe@arm.com

BUG=generate micro mutable script does not support multiple models and handle custom operators correctly


advaitjain (approved):
Approving with the following comments:

We would like to move to an end to end flow where the custom Op resolver is generated from the TFLM Python bindings and that should hopefully remove the need to parse the header files manally to detect whether an Op is supported in TFLM or not.

No clear timeline for such a change and in the interim, having this script is good with the understanding that it is and early prototype and subject to change.

PR #1607, opened 2022-12-02
Remove HIFI4_INTERNAL from the code completely.

tanmaydas82:
BUG=https://b.corp.google.com/issues/239852051

This PR when land on google3, will not fail any CI build (this is a experimental CL to support this : https://critique.corp.google.com/cl/492504595)

### DIFF
--- a/tensorflow/lite/micro/tools/make/ext_libs/xtensa.inc
+++ b/tensorflow/lite/micro/tools/make/ext_libs/xtensa.inc
@@ -56,7 +56,7 @@ ifeq ($(TARGET_ARCH), hifi5)
     -I$(NNLIB_PATH)/algo/common/include/ \
     -I$(NNLIB_PATH)/algo/ndsp/hifi5/include/
 
-else ifeq ($(TARGET_ARCH), $(filter $(TARGET_ARCH), hifi4 hifi4_internal))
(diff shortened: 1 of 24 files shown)


tensorflow/lite/micro/tools/make/ext_libs/xtensa.inc
```diff
@@ -56,7 +56,7 @@ ifeq ($(TARGET_ARCH), hifi5)
     -I$(NNLIB_PATH)/algo/common/include/ \
     -I$(NNLIB_PATH)/algo/ndsp/hifi5/include/
 
-else ifeq ($(TARGET_ARCH), $(filter $(TARGET_ARCH), hifi4 hifi4_internal))
+else ifeq ($(TARGET_ARCH), $(filter $(TARGET_ARCH), hifi4))
```
advaitjain, line 59: we can remove the filter. it can be `ifeq ($(TARGET_ARCH), hifi4)`

advaitjain (changes requested):
thanks! one nit change and then we're ready to go.

tensorflow/lite/micro/tools/make/ext_libs/xtensa.inc
tanmaydas82: Done. Thank you for the suggestion!

advaitjain (approved):

PR #1614, opened 2022-12-08
clanup lifetimes of unused tensors

vamsimanchala:
This PR-
1) Adds additional testing to lock-down the behavior of how the tensor lifetimes are marked in AllocationInfo struct when the if-then subgraph is empty and when the subgraph output tensors are unused. 
2) Fixes a corner case where the subgraph outputs can remain unused. This will mark the tensor lifetime of these tensors accurately.

BUG=http://b/257084942


tensorflow/lite/micro/micro_allocation_info.cc
```diff
@@ -245,6 +241,9 @@ TfLiteStatus AllocationInfoBuilder::MarkAllocationLifetimes(
     const int tensor_index = subgraph->inputs()->Get(i);
     AllocationInfo* current = &subgraph_allocation_info[tensor_index];
     UpdateFirstCreated(current, allocation_scope_count_);
+    // This will ensure that tensors that are outputs from an
+    // OP but not inputs to any other OP also have a reasonable lifetime.
```
advaitjain, line 245: is this comment accurate? I would expect to say something along the lines of tensors that are inputs to the subgraphs but not used in any ops (see b/<> for an example of when this can happen).

tensorflow/lite/micro/micro_allocation_info.cc
```diff
@@ -316,6 +315,11 @@ TfLiteStatus AllocationInfoBuilder::MarkAllocationLifetimes(
        subgraph->outputs() != nullptr && i < subgraph->outputs()->size(); ++i) {
     const int tensor_index = subgraph->outputs()->Get(i);
     AllocationInfo* current = &subgraph_allocation_info[tensor_index];
+    // Make sure to assign the First created value of the subgraph output
+    // This will handle the case where the subgraph is empty or the output
+    // tensors are unused by the subgraph. This is a sanity measure to
```
advaitjain, line 320: output tensors unused by the subgraph is not clear to me. Let's remove that comment.

tensorflow/lite/micro/micro_allocation_info.cc
```diff
@@ -316,6 +315,11 @@ TfLiteStatus AllocationInfoBuilder::MarkAllocationLifetimes(
...
+    // make sure the memory planner gets clean data to work with.
```
advaitjain, line 321: nit: let's change

```
// This is a sanity measure to make sure the memory planner gets clean data to work with.
```

to
```
// This helps ensure all tensors have valid lifetimes before those are used by the memory planner.
```

advaitjain (changes requested):
looks good to me, just some small nits.

tensorflow/lite/micro/micro_allocation_info.cc
vamsimanchala: Done.

tensorflow/lite/micro/micro_allocation_info.cc
vamsimanchala: Done.

advaitjain (approved):

PR #1616, opened 2022-12-13
Sync from upstream TF and cherrypick TFLM changes from http://cl/493746273

advaitjain:
With this change, the sync between TF and TFLM should be fixed. However, the new test cases from http://cl/493746273 still need to be ported to TFLM.

BUG=http://b/260374880
NO_CHECK_TFLITE_FILES=manual sync from upstream


advaitjain:
This PR is ready for review

paulinesho (approved):

PR #1618, opened 2022-12-14
Optimized kernel reference fallbacks: FULLY_CONNECTED

ddavis-2015:
@tensorflow/micro

Remove conditional compilation for kernel tests for the following platforms: Xtensa, CEVA
Add optimized kernel reference fallback code for the following platforms: Xtensa

bug=fixes Add reference fallbacks to optimized kernels: FULLY_CONNECTED #1624


tensorflow/lite/micro/kernels/xtensa/fully_connected.cc
```diff
@@ -37,18 +37,27 @@ TfLiteStatus CalculateOpData(TfLiteContext* context,
...
+    // https://github.com/tensorflow/tensorflow/issues/44912 for additional
+    // context.
+    TFLITE_DCHECK(filter->params.zero_point == 0);
+    data->filter_zero_point = filter->params.zero_point;
+    data->output_zero_point = output->params.zero_point;
+
+    return CalculateActivationRangeQuantized(context, activation, output,
+                                             &data->output_activation_min,
+                                             &data->output_activation_max);
+  }
```
advaitjain, line 58: can we directly call https://github.com/tensorflow/tflite-micro/blob/b9b0d6343f75e3839b9b244632b4bf4b18dd56a8/tensorflow/lite/micro/kernels/fully_connected.h#L66-L69 instead of duplicating the code?

tensorflow/lite/micro/kernels/xtensa/fully_connected.cc
```diff
@@ -211,23 +214,85 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
+      MicroPrintf("Input type %s (%d) not supported.",
+                  TfLiteTypeGetName(input->type), input->type);
+      return kTfLiteError;
+    }
   }
+
+  return kTfLiteOk;
 }
 
 }  // namespace
```
advaitjain, line 298: let's also add a kernel variant for only the INT8 similar to https://github.com/tensorflow/tflite-micro/blob/b9b0d6343f75e3839b9b244632b4bf4b18dd56a8/tensorflow/lite/micro/kernels/fully_connected.h#L76-L79

This will ensure that any binary that registers only INT8 (such as https://github.com/tensorflow/tflite-micro/blob/b9b0d6343f75e3839b9b244632b4bf4b18dd56a8/tensorflow/lite/micro/benchmarks/keyword_benchmark.cc#L56) will not see any size increase for xtensa.

And it would be good to verify that the size for keyword_benchmark is unchanged with this PR.

advaitjain (changes requested):

PR #1619, opened 2022-12-14
Arena usage, input data and output data is published by micro_benchmark.

tanmaydas82:
BUG=https://b.corp.google.com/issues/262609701


tanmaydas82:
This change is not required as we are using the tflm recording_micro_interpreter in place for seanet_benchmark_runner.

PR #1642, opened 2023-01-04
Flatten kernel namespace: PAD

### LINKED ISSUE #1625: Flatten kernel namespace: PAD
@tensorflow/micro

Flatten the C++ namespace of all kernel operators as per abseil / Tip of the Week #130: Namespace Naming

This issue pertains to: PAD

ddavis-2015:
@tensorflow/micro 

Flatten namespace for kernel operator PAD

Move Prepare inside of anonymous namespace

bug=fixes #1625


advaitjain:
Pad was flattened with https://github.com/tensorflow/tflite-micro/pull/1599 (see https://github.com/tensorflow/tflite-micro/pull/1599#discussion_r1035412043) and the reason for having PadPrepare not in the anonymous namespace is that it is declared in pad.h and used externally. We don't need any changes to the Pad kernel.

ddavis-2015:
PadPrepare is used internally.

PR #1645, opened 2023-01-04
Use xtensa docker container version 0.3

tanmaydas82:
BUG=https://b.corp.google.com/issues/263509675


advaitjain (approved):
Approving and then merging without waiting for the CI because the changes from this PR need to be in the main branch for them to be used.

The expected behavior is that once this PR is merged, https://github.com/tensorflow/tflite-micro/pull/1638 will pass all the branch protection rules and we will be good to go.

There is always a chance that something is different between testing locally and as part of CI so we will monitor.

PR #1648, opened 2023-01-05
Conslidate MicroProfiler events before clearing

paulinesho:
Currently, GetTotalTicks() and LogTicksPerTagCsv() are both returning incorrect numbers. This is only evident when many iterations are run. The issue is that every call to BeginEvent() adds a new unique entry, and when the number of entries exceeds kMaxEvents, the existing events are simply cleared by setting num_events_ = 0. This means that GetTotalTicks() and LogTicksPerTagCsv() are not reporting those events that are previously cleared without any caching or consolidation.

This PR moves the current events into `total_ticks_per_tag` which keeps track of total ticks of each unique tag before clearing them. As a result, we don't need to call ClearEvents() before each iteration and keep track of the ticks reported by GetTotalTicks() every iteration. Instead, GetTotalTicks() can be called at the end after all iterations to return all ticks. LogTicksPerTagCsv() is also now reporting all events, not just events that haven't been cleared by BeginEvent().

BUG=b/264555999


paulinesho:
Closing in favor of https://github.com/tensorflow/tflite-micro/pull/1835

PR #1668, opened 2023-01-18
build: add means of linking with C libs in Py packages

rkuester:
Extend the Python repository_rule used to create external repositories, adding targets for C-language binary libraries shipped inside Python packages; e.g., that shipped in package tensorflow-gpu.

These targets are to be used as dependencies by C-language targets.

Note: when debugging the build, it can be helpful to examine the repository directory and BUILD file this repository_rule generates in the bazel cache.

Begin using a python/ directory at the root of the project for code that is specific to Python.

Upgrade to the latest version of rules_python first. Note that the unit test to keep requirements.in and requirements.txt is disabled (specifically with 
ec6bdc5d4443285d28a44076f06c203d9582e4a1)

Add a unit test for this feature.

BUG=see description


WORKSPACE
```diff
@@ -34,13 +34,10 @@ http_archive(
...
-
-# Create a central external repo that contains Bazel targets for all the
-# third-party packages specified in the requirements.txt file.
-pip_install(
-   name = "tflm_pip_deps",
-   requirements = "//third_party:python_requirements.txt",
+load("@rules_python//python:pip.bzl", "pip_parse")
+pip_parse(
+    name = "tflm_pip_deps",
+    requirements_lock = "@//third_party:python_requirements.txt",
```
advaitjain, line 40: remove the leading @?

python/py_pkg_cc_deps.bzl
```diff
@@ -83,30 +83,29 @@ repository named `tflm_pip_deps` via @rules_python in the WORKSPACE:
     )
 ```
 
-2. Use the repository rule `tflm_py_cc_headers` in the WORKSPACE to create an
-external repository with a target `@numpy_headers:numpy_headers`, passing the
-py_library target from @tflm_pip_deps obtained via requirement() and an
+2. Use the repository rule `py_pkg_cc_deps` in the WORKSPACE to create an
+external repository with a target `@numpy_cc_deps//:cc_headers`, passing the
```
advaitjain, line 135: Does this work `@numpy_cc_deps:cc_headers`?

python/py_pkg_cc_deps.bzl
```diff
@@ -83,30 +83,29 @@ repository named `tflm_pip_deps` via @rules_python in the WORKSPACE:
...
 a dependency to a rule that needs the headers, e.g., the cc_library()-based
-pybind_library(). Note the target can be spelled as the default target of the
-repository's top-level package:
+pybind_library():
 ```
     pybind_library(
         name = "your_extension_lib",
         srcs = [...],
-        deps = ["@numpy_headers", ...],
+        deps = ["@numpy_cc_deps//:cc_headers", ...],
```
advaitjain, line 156: maybe
```BUILD
@numpy_cc_deps:cc_headers
```

tensorflow/lite/micro/python/interpreter/src/BUILD
```diff
@@ -27,7 +27,7 @@ pybind_library(
         "//tensorflow/lite/micro:micro_framework",
         "//tensorflow/lite/micro:op_resolvers",
         "//tensorflow/lite/micro:recording_allocators",
-        "@numpy_headers",
+        "@numpy_cc_deps//:cc_headers",
```
advaitjain, line 30: maybe
```BUILD
@numpy_cc_deps:cc_headers
```

python/tests/BUILD
```diff
@@ -0,0 +1,13 @@
...
+    name = "cc_deps_link_test",
+    size = "small",
+    srcs = [
+        "cc_deps_link_test.cc",
+    ],
+    copts = [
+        "-std=c++14",
+    ],
+    deps = [
+        "@tensorflow_cc_deps//:cc_library",
```
advaitjain, line 11: maybe `"@tensorflow_cc_deps:cc_library"`?

python/tests/BUILD
```diff
@@ -0,0 +1,13 @@
+cc_test(
+    name = "cc_deps_link_test",
+    size = "small",
+    srcs = [
+        "cc_deps_link_test.cc",
+    ],
+    copts = [
+        "-std=c++14",
+    ],
```
advaitjain, line 9: let's make a bug and add a `TODO(b/)` to see if this needs to be made more accessible to other targets in the repo. And give reason for why it is needed.

This may be useful as we think about generalizing:
https://github.com/tensorflow/tflite-micro/blob/836a4ad1bc14d94ba10ac289ffe5fb797d8d0fbe/tensorflow/lite/micro/build_def.bzl#L1-L7

python/tests/BUILD
rkuester, line 11: Is "//" necessary?

python/py_pkg_cc_deps.bzl
```diff
@@ -1,100 +1,148 @@
...
-    # rules_python unpacks the Python package within the site-packages directory.
-    path = path.get_child("site-packages")
+    return result
 
-    # Append the include_prefix, if any. get_child() needs one component at a
-    # time.
-    if include_prefix:
-        for c in include_prefix.split("/"):
-            path = path.get_child(c)
+def _make_build_file(basedir, include_paths, libs):
```
advaitjain, line 34: let's add to the PR description a brief set of steps to inspect the BUILD file that gets created for tensorflow-cpu as a result of this rule.

advaitjain (changes requested):
small nit comments. looks really good!

WORKSPACE
rkuester: Yes, fixed such redundancies throughout WORKSPACE in 5db747e.

python/py_pkg_cc_deps.bzl
rkuester: No, the `//` is required when referring an external repository in a label—per the [Bazel docs on labels](https://bazel.build/concepts/labels) and as verified by trying it out.

tensorflow/lite/micro/python/interpreter/src/BUILD
rkuester: No, the `//` is required when referring an external repository in a label—per the [Bazel docs on labels](https://bazel.build/concepts/labels) and as verified by trying it out.

python/tests/BUILD
rkuester: Added in 470fffb.

python/py_pkg_cc_deps.bzl
rkuester: Added in the PR description and in in-line comments in the implementation in 8f2073a.

Rather than give steps, I suggested finding and examining the repository directory created in the cache when debugging the build or implementation. Giving steps would have been awkward considering the context, and probably unnecessarily detailed.

python/tests/BUILD
rkuester: Yes, the // is required when referring an external repository in a label—per the [Bazel docs on labels](https://bazel.build/concepts/labels) and as verified by trying it out.

advaitjain (dismissed):

advaitjain:
I have updated the PR description to call out that test for requirements.in and requirements.txt consistency is currently disabled. Ideally we should have a bug for that with a corresponding TODO(b/) but it's not worth delaying this PR for that.

advaitjain (approved):

PR #1671, opened 2023-01-19
Disable lstm_eval_test with asan and ubsan.

advaitjain:
Manually made this change in http://cl/502285636

BUG=http://b/266028903


tanmaydas82 (approved):
LGTM

PR #1676, opened 2023-01-20
Add support for offline planning of variable tensors

mansnils:
* New assumption that offline planner (in this case Vela) supports variable tensors
 * Move variable tensor allocation in the online planner to later in the memory plannign phase so that we have the opportunity to get the offfsets from the offline planner for these as well.

Co-authored-by: Fredrik Svedberg fredrik.svedberg@arm.com

BUG=Offline planner does not support variable tensors.


tensorflow/lite/micro/micro_allocation_info.cc
```diff
@@ -204,6 +204,10 @@ TfLiteStatus AllocationInfoBuilder::InitializeAllocationInfo(
           (current->bytes != 0);
       if (offline_offsets) {
         current->offline_offset = offline_offsets[i];
+        if (subgraph->tensors()->Get(i)->is_variable() &&
```
advaitjain, line 207: can we add a comment explaining this? it seems subtle.

tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -819,6 +814,16 @@ TfLiteStatus MicroAllocator::CommitStaticMemoryPlan(
   const int32_t* offline_planner_offsets = nullptr;
   TF_LITE_ENSURE_STATUS(
       builder.GetOfflinePlannedOffsets(&offline_planner_offsets));
+
+  // Allocate buffers for variable tensors.
```
advaitjain, line 818: let's add a comment about why the variable allocation needs to be here. (i.e. something about its interaction with the offline planner)

advaitjain (changes requested):

tensorflow/lite/micro/micro_allocation_info.cc
mansnils: We need to set needs_allocating here only for offline planned variable tensors and not for all tensors. Added a comment.

tensorflow/lite/micro/micro_allocator.h
```diff
@@ -248,8 +248,9 @@ class MicroAllocator {
   virtual TfLiteStatus AllocateTfLiteEvalTensors(
       const Model* model, SubgraphAllocations* subgraph_allocations);
   // Allocates persistent tensor buffers for variable tensors in the subgraph.
-  virtual TfLiteStatus AllocateVariables(const SubGraph* subgraph,
-                                         TfLiteEvalTensor* eval_tensors);
+  virtual TfLiteStatus AllocateVariables(
+      const SubGraph* subgraph, TfLiteEvalTensor* eval_tensors,
+      const int32_t* offline_planner_offsets = nullptr);
```
advaitjain, line 253: is this default argument needed? doesn't seem like it.

tensorflow/lite/micro/micro_allocator.h
```diff
@@ -248,8 +248,9 @@ class MicroAllocator {
   virtual TfLiteStatus AllocateTfLiteEvalTensors(
       const Model* model, SubgraphAllocations* subgraph_allocations);
   // Allocates persistent tensor buffers for variable tensors in the subgraph.
-  virtual TfLiteStatus AllocateVariables(const SubGraph* subgraph,
-                                         TfLiteEvalTensor* eval_tensors);
+  virtual TfLiteStatus AllocateVariables(
```
advaitjain, line 251: can you add a comment about what the offline_planner_offsets parameter is for?

tensorflow/lite/micro/recording_micro_allocator.h
```diff
@@ -77,8 +77,9 @@ class RecordingMicroAllocator : public MicroAllocator {
       const Model* model, SubgraphAllocations* subgraph_allocations) override;
   TfLiteStatus AllocateTfLiteEvalTensors(
       const Model* model, SubgraphAllocations* subgraph_allocations) override;
-  TfLiteStatus AllocateVariables(const SubGraph* subgraph,
-                                 TfLiteEvalTensor* eval_tensors) override;
+  TfLiteStatus AllocateVariables(
+      const SubGraph* subgraph, TfLiteEvalTensor* eval_tensors,
+      const int32_t* offline_planner_offsets = nullptr) override;
```
advaitjain, line 82: same comment here. let's remove the default nullptr parameter and instead change RecordingMicroInterpreter to use the updated API.

advaitjain (changes requested):
sorry missed a few small things in my last review around not having default arguments.

tensorflow/lite/micro/micro_allocator.h
mansnils: no removing it

tensorflow/lite/micro/micro_allocator.h
mansnils: done

tensorflow/lite/micro/recording_micro_allocator.h
mansnils: Done

mansnils:
All comments resolved and ready for review.

advaitjain (approved):

advaitjain (approved):

PR #1682, opened 2023-01-25
Redesigned reference LSTM kernels

rewu93:
BUG=http://b/266447344

New design for the reference LSTM kernels. 
* Support float, int8 activation int8 weight and int16 activation int8 weight quantized unidirectional LSTM models.  Use template and function overloading to remove code duplications. 
* Use the existing reference kernel implementation (fully connected and mul) inside the LSTM kernel, providing flexibility for optimizations. 
* Readable and expandable unit tests 

Note:Only the standard LSTM layer is supported (defined in https://keras.io/api/layers/recurrent_layers/lstm/)


advaitjain (approved):
Approving based on the design review and discussions. We should expect to make changes incrementally once this is merged for style, clarity, bug fixes etc. as and when we encounter issues.

PR #1693, opened 2023-01-30
Fix hedron_compile_commands fetch with updated sha256

suleshahid:
hedron_compile_commands is causing build to fail due to fetch error with wrong sha256. Using expected sha256 fixes error.

More info to follow in bug for root cause.
BUG=[267220031](https://b.corp.google.com/issues/267220031)


suleshahid:
Was a github wide issue, they are reverting back to old checksums.

PR #1694, opened 2023-01-30
hello world python script and test.

tanmaydas82:
This PR contains the following changes 
 
* Python scripts to evaluate and test
* Changed model to be floating point (was int8 prior to this PR)

We can build and run the `evaluate.py` script by

```
bazel build :evaluate
bazel run :evaluate

# If you want to run the tflite interpreter instead of tflm interpreter
bazel run :evaluate -- --use_tflite
```

We can run the `evaluate_test.py` test using the following 
```
bazel build :evaluate_test
bazel run :evaluate_test
```
Please look at the bug below for the future plans about hello_world.
BUG=https://b.corp.google.com/issues/267222141


tensorflow/lite/micro/examples/hello_world/evaluate.py
```diff
@@ -0,0 +1,142 @@
...
+import os
+from absl import app
+from absl import flags
+import numpy as np
+import matplotlib.pyplot as plt
+from tensorflow.python.platform import resource_loader
+from tflite_micro.tensorflow.lite.micro.python.interpreter.src import \
+  tflm_runtime
+
+_MODEL_PATH = flags.DEFINE_string(
```
paulinesho, line 24: Can we make this default to `hello_world.tflite`? More often than not that's the only model that will be used with this script.

Will resume reviewing after changes suggested by the team are in. It's probably a good idea as well to separate py and cc files into two subdirectories since there're already a lot of files in `.../hello_world/`. We can chat more about it next time.

tanmaydas82: Thank you for your feedback here and in the meeting. I have updated the code according to the notes from the meeting. PTAL and let me know your thoughts. We can discuss about moving the python and cc file into two subdirectories but I think the cc files will be removed once the python ones are landed.

tensorflow/lite/micro/examples/hello_world/evaluate.py
```diff
@@ -0,0 +1,142 @@
...
+def get_quantization_params(interpreter_io_details):
+  quantize_params = interpreter_io_details.get('quantization_parameters')
+  scale = quantize_params.get('scales')
+  zero_point = quantize_params.get('zero_points')
+  return scale, zero_point
+
+
+# Invoke the tflm interpreter with x_values in the range of [0, 2*PI] and
+# returns the prediction of the interpreter.
+def get_tflm_prediction(hello_world_model_path, x_values):
```
paulinesho, line 75: Just `model_path` should be sufficient.

tensorflow/lite/micro/examples/hello_world/evaluate.py
```diff
@@ -0,0 +1,142 @@
...
+
+  y_predictions = np.empty(x_values.size, dtype=np.float32)
+  i = 0
+  for x_value in x_values:
+    y_predictions[i] = invoke_tflm_interpreter(input_shape,
+                                               tflm_interpreter,
+                                               x_value,
+                                               input_index=0,
+                                               output_index=0)
+    i += 1
```
paulinesho, line 89: Use `enumerate()` here instead. [go/python-tips/023](https://goto.google.com/python-tips/023)

tensorflow/lite/micro/examples/hello_world/evaluate.py
```diff
@@ -0,0 +1,142 @@
...
+def main(_):
+  hello_world_no_quant_model_path = os.path.join(PREFIX_PATH,
+                                                 'hello_world.tflite')
+
+  x_values = generate_random_input()
+
+  # Calculate the corresponding sine values
+  y_true_values = np.sin(x_values).astype(np.float32)
+
+  if _USE_TFLITE_INTERPRETER.value:
```
paulinesho, line 127: Do we want this to be an addition instead of substitution? i.e. if `use_tflite` is set, run both TFL and TFLM, instead of just TFL.

tensorflow/lite/micro/examples/hello_world/evaluate.py
```diff
@@ -0,0 +1,142 @@
...
+  x_values = np.random.uniform(low=0, high=2 * np.pi,
+                               size=sample_count).astype(np.float32)
+  # Shuffle the values to guarantee they're not in order
+  np.random.shuffle(x_values)
+  return x_values
+
+
+# Get the metadata like scales and zero_points from the interpreter input/output
+# details.
+def get_quantization_params(interpreter_io_details):
```
paulinesho, line 66: I still think we should place this function in the scope where it's needed, i.e. `evaluate_test.py`. Since we don't use any quantization in this file, it might just cause confusion.

tensorflow/lite/micro/examples/hello_world/evaluate_test.py
```diff
@@ -0,0 +1,98 @@
...
+  model_path = os.path.join(PREFIX_PATH, 'hello_world.tflite')
+  input_shape = (1, 1)
+  output_shape = (1, 1)
+  # Create the tflm interpreter
+  tflm_interpreter = tflm_runtime.Interpreter.from_file(model_path)
+
+  def test_init_error_handling(self):
+    with self.assertRaisesWithPredicateMatch(ValueError,
+                                             'Invalid model file path'):
+      tflm_runtime.Interpreter.from_file('wrong.tflite')
```
paulinesho, line 38: I'm not sure if this is necessary here. This is testing the functionality of the TFLM interpreter itself, not the hello_world example.

tensorflow/lite/micro/examples/hello_world/train/train_hello_world_model.py
```diff
@@ -468,8 +468,6 @@ def representative_dataset():
   converter.optimizations = [tf.lite.Optimize.DEFAULT]
   # Enforce integer only quantization
   converter.target_spec.supported_ops = [tf.lite.OpsSet.TFLITE_BUILTINS_INT8]
-  converter.inference_input_type = tf.int8
-  converter.inference_output_type = tf.int8
```
paulinesho, line 472: Is the checked in hello_world.tflite the new output of conversion using this script?

tensorflow/lite/micro/examples/hello_world/evaluate.py
tanmaydas82: Done

tensorflow/lite/micro/examples/hello_world/evaluate.py
tanmaydas82: Thank you for the suggestion!

tensorflow/lite/micro/examples/hello_world/evaluate_test.py
tanmaydas82: You are right! This is copied from interpreter test, so removing it.

tensorflow/lite/micro/examples/hello_world/train/train_hello_world_model.py
tanmaydas82: Yes, the hello_world.tflite file that is part of the PR is the output of the this conversion.

tensorflow/lite/micro/examples/hello_world/evaluate.py
tanmaydas82: Done.

tensorflow/lite/micro/examples/hello_world/evaluate.py
```diff
@@ -0,0 +1,133 @@
...
+from tflite_micro.tensorflow.lite.micro.python.interpreter.src import tflm_runtime
+
+_USE_TFLITE_INTERPRETER = flags.DEFINE_bool(
+    'use_tflite',
+    False,
+    'Inference with the TF Lite interpreter instead of the TFLM interpreter',
+)
+
+x_range = 2 * np.pi
+inferences_per_cycle = 1000
```
advaitjain, line 31: these apprear to be unused

tensorflow/lite/micro/examples/hello_world/evaluate_test.py
```diff
@@ -0,0 +1,103 @@
...
+import numpy as np
+
+from tensorflow.python.framework import test_util
+from tensorflow.python.platform import resource_loader
+from tensorflow.python.platform import test
+from tflite_micro.tensorflow.lite.micro.python.interpreter.src import tflm_runtime
+from tflite_micro.tensorflow.lite.micro.examples.hello_world import evaluate
+
+PREFIX_PATH = resource_loader.get_path_to_datafile('')
+
```
advaitjain, line 25: leading underscore for internal module level constants

`_PREFIX_PATH`

https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations

advaitjain (changes requested):

tensorflow/lite/micro/examples/hello_world/evaluate.py
tanmaydas82: Removed. Thank you for pointing this out.

tensorflow/lite/micro/examples/hello_world/evaluate_test.py
tanmaydas82: Done.

advaitjain (approved):

PR #1703, opened 2023-02-01
splitting some tests for parallel speedups.

jwithers:
Have split the presubmit Cortex-M tests into bluepill and stm32f4 jobs as well as splitting the Makefile x86 presubmit into two jobs. This roughly halves the time for ci.yml to run (from around an hour to roughly 30 minutes). Unfortunately, it's all for naught right now because the xtensa hifimini test added to xtensa.xml last week is a pig that takes 45 minutes to an hour.

Note that these times are approximate because there is a 20%+ time variance depending on github's weather. Also note there are occasional pathological runtimes where a job can take twice as long as normal to run for no apparent reason. During testing this PR I had a bazel (presubmit) that took over an hour when the normal run time is ~32 minutes. 

Also note that CI on this is going to fail because we are changing some required test names. The following changes will need to be implemented in our branch protections job list:

`Makefile x86 (presubmit)` is now `Makefile x86 One (presubmit)` and `Makefile x86 Two (presubmit)`
`Cortex-M (presubmit)` is now `Cortex-M Bluepill (presubmit)` and `Cortex-M stm32f4 (presubmit)`

BUG=https://issuetracker.google.com/issues/266654467

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '3.10'
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           ref: ${{ inputs.trigger-sha }}
       - name: Install dependencies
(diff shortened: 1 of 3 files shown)


jwithers:
[List of test runtimes at bug.](https://issuetracker.google.com/issues/266654467)

jwithers:
Have created a new PR covering this issue: https://github.com/tensorflow/tflite-micro/pull/1746

PR #1714, opened 2023-02-06
Refactors unpacking outside of deptwise_conv and fully_connected reference kernel

turbotoribio:
-Refactors unpacking outside of deptwise_conv and fully_connected reference kernel for 4bit test case
-allows 4 bit filters to be used with optimized kernels in TfLite for depthwise_conv and fully_connected ops
-test for correctness by running fully_connected and depthwise_conv  in TfLite and TFLM

Corresponding g3 cl - [cl/504384976](https://critique.corp.google.com/cl/504384976)
BUG=[265045979](https://b.corp.google.com/issues/265045979)


advaitjain (approved):
approving since it is already reviewed internally and will fix the TF to TFLM sync.

PR #1724, opened 2023-02-08
Xtensa ci fix for vision p6 

turbotoribio:
Xtensa build is breaking due to failing int4 Conv test for vision p6 build. This is occuring due to changes to recent changes to Xtensa kernels where int4 kernel is now using the optimized kernels vs the reference kernels that were being used before for int4.

BUG=[268365158](https://b.corp.google.com/issues/268365158)


tensorflow/lite/micro/kernels/conv_test.cc
```diff
@@ -61,6 +61,10 @@ static TfLiteConvParams common_conv_params = {
 
 TF_LITE_MICRO_TESTS_BEGIN
 
+#if !defined(XTENSA)  // TODO(b/170321206): xtensa kernels are less general than
```
paulinesho, line 64: If only VISION_P6 is broken, ifdef VISION_P6 instead and attach the correct bug link.

turbotoribio: Got it !

paulinesho (approved):

PR #1728, opened 2023-02-09
Make Xtensa Vision P6 nightly build green

turbotoribio:
Xtensa build is breaking due to failing int4 depthwise_conv/fully_connected test for vision p6 build. This is occuring due to changes to recent changes to Xtensa kernels where int4 kernel is now using the optimized kernels vs the reference kernels that were being used before for int4.

This change disables the test so that the nighlty build is green again.

Root cause for the test failure will be investigated and fixed in a follow-up change.

Note: Previously the int4 tests were being (incorrectly) disabled for all Xtensa targets. This was an oversight and the current PR only disables the test for Vision P6.

BUG=[268365158](https://b.corp.google.com/issues/268365158)

### DIFF
--- a/tensorflow/lite/micro/kernels/depthwise_conv_test.cc
+++ b/tensorflow/lite/micro/kernels/depthwise_conv_test.cc
@@ -936,7 +936,7 @@ TF_LITE_MICRO_TEST(Int8Input32x1Filter32x1ShouldMatchGolden) {
                               kQuantizationTolerance, kTensorsSize, tensors));
 }
 
-#if !defined(VISION_P6) && !defined(XTENSA)
+#if !defined(VISION_P6)
 // TODO(b/268384678): xtensa vision p6 kernels break
 // this test, will if def till properly investigated.
 
--- a/tensorflow/lite/micro/kernels/depthwise_conv_test.cc
+++ b/tensorflow/lite/micro/kernels/depthwise_conv_test.cc
@@ -936,7 +936,7 @@ TF_LITE_MICRO_TEST(Int8Input32x1Filter32x1ShouldMatchGolden) {
                               kQuantizationTolerance, kTensorsSize, tensors));
 }
 
-#if !defined(VISION_P6) && !defined(XTENSA)  
+#if !defined(VISION_P6) && !defined(XTENSA)
 // TODO(b/268384678): xtensa vision p6 kernels break
 // this test, will if def till properly investigated.
 
--- a/tensorflow/lite/micro/kernels/depthwise_conv_test.cc
+++ b/tensorflow/lite/micro/kernels/depthwise_conv_test.cc
@@ -29,6 +29,20 @@ namespace {
 // DepthwiseConv.
 constexpr int kOutputTensorIndex = 3;
 
+// TODO(b/268384678): xtensa vision p6 kernels breaks int4 test
+// due to recent added optimized kernel support to xtensa for int4.
+// The corresponding test is disabled while 
+// investigation is being done. Corresponding variables
+// used only in that test have to be if def'd out
+// to avoid unused variable errors for vision p6.  
+
+#if !defined(VISION_P6)
+
+constexpr int kMaxFilterChannels = 64;
+constexpr int kMaxBiasChannels = 64;
+
+#endif  // !defined(VISION_P6)
+
 // Creates a DepthwiseConv opeerator, calls it with the provided input tensors
 // and some defaults parameters, and compares the output with
 // expected_output_data.
@@ -75,12 +89,12 @@ TfLiteStatus ValidateDepthwiseConvGoldens(
   return kTfLiteOk;
 }
 
-// TODO(b/268384678): xtensa vision p6 kernels break
-// this test, will if def till properly investigated.
+// TODO(b/268384678): xtensa vision p6 kernels breaks int4 test
+// due to recent added optimized kernel support to xtensa for int4.
+// The corresponding test is disabled while this is investegated in 
+// order for the vision p6 nightly build to be green. 
 #if !defined(VISION_P6)
 
-constexpr int kMaxFilterChannels = 64;
-constexpr int kMaxBiasChannels = 64;
 void TestDepthwiseConvQuantizedPerChannel(
     int* input_dims_data, const float* input_data, int8_t* input_quantized,
     float input_scale, int input_zero_point, int* filter_dims_data,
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/kernels/depthwise_conv_test.cc
```diff
@@ -983,4 +936,59 @@ TF_LITE_MICRO_TEST(Int8Input32x1Filter32x1ShouldMatchGolden) {
                               kQuantizationTolerance, kTensorsSize, tensors));
 }
 
+
+#if !defined(VISION_P6) && !defined(XTENSA)
```
advaitjain, line 939: should we only disable for P6 and leave for other XTENSA  targets?

advaitjain (changes requested):

tensorflow/lite/micro/kernels/depthwise_conv_test.cc
turbotoribio: xtensa was already disabled

advaitjain: yes, let's enable. I would think that removing the ifdef(XTENSA) was missed in the change when we enabled optimized kernels to also support int4.

turbotoribio: Done !

tensorflow/lite/micro/kernels/depthwise_conv_test.cc
```diff
@@ -79,34 +75,10 @@ TfLiteStatus ValidateDepthwiseConvGoldens(
...
-      CreateTensor(input_data, input_dims),
-      CreateTensor(filter_data, filter_dims),
-      CreateTensor(bias_data, bias_dims),
-      CreateTensor(output_data, output_dims),
-  };
-
-  ValidateDepthwiseConvGoldens(expected_output_data, output_dims_count,
-                               conv_params, 1e-5, tensors_size, tensors);
-}
+#if !defined(VISION_P6)
```
paulinesho, line 78: A little confused now. Is it necessary to hide `TestDepthwiseConvQuantizedPerChannel()` from VISION_P6 and `TestDepthwiseConvFloat()` from XTENSA? Are these existing issues or related to INT4 changes?

tensorflow/lite/micro/kernels/depthwise_conv_test.cc
```diff
@@ -175,6 +147,35 @@ void TestDepthwiseConvQuantizedPerChannel(
                                               output_dims_count, conv_params,
                                               1.0, tensors_size, tensors));
 }
+#endif  // !defined(VISION_P6)
+
+#if !defined(XTENSA)  // Needed to avoid build errors from unsused functions.
```
advaitjain, line 152: let's expand this comment:

```
// Xtensa kernels do not support float activations., and the corresponding tests are
// disabled. As a result, helper functions that are only needed for float kernel tests
// also need to be ifdef'd out to avoid build errors due to unused functions.
```

tensorflow/lite/micro/kernels/fully_connected_test.cc
```diff
@@ -44,7 +44,7 @@ const float simple_weights_data[] = {
 };
 // TODO(b/258710417): INT4 isn't currently supported on Hexagon.
 // TODO(b/265349042): INT4 isn't currently supported on Hifimini.
-#if !defined(HEXAGON) && !defined(HIFIMINI)
+#if !defined(HEXAGON) && !defined(HIFIMINI) && !defined(VISION_P6)
```
advaitjain, line 47: bug?

advaitjain (changes requested):

tensorflow/lite/micro/kernels/fully_connected_test.cc
turbotoribio: Done

tensorflow/lite/micro/kernels/depthwise_conv_test.cc
turbotoribio: got it !

tensorflow/lite/micro/kernels/depthwise_conv_test.cc
turbotoribio: Xtensa kernels do not support float inputs, due to this corresponding tests are disabled. So helper functions / variables have to be disabled in order to avoid unused function/variable build error.

tensorflow/lite/micro/kernels/depthwise_conv_test.cc
```diff
@@ -79,34 +75,12 @@ TfLiteStatus ValidateDepthwiseConvGoldens(
...
-      CreateTensor(filter_data, filter_dims),
-      CreateTensor(bias_data, bias_dims),
-      CreateTensor(output_data, output_dims),
-  };
-
-  ValidateDepthwiseConvGoldens(expected_output_data, output_dims_count,
-                               conv_params, 1e-5, tensors_size, tensors);
-}
+// TODO(b/268384678): xtensa vision p6 kernels break
+// this test, will if def till properly investigated.
```
advaitjain, line 79: this comment does not describe why the code has been ifdef'd. If you could update the comment to reflect the reason, that would be great.

tensorflow/lite/micro/kernels/depthwise_conv_test.cc
```diff
@@ -79,34 +75,12 @@ TfLiteStatus ValidateDepthwiseConvGoldens(
...
+#if !defined(VISION_P6)
 
+constexpr int kMaxFilterChannels = 64;
+constexpr int kMaxBiasChannels = 64;
```
advaitjain, line 83: globals should be at the top of the file (even if it means you need another ifdef. Let's move to the top, add an ifdef and add an appropriate comment for why the ifdef is needed.

advaitjain (changes requested):

tensorflow/lite/micro/kernels/depthwise_conv_test.cc
turbotoribio: Okay moved it to the top ! And updated corresponding comment. Is this descriptive enough on the why ?

tensorflow/lite/micro/kernels/depthwise_conv_test.cc
turbotoribio: Updated the comment , is this descriptive enough on the why ?

advaitjain (dismissed):

advaitjain (approved):

PR #1734, opened 2023-02-11
Remove the LSTM Internal implementation 

rewu93:
BUG=http://b/266447344

Benchmark using the int8 PTQ model from the MNIST LSTM example (https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/examples/mnist_lstm/trained_lstm_quant.tflite) 

image


tensorflow/lite/micro/kernels/lstm_eval_test.h
```diff
@@ -319,6 +322,14 @@ OpDataLSTM CreateLstmOpDataFloat(
 }
 
 /*Test Functions Below Here*/
+template <typename T>
```
advaitjain, line 325: not sure that this change needs to be part of the current PR, but let's go for it since there are a lot of moving pieces.

advaitjain (approved):

PR #1738, opened 2023-02-14
Deprecate parameter for number of resource variables

suleshahid:
Continued work on automating resource variables from python. After removing usage of the parameter, this PR removes it completely from the function call.

In addition, we now print the number of resource variables if it is greater than zero, so the user can use it to get the count when doing C++.

BUG=[251851084](https://b.corp.google.com/issues/251851084)


tensorflow/lite/micro/python/interpreter/src/tflm_runtime.py
```diff
@@ -39,16 +35,14 @@ def __init__(self,
...
-    if num_resource_variables != 0:
-      print(
-          'WARNING: num_resource_variables is deprecated. Resource variable count is ',
-          'calculated automatically.')
-    number_resource_variables = flatbuffer_utils.count_resource_variables(
+    num_resource_variables = flatbuffer_utils.count_resource_variables(
         model_data)
+    if num_resource_variables != 0:
+      print("Note: Model makes use of resource variables, count = ",
+            num_resource_variables)
```
advaitjain, line 42: let's print even if the number is zero. So there is no ambiguity.

suleshahid: Done.

advaitjain (approved):

advaitjain (approved):

advaitjain (approved):

PR #1745, opened 2023-02-15
We can now measure the latency of each test files.

tanmaydas82:
While finding out the slow tests in the test suite for a particular xtensa core, it required code changes to find out the tests that were taking long. The tests that run using makefile does not have a log about the duration it took for a particular test
file to run. We are adding that log to all the tests. The log message will look like below 
```
Running feature_provider_mock_test took 0.017 seconds
```

BUG=https://b.corp.google.com/issues/268565399


advaitjain (approved):

tanmaydas82:
@Mergifyio refresh

PR #1750, opened 2023-02-16
Adds default values to MicroResourceVariable

suleshahid:
Currently we memset MicroResourceVariables buffers to 0 if it is null, and in reset we memset to 0 again.
This change makes it so that we memset to the zero_point instead.

To get the information of the zero_point value while reset, we need to store it in the MicroResourceVariable class.

BUG=[269648474](https://b.corp.google.com/issues/269648474)


tensorflow/lite/micro/micro_resource_variable.cc
```diff
@@ -96,9 +97,14 @@ TfLiteStatus MicroResourceVariables::Allocate(int id, TfLiteContext* context,
       MicroPrintf("Failed to allocate resource buffer.");
       return kTfLiteError;
     }
-    // Zero out resource buffers by deafult. Buffers can be initialized to
-    // nonzero values using ASSIGN_VARIABLE.
-    memset(variable.resource_buffer, 0, variable.bytes);
+    // Set resource buffers to the zero_point by default. Buffers can be
```
advaitjain, line 100: let's add a comment here that link to a specific comment within the bug for more context. That will help the reader understand why this was done. And then we can figure out if zero point is the generic solution or if we want to go in a different direction later.

advaitjain (changes requested):
what is MRV in the PR description? can you update it to the full text?

tensorflow/lite/micro/micro_resource_variable.cc
suleshahid: Done.

advaitjain (dismissed):

suleshahid:
Note, not going to merge this yet. I have cl/510282473 for testing this in g3 as well and we are probably going to need to change a lot of arena sizes if they were set exact and use resource variables.

paulinesho (approved):

advaitjain (approved):

PR #1754, opened 2023-02-17
Add int16 support for SQUARED_DIFFERENCE

rewu93:
BUG=http://b/269352046

Add INT16 support for SQUARED_DIFFERENCE by templating the kernel implementation. Reuse int8 test settings to test the int16 implementation as well


tensorflow/lite/micro/kernels/squared_difference.cc
```diff
@@ -124,6 +124,56 @@ TfLiteStatus SquaredDifferencePrepare(TfLiteContext* context,
         std::numeric_limits<int8_t>::min();
     data->arithmetic_params.quantized_activation_max =
         std::numeric_limits<int8_t>::max();
+  } else if (input1->type == kTfLiteInt16) {
```
advaitjain, line 127: only the TF_LITE_ENSUREs are different between int8 and int16, right? And the last QuantizeMultipler.

Can we share the rest of the code? We will still need an if for quantized type vs float, right?

If it helps you can break up the logic into a PrepareQuantized that only gets called if input type is not a float.

tensorflow/lite/micro/kernels/squared_difference.cc
```diff
@@ -124,6 +124,56 @@ TfLiteStatus SquaredDifferencePrepare(TfLiteContext* context,
...
+    const auto& input1_quantization_params = input1->params;
+    const auto& input2_quantization_params = input2->params;
+    const auto& output_quantization_params = output->params;
```
advaitjain, line 130: let's not use auto here since the types are not apparent from the immediate context.

advaitjain (changes requested):

tensorflow/lite/micro/kernels/squared_difference.cc
rewu93: Done.

tensorflow/lite/micro/kernels/squared_difference.cc
rewu93: Done.

advaitjain (approved):

PR #1756, opened 2023-02-17
Match name change in upstream (http://cl/510336249)

advaitjain:
The reason for this name change is to allow for a BUILD target in upstream TF Lite as well that explicitly avoids pulling in Eigen headers.

Also, update the existing dependencies on `tensor_utils` to the updated `tensor_utils_no_eigen`.

BUG=http://b/269699951


suleshahid (approved):
LGTM.

PR #1773, opened 2023-02-20
Generate LSTM kernel testing golden values using python 

rewu93:
BUG=http://b/269673624
Python code that generate the expected values used in C++ lstm kernel testing.


advaitjain (approved):
I'm not confident that this is a good long term location for this code, but I don't have a better suggestion right now. Aproving.

PR #1792, opened 2023-02-28
Import tflite code to tflm, manually.

tanmaydas82:
The automatic sync job from tflite to tflm is failing due to a unsigned and signed long int comparison failure . The two commits can be described as

1. The output of the sync_from_upstream_tf.sh. This script pull in the differences from the tflite repo to tflm repo.
2. http://b/271154304: We cannot compare between unsigned and signed long int.

BUG=http://b/271154304
NO_CHECK_TFLITE_FILES=automated sync file changes with some bug fixes

### DIFF
--- a/tensorflow/lite/c/common.h
+++ b/tensorflow/lite/c/common.h
@@ -38,10 +38,4 @@ limitations under the License.
 
 #include "tensorflow/lite/core/c/common.h"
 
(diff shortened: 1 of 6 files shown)


tanmaydas82:
This PR is just to show that the fix works. We are fixing the issue at the source of truth.

PR #1795, opened 2023-03-01
Fix the project generation for hello_world.

tanmaydas82:
This PR is going to fix the project generation CI build. 

We are changing evaluate_test to not use the TFLM unit testing framework. This will allow the binary to be used independent of the TFLM repository which can be helpful when porting TFLM.

At the same time, we would like to continue to be able to have evaluate_test be part of the unit test suite which requires returning a specific string when the test is successful.

BUG=http://b/267222141


tensorflow/lite/micro/examples/hello_world/evaluate_test.cc
```diff
@@ -49,68 +46,94 @@ TF_LITE_MICRO_TEST(LoadFloatModelAndPerformInference) {
...
 
-TF_LITE_MICRO_TESTS_END
+int main(int argc, char* argv[]) {
+  int status = LoadFloatModelAndPerformInference();
+  if (status == kTfLiteOk) {
+    MicroPrintf("~~~ALL TESTS PASSED~~~\n");
+    return kTfLiteOk;
+  } else {
+    MicroPrintf("~~~SOME TESTS FAILED~~~\n");
+    return kTfLiteError;
```
advaitjain, line 137: let's add a comment about why these specific strings are needed. i.e without these this binary can not be used as part of the unit tests with the TFLM makefiles.

advaitjain (changes requested):
Please also expand the PR description with context on what the change is.

We are changing evaluate_test to not use the TFLM unit testing framework. This will allow the binary to be used independent of the TFLM repository which can be helpful when porting TFLM.

At the same time, we would like to continue to be able to have evaluate_test be part of the unit test suite which requires returning a specific string when the test is successful.

tensorflow/lite/micro/examples/hello_world/evaluate_test.cc
tanmaydas82: done

advaitjain (approved):

PR #1817, opened 2023-03-10
Explicitly use TfLiteRegistration_V1 in TFLM.

advaitjain:
Change generated with the following command:
```bash
cd tensorflow/lite/micro/
sed -i 's/TfLiteRegistration/TfLiteRegistration_V1/g' `find . -name "*.h" -o -name "*.cc"`
cd -
cd third_party/hexagon/
sed -i 's/TfLiteRegistration/TfLiteRegistration_V1/g' `find . -name "*.h" -o -name "*.cc"`
cd -
tensorflow/lite/micro/tools/ci_build/test_code_style.sh --fix_formatting
```

BUG=http://b/272808609

PR #1818, opened 2023-03-10
Decouple TFLM and TF Lite OpResolvers.

advaitjain:
With this change, the `MicroOpResolver` no longer inherits from the TfLite `OpResolver`. As a result:
 * `GetRegistrationFromOpCode` is now duplicated between TF Lite and TFLM.
 * This duplication is a necessary step to allow TFLM to pin to `TfLiteRegistration_V1`
 * A follow-up PR will introduce a find and replace modification to switch everything over to `TfLiteRegistration_V1`: https://github.com/tensorflow/tflite-micro/pull/1817
 * A side-benefit of this duplication is that the error message from unresolved OPs no longer has the version number which has been confusing for TFLM users for a long time.

BUG=http://b/272808609


vamsimanchala (approved):
LGTM

PR #1830, opened 2023-03-15
Add check that output is quantized.

tanmaydas82:
BUG=http://b/267818294


tensorflow/lite/micro/kernels/add_common.cc
```diff
@@ -90,6 +90,11 @@ TfLiteStatus AddPrepare(TfLiteContext* context, TfLiteNode* node) {
   TfLiteTensor* output =
       micro_context->AllocateTempOutputTensor(node, kAddOutputTensor);
   TF_LITE_ENSURE(context, output != nullptr);
+  // This is to ensure that for int8/int16 output type,output does not have
+  // a "no-quantization" quantization type.
+  if (output->type == kTfLiteInt8 || output->type == kTfLiteInt16) {
+    TFLITE_CHECK_NE(output->quantization.type, kTfLiteNoQuantization);
```
emmett-google, line 96: I'd suggest moving this into `CalculateOpDataAdd` in Line 42 for a few long term maintenance reasons:

 * The assert is closer to the code that assumes quantization.
 * One fewer branch to maintain for checking output types. If someone extends the branch logic in `CalculateOpDataAdd`, it will only need to be changed in on place.

tanmaydas82: Thank you for the suggestion! Made the change.

tensorflow/lite/micro/kernels/add_common.cc
```diff
@@ -39,6 +39,10 @@ TfLiteStatus CalculateOpDataAdd(TfLiteContext* context, TfLiteAddParams* params,
   data->requires_broadcast = !HaveSameShapes(input1, input2);
 
   if (output->type == kTfLiteInt8 || output->type == kTfLiteInt16) {
+    // This is to ensure that for int8/int16 output type,output does not have
```
advaitjain, line 42: Nit: Remove the comment. It's basically explaining the code.

tanmaydas82: Done.

advaitjain (approved):

PR #1831, opened 2023-03-15
changed to pull_request_target to allow git command to tty

jwithers:
post-merge.yml is failing due to the git commands there not being able to get a tty. I think this is because this needs a pull_request_target trigger instead of pull_request.


advaitjain (approved):

advaitjain:
Merged without waiting for CI since these changes will not be tested with any of the pre-submit checks.

PR #1834, opened 2023-03-17
Speedup for cc byte array generation.

emmett-google:
Previously the file was read one byte at a time. This improves this behavior to read a file at once, and then convert to a hex array.

The speed up is particularly noticeable when generating multiple files or a single large file.

BUG=https://b/273650840


paulinesho (approved):
Thanks for making this more efficient!

PR #1844, opened 2023-03-23
feat(python): add crosscheck module for testing micro vs lite

rkuester:
Add a Python module `tflite_micro.test_utils.crosscheck` for
testing the micro interpreter against the lite interpreter. The
initial facility is the `versus_lite` function, which accepts any
model, runs both interpreters with the same random input, and
expects equivalent outputs.

This initial implementation may have some limitations in the
characteristics of models it accepts, the degree to which
quantized ranges are uniformly exercised, etc. Improvement over
time is expected. The name crosscheck is general enough to
include testing micro against other interpreters in the future.

Include unit tests which test the module and serve as a simple
example of its use. Copy the .tflite model file to use in the
unit tests from the micro_speech example as of https://github.com/tensorflow/tflite-micro/commit/75f52901ebf5f1d6fbcef7e65e5e99e31ac63e44. Copy
rather than reference to avoid depending on the micro_speech
example's private implementation.

BUG=see description


python/tflite_micro/test_utils/crosscheck_test.model.tflite
rkuester, line 1: Make a bug about using a checked-in .tflite file here.

python/tflite_micro/test_utils/crosscheck_test.py
```diff
@@ -0,0 +1,95 @@
...
+"""Unit tests for crosscheck"""
+
+import numpy as np
+from tensorflow import keras
+from tensorflow import test
+from tensorflow.python.platform import resource_loader
+from tflite_micro.python.tflite_micro.test_utils import crosscheck
+from tflite_micro.tensorflow.lite.micro.python.interpreter.src import tflm_runtime
+
+_TEST_MODEL_PATH = resource_loader.get_path_to_datafile(
```
rkuester, line 24: Make a bug about the resource loader only working with data files in the same directory as the source file.

PR #1847, opened 2023-03-23
Int32 add

emmett-google:
This mirrors TfLite support for unquantized Int32 add operations.

BUG=https://b/273650194


tensorflow/lite/micro/kernels/add_test.cc
```diff
@@ -106,6 +132,26 @@ void TestAddFloat(int* input1_dims_data, const float* input1_data,
...
+
+  constexpr int inputs_size = 2;
+  constexpr int outputs_size = 1;
+  constexpr int tensors_size = inputs_size + outputs_size;
+  TfLiteTensor tensors[tensors_size] = {
+      CreateTensor(input1_data, input1_dims),
+      CreateTensor(input2_data, input2_dims),
+      CreateTensor(output_data, output_dims),
+  };
+  ValidateAddGoldensInt(tensors, tensors_size, expected_output, output_data,
```
paulinesho, line 151: Is there any issue using `ValidateAddGoldens()` here instead? `TF_LITE_MICRO_EXPECT_NEAR` skips the tolerance if the two values are equal. If not, I think it's best to avoid adding yet another helper function.

tensorflow/lite/micro/kernels/add_test.cc
```diff
@@ -27,15 +27,16 @@ namespace testing {
 namespace {
 
 // Shapes and values for mixed broadcast tests.
-constexpr int broadcast_output_dims_count = 36;
-constexpr int broadcast_num_shapes = 4;
+constexpr int kBroadcastOutputDimsCount = 36;
```
paulinesho, line 30: Thanks for making these changes as you go along!

tensorflow/lite/micro/kernels/cmsis_nn/add.cc
```diff
@@ -302,8 +341,10 @@ TfLiteStatus EvalAdd(TfLiteContext* context, TfLiteNode* node) {
   TFLITE_DCHECK(node->user_data != nullptr);
   const OpData* data = static_cast<const OpData*>(node->user_data);
 
-  if (output->type == kTfLiteFloat32) {
-    EvalAddFloat(context, node, params, data, input1, input2, output);
+  if (output->type == kTfLiteFloat32 ||
+      (input1->type == kTfLiteInt32 && input2->type == kTfLiteInt32)) {
```
paulinesho, line 345: Is it sufficient/equivalent to just check `output->type`? It's a bit more readable as it's what the rest of the function does.

tensorflow/lite/micro/test_helpers.h
```diff
@@ -216,7 +216,6 @@ TfLiteTensor CreateTensor(const T* data, TfLiteIntArray* dims,
   result.is_variable = is_variable;
   result.allocation_type = kTfLiteMemNone;
   result.data.data = const_cast<T*>(data);
-  result.quantization = {kTfLiteAffineQuantization, nullptr};
```
paulinesho, line 219: Oof, thanks for catching this!

tensorflow/lite/micro/kernels/add_test.cc
emmett-google: Good call, I assumed it did a floating point comparison but looking at the macro it should be safe with int32.

tensorflow/lite/micro/kernels/cmsis_nn/add.cc
emmett-google: I can switch to use `output->type` for consistency.

Longer term we may need a larger refactor where we want to check the quantization params. For example, you could have a quantized add and unquantized add that both emit `int16`.

tensorflow/lite/micro/kernels/add_test.cc
emmett-google: Fixing lint issues can be very fulfilling! 🧹 ✨

tensorflow/lite/micro/test_helpers.h
emmett-google: I had a hard time tracking down where this line came from, and it seems like an error. I couldn't find any cases that assumed this incorrect behavior and tests are passing with this fix.

paulinesho (approved):

PR #1855, opened 2023-03-28
Update Dockerfile.micro to use Python 3.10

tanmaydas82:
Additionally:
1. update to use `bullseye`
2. change how we `install clang` to make it easier to maintain and hopefully avoid occasional errors with the debian package installation

BUG=http://b/275010053


tensorflow/lite/micro/tools/ci_build/test_code_style.sh
```diff
@@ -34,6 +34,8 @@ set +e
 # --fix_formatting to let the script fix both code and build file format error.
 FIX_FORMAT_FLAG=${1}
 
+git config --global --add safe.directory $(pwd)
```
advaitjain, line 37: Can you describe why this is needed?

I'm not convinced that this is the right location and knowing more about the specific error will help me recommend a more suitable location for this.

advaitjain (changes requested):

tensorflow/lite/micro/tools/ci_build/test_code_style.sh
tanmaydas82: Let me tell you the truth! I am also not 100% sure that this is the correct location for this change. The error I was hitting into is 
```
fatal: detected dubious ownership in repository at '/github/workspace'
```
The error is coming with a solution 
```
git config --global --add safe.directory /github/workspace
```
I figured out, while running the solution locally, that `/github/workspace` should be the `pwd`.  I have updated the bug with details of the changes. Please let me know if you have other questions. Looking forward to your suggestion of the exact location for the change. Thank you!

advaitjain: let's continue the conversation from http://b/275010053#comment8

advaitjain (approved):

tanmaydas82:
@mergifyio refresh

PR #1862, opened 2023-03-30
Removed a few more files that are specific to STM32F4.

advaitjain:
Missed removing these files with https://github.com/tensorflow/tflite-micro/pull/1861

BUG=http://b/275917313


tanmaydas82 (approved):
LGTM

PR #1866, opened 2023-03-30
Switch dequantize scale to operate on floats.

emmett-google:
This speeds up the reference implementation to avoid double multiplication.

In addition, we delay the type promotion for the input type until it is used in subtraction. 

BUG=https://b/271611642


emmett-google:
Upstreaming this change to TfLite

PR #1882, opened 2023-04-05
Change RuntimeShape to support max size of 6 dimensions.

suleshahid:
Our upstream sync with TF is broken due to [b/275337654](http://b/275337654) which introduces 6D usage in the Add operator.

RuntimeShape.h is managed separately for TFLM vs TFLite.
Since we do not have dynamic memory allocation, we have to explicitly set the max number of dimensions, as previously it was 5.

* Changing the max size to 6 to support the changes to Add.
BUG=[276954955](https://b.corp.google.com/issues/276954955)


advaitjain (commented):
Will it be possible to change kmaxsmall size to 6 in a separate PR and then have the sync pr get merged as is?

suleshahid:
> Will it be possible to change kmaxsmall size to 6 in a separate PR and then have the sync pr get merged as is?

Yes. Will remove those files from this PR.

advaitjain (approved):

Issue #1891, opened 2023-04-10
Replace renode with QEMU for cross compile testing


mbernico:
Renode is currently used for emulation, however it's somewhat slow and complicated.  It's probably 'too much' for  CI testing.  We should investigate using QEMU in place of Renode.

mbernico:
Adding some direction that was originally in #1892 for easier access by @advaitjain 

Next steps for this PR:

move qemu installation out of the make build system and to the ci/ folder - done
assume that qemu is in $PATH and require TFLM users to set that up. - done
remove the CI bits from this PR. - done
clean up and marge this PR with the ability to manually run tensorflow/lite/micro/tools/ci_build/test_cortex_m_binary.sh. - done 
Rename to test_cortex_m_qemu.sh for consistency. - done

add qemu installation to the Docker container - done

add new workflow to ci.yml that uses the docker container to execute test_cortex_m_qemu.sh, similar to
[tflite-micro/.github/workflows/ci.yml](https://github.com/tensorflow/tflite-micro/blob/e1bba19f71774b83b66a6c8e27fca62f54e32e19/.github/workflows/ci.yml#L182-L184)

Even more follow-up (and reminder for myself):

before we disable bluepill we will want to have a build with -nostdlib similar to
[tflite-micro/tensorflow/lite/micro/tools/make/targets/bluepill_makefile.inc](https://github.com/tensorflow/tflite-micro/blob/e1bba19f71774b83b66a6c8e27fca62f54e32e19/tensorflow/lite/micro/tools/make/targets/bluepill_makefile.inc#L32)

Line 32 in [e1bba19](https://github.com/tensorflow/tflite-micro/commit/e1bba19f71774b83b66a6c8e27fca62f54e32e19)

 -nostdlib 

since that is the only place where the CI detects dynamic memory allocation and other non-embedded friendly code creeping into the framework. I need to think about the right approach for that.

PR #1892, opened 2023-04-10
Add QEMU Emulation to CI flow

mbernico:
Addresses #1891 and adds Cortex M QEMU based unit tests to the CI flow.

Note this doesn't remove or change anything related to renode, for the sake of brevity, but we can remove renode here if desired.


tensorflow/lite/micro/tools/make/targets/cortex_m_generic_makefile.inc
```diff
@@ -204,3 +241,23 @@ ifneq ($(TARGET_ARCH), project_generation)
...
+  $(TENSORFLOW_ROOT)tensorflow/lite/micro/micro_allocator_test.cc \
+  $(TENSORFLOW_ROOT)tensorflow/lite/micro/memory_arena_threshold_test.cc
+
+# flatbuffer_utils_test is intentionaly disabled because the flexbuffer builder
+# uses dynamic memory.
+EXCLUDED_TESTS += $(TENSORFLOW_ROOT)tensorflow/lite/micro/flatbuffer_utils_test.cc
+
+MICROLITE_TEST_SRCS := $(filter-out $(EXCLUDED_TESTS), $(MICROLITE_TEST_SRCS))
+
+# If the target is supported by QEMU, setup the test and size scripts.
```
shlmregev, line 260: I think all the architectures are supported, so we can remove this condition and add it only if necessary.

tensorflow/lite/micro/tools/make/targets/cortex_m_generic_makefile.inc
```diff
@@ -16,6 +16,21 @@
 # Generic Makefile target for ARM Cortex M builds.
 # For more info see: tensorflow/lite/micro/cortex_m_generic/README.md
 
+DOWNLOAD_RESULT := $(shell $(MAKEFILE_DIR)/arm_gcc_download.sh ${DOWNLOADS_DIR} $(TENSORFLOW_ROOT))
```
shlmregev, line 19: There needs to be a mechanism to use a different toolchain than the one downloaded. The corstone300 makefile has something like this:
  TARGET_DEFAULT_TOOLCHAIN_ROOT := $(MAKEFILE_DIR)/downloads/gcc_embedded/bin/
  TARGET_TOOLCHAIN_ROOT := $(TARGET_DEFAULT_TOOLCHAIN_ROOT)
  ifeq ($(TARGET_TOOLCHAIN_ROOT), $(TARGET_DEFAULT_TOOLCHAIN_ROOT))
    DOWNLOAD_RESULT := $(shell $(MAKEFILE_DIR)/arm_gcc_download.sh ${MAKEFILE_DIR}/downloads)
    ifneq ($(DOWNLOAD_RESULT), SUCCESS)
      $(error Something went wrong with the GCC download: $(DOWNLOAD_RESULT))
    endif
  endif
Then you can override the toolchain selection by defining TARGET_TOOLCHAIN_ROOT on the 'make' command line.
To use the toolchain, we also need to add:
export PATH := $(TARGET_TOOLCHAIN_ROOT):$(PATH)
I'm not sure how it works for you know - it's possible that you have the path updated in your env.

mansnils:
The only target that use Renode is Bluepill as far as I know.

The cortex_m_generic target is called generic since it is not bound to any specific platform within the TFLM repo. So cortex_m_generic does not build any unit tests or examples, just the micro_lite lib. The intention is then to include the lib within an application. The cortex_m_corstone_300 target is the equivalent target where Arm(R) Corstone(TM)-300 software is used to run unit tests and examples. Therefore I think for Qemu there should be a new target, cortex_m_qemu.

tensorflow/lite/micro/testing/test_with_qemu.sh
```diff
@@ -0,0 +1,47 @@
+
+#!/bin/bash -e
```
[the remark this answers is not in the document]
mbernico: Currently that's true, but QEMU does support many different CPUs beyond the Cortex M.  What if we call it test_emulated_binary.sh?

tensorflow/lite/micro/tools/make/targets/cortex_m_generic_makefile.inc
mbernico: that's a good callout.  Updated, thanks!

tensorflow/lite/micro/tools/make/targets/cortex_m_generic_makefile.inc
mbernico: Updated to add the export.  The rest of this existed already on line 170

tensorflow/lite/micro/tools/make/qemu_download.sh
```diff
@@ -0,0 +1,63 @@
...
+  mkdir ${DOWNLOADED_QEMU_PATH}
+  LINUX_PORTABLE_URL="https://download.qemu.org/qemu-6.2.0.tar.xz"
+  TEMP_ARCHIVE="/tmp/qemu.tar.xz"
+
+  echo >&2 "Downloading from url: ${LINUX_PORTABLE_URL}"
+  wget ${LINUX_PORTABLE_URL} -O ${TEMP_ARCHIVE} >&2
+  tar xJf ${TEMP_ARCHIVE} --strip-components=1 --directory ${DOWNLOADED_QEMU_PATH} >&2
+  cd ${DOWNLOADED_QEMU_PATH}
+  ./configure
+  make -j8
```
advaitjain, line 59: The way our CI is set up, this will end up rebuilding qemu from source each time. If building from source is unavoidable, we should change the approach here to install qemu external to the make command and just assume that it is pre-installed.

This will mean running the qemu test script via a docker container instead of directly from the GitHub runner.

advaitjain (commented):
I'm in the process of trying this out locally and will add comments or directly push commits in a bit.

Adding in one comment as a reminder to revisit how we set this up.

mansnils (changes requested):
I think this should be in a new makefile, for example cortex_m_qemu_makefille.inc. See this comment for the reasoning why: https://github.com/tensorflow/tflite-micro/pull/1892#issuecomment-1503042781

advaitjain:
> I think this should be in a new makefile, for example cortex_m_qemu_makefille.inc. See this comment for the reasoning why: [#1892 (comment)](https://github.com/tensorflow/tflite-micro/pull/1892#issuecomment-1503042781)

Its going to take me a few days to give this a try and form an opinion on how we might want to structure this. @mansnils, I'll consult with you as well.

mbernico:
> > I think this should be in a new makefile, for example cortex_m_qemu_makefille.inc. See this comment for the reasoning why: [#1892 (comment)](https://github.com/tensorflow/tflite-micro/pull/1892#issuecomment-1503042781)
> 
> Its going to take me a few days to give this a try and form an opinion on how we might want to structure this. @mansnils, I'll consult with you as well.

@mansnils It's a very good point.  @advaitjain I'll wait for your direction on this one.

Generally, I think if we were trying to emulate a specific board (aka bluepill) instead of all cortex-m or riscv (@shlmregev ) then its probably cleaner to use a new target. Since we're trying to emulate / exercise the binary across all supported CPUs, it probably makes sense to keep it this way.  Also this impacts the design of #1894 

That said,  don't really have a strong opinion here. I'm the newbie so someone tell me what to do. 👶

tensorflow/lite/micro/cortex_m_generic/micro_time.cc
```diff
@@ -31,6 +35,18 @@ namespace tflite {
 uint32_t ticks_per_second() { return 0; }
 uint32_t GetCurrentTimeTicks() { return 0; }
 
+#elif defined(USE_SEMIHOSTING)
+
+// For QEMU Semihosting these reqeusts are passed to the host OS.
+uint32_t ticks_per_second() { return 0; }
+
+uint32_t GetCurrentTimeTicks() {
+  uint32_t ms = std::chrono::duration_cast<std::chrono::milliseconds>(
```
paulinesho, line 44: @mbernico Thank you so much for this PR! I gave it a go and noticed that this line always returns the same (non-zero) number each run. This is not the case on x86. Any idea what the issue is?

mbernico: Very good question.  I'm not much of a C++ developer yet, but here is what I think is happening...  

In the makefile, line 229 I'm using semihosting with ARM (PLATFORM_FLAGS += --specs=rdimon.specs)

When you do that, it stubs out system calls that would need to be defined if we were compiling for a specific board, and generally passes them to the host operating system when QEMU is used.  This allows me to run the tests in user space, really quickly, but it's not a perfect test because we aren't building against a specific device.

To that end, I think that std::chrono::system_clock::now() probably makes a call to a system call that isn't built out.  It's just a stub.  See this for a discussion on the same. https://community.renesas.com/mcu-mpu/embedded-system-platform/f/forum/6267/gettimeofday-always-returns-17-for-the-time

It sounds like alternatively I could have implemented my own version of all those system calls but I'm not even close to smart enough to do that. So, instead I set the ticks to 0 so that test would pass.  :)

advaitjain:
Let's do this in smaller chunks. Goal of the current PR should be to get to functionality similar to bluepill + renode manually verified on the command line without any CI integration.

I agree with @mansnils that we should create a separate makefile for cortex_m_qemu and leave cortex_m_generic unchanged.

The TFLM make build system is hard to work with so we'll need to do some smoke checks and I did some prototyping with [`5e68c38` (#1892)](https://github.com/tensorflow/tflite-micro/pull/1892/commits/5e68c385331208350808cdfa77b60a095ef9511a)

Here is what I tested:

Command:
```
make -f tensorflow/lite/micro/tools/make/Makefile TARGET=cortex_m_qemu TARGET_ARCH=cortex-m3 OPTIMIZED_KERNEL_DIR=cmsis_nn test_kernel_add_test -j8
```

 * confirmed that the test passed
 * manually introduce failure in `micro/kernels/add_test.cc` and confirmed that the failure was detected
 * confirmed that the logs were stored in `/tmp/test_cortex_m_qemu/gen/cortex_m_qemu_cortex-m3_default/bin/kernel_add_test/logs.txt`
 * manually introduced syntax error in `micro/kernels/cmsis_nn/add.cc` and confirmed that the CMSIS kernels were being used (i.e. build failed).

In the process of this testing, I had to update the test_latency_log.sh script to allow for variable number of parameters.

Next steps for this PR:
 * move qemu installation out of the make build system and to the `ci/` folder
 * assume that qemu is in `$PATH` and require TFLM users to set that up.
 * remove the CI bits from this PR.
 * clean up and marge this PR with the ability to manually run `tensorflow/lite/micro/tools/ci_build/test_cortex_m_binary.sh`. Rename to `test_cortex_m_qemu.sh` for consistency.

Follow-up PR:
 * add qemu installation to the Docker container
 * add new workflow to ci.yml that uses the docker container to execute `test_cortex_m_qemu.sh`, similar to https://github.com/tensorflow/tflite-micro/blob/e1bba19f71774b83b66a6c8e27fca62f54e32e19/.github/workflows/ci.yml#L182-L184


Even more follow-up (and reminder for myself):
 * before we disable bluepill we will want to have a build with `-nostdlib` similar to https://github.com/tensorflow/tflite-micro/blob/e1bba19f71774b83b66a6c8e27fca62f54e32e19/tensorflow/lite/micro/tools/make/targets/bluepill_makefile.inc#L32 since that is the only place where the CI detects dynamic memory allocation and other non-embedded friendly code creeping into the framework. I need to think about the right approach for that.

mbernico:
> Let's do this in smaller chunks. Goal of the current PR should be to get to functionality similar to bluepill + renode manually verified on the command line without any CI integration.
> 
> I agree with @mansnils that we should create a separate makefile for cortex_m_qemu and leave cortex_m_generic unchanged.
> 
> The TFLM make build system is hard to work with so we'll need to do some smoke checks and I did some prototyping with [`5e68c38` (#1892)](https://github.com/tensorflow/tflite-micro/pull/1892/commits/5e68c385331208350808cdfa77b60a095ef9511a)
> 
> Here is what I tested:
> 
> Command:
> 
> ```
> make -f tensorflow/lite/micro/tools/make/Makefile TARGET=cortex_m_qemu TARGET_ARCH=cortex-m3 OPTIMIZED_KERNEL_DIR=cmsis_nn test_kernel_add_test -j8
> ```
> 
> * confirmed that the test passed
> * manually introduce failure in `micro/kernels/add_test.cc` and confirmed that the failure was detected
> * confirmed that the logs were stored in `/tmp/test_cortex_m_qemu/gen/cortex_m_qemu_cortex-m3_default/bin/kernel_add_test/logs.txt`
> * manually introduced syntax error in `micro/kernels/cmsis_nn/add.cc` and confirmed that the CMSIS kernels were being used (i.e. build failed).
> 
> In the process of this testing, I had to update the test_latency_log.sh script to allow for variable number of parameters.
> 
> Next steps for this PR:
> 
> * move qemu installation out of the make build system and to the `ci/` folder
> * assume that qemu is in `$PATH` and require TFLM users to set that up.
> * remove the CI bits from this PR.
> * clean up and marge this PR with the ability to manually run `tensorflow/lite/micro/tools/ci_build/test_cortex_m_binary.sh`. Rename to `test_cortex_m_qemu.sh` for consistency.
> 
> Follow-up PR:
> 
> * add qemu installation to the Docker container
> * add new workflow to ci.yml that uses the docker container to execute `test_cortex_m_qemu.sh`, similar to https://github.com/tensorflow/tflite-micro/blob/e1bba19f71774b83b66a6c8e27fca62f54e32e19/.github/workflows/ci.yml#L182-L184
> 
> Even more follow-up (and reminder for myself):
> 
> * before we disable bluepill we will want to have a build with `-nostdlib` similar to https://github.com/tensorflow/tflite-micro/blob/e1bba19f71774b83b66a6c8e27fca62f54e32e19/tensorflow/lite/micro/tools/make/targets/bluepill_makefile.inc#L32
>    since that is the only place where the CI detects dynamic memory allocation and other non-embedded friendly code creeping into the framework. I need to think about the right approach for that.

SGTM,  I updated the PR to match this plan.

.github/workflows/ci.yml
```diff
@@ -198,7 +198,7 @@ jobs:
         run: |
           pip3 install Pillow
           pip3 install Wave
-          pip3 install numpy
+        run: pip3 install numpy
```
advaitjain, line 201: is this a typo? was this change needed?

ci/Dockerfile.hexagon
```diff
@@ -12,7 +12,7 @@ RUN \
     wget \
     python3 \
     python \
-    python3-pip
+    python3-pip
```
advaitjain, line 15: remove trailing whitespace?

ci/Dockerfile.micro
```diff
@@ -52,9 +55,6 @@ RUN pip install matplotlib
 
 RUN pip install six
 
-# Install Renode test dependencies
-RUN pip install pyyaml requests psutil robotframework==4.0.1
-
```
advaitjain, line 57: let's keep this for now.

tensorflow/lite/micro/cortex_m_generic/debug_log.cc
advaitjain, line 1: changes in this file can be reverted now that we are not using qemu with cortex_m_generic?

tensorflow/lite/micro/cortex_m_generic/micro_time.cc
advaitjain, line 1: changes in this file can be reverted now that we are not using qemu with cortex_m_generic?

tensorflow/lite/micro/cortex_m_generic/micro_time.cc
advaitjain: let's do this in smaller chunks. For now we are going with a separate target `cortex_m_qemu` so we will end up using `micro/micro_time.cc` which defaults to returning 0.

We can investigate other ways to perhaps get meaninful profiling information from qemu later.

tensorflow/lite/micro/docs/qemu.md
```diff
@@ -0,0 +1,27 @@
...
+QEMU can quickly test unit tests that are cross compiled for non x64_86 
+hardware. 
+
+# Running Unit Tests
+All unit tests can be ran using `tensorflow/lite/micro/tools/ci_build/test_cortex_m_qemu.sh` for the cortex-m processor.  
+
+# Useful External Links for QEMU
+The current QEMU implementation uses `user` mode.  The documentation for [user mode is here](https://www.qemu.org/docs/master/user/index.html).
+
+QEMU uses ARM [semihosting](https://github.com/ARM-software/abi-aa/blob/main/semihosting/semihosting.rst) to replace newlib system calls for specific boards with the host OS.  Further documentation on how this works is contained in `cortex_m_generic_makefile.inc` as well as `tensorflow/lite/micro/cortex_m_generic/`.
```
advaitjain, line 27: reminder to update the docs. Perhaps best to wait until we settle on the final structure of the code.

tensorflow/lite/micro/testing/test_emulated_binary.sh
advaitjain, line 1: remove?

tensorflow/lite/micro/tools/ci_build/test_cortex_m_qemu.sh
```diff
@@ -0,0 +1,36 @@
...
+# 2 - (optional) EXTERNAL_DIR: Path to the external directory that contains external code
+# Tests the microcontroller code with QEMU emulator
+
+set -e
+pwd
+
+TENSORFLOW_ROOT=${1}
+EXTERNAL_DIR=${2}
+TARGET=cortex_m_qemu
+TARGET_ARCH=${3:-cortex-m3}
```
advaitjain, line 27: let's also specify `OPTIMIZED_KERNEL_DIR=cmsis_nn` so that the optimized kernels are tested.

tensorflow/lite/micro/tools/make/targets/cortex_m_qemu_makefile.inc
```diff
@@ -0,0 +1,48 @@
...
+  -mfpu=auto \
+  -funsigned-char \
+  -mlittle-endian \
+  -fomit-frame-pointer \
+
+# Enable semihosting for QEMU to enable host system calls for debug_log and micro_time.
+PLATFORM_FLAGS += --specs=rdimon.specs
+# USE_SEMIHOSTING enables host system calls for debug_log and micro_time.
+# Please see micro/cortex_m_generic micro_time.cc and debug_log.cc for more.
+PLATFORM_FLAGS += -DUSE_SEMIHOSTING
```
advaitjain, line 43: this is not needed.

When `TARGET=cortex_m_qemu`,  we will end up using: https://github.com/tensorflow/tflite-micro/blob/2e2dfae3f6ee355649feb6ecc9dd6b13fd60ab21/tensorflow/lite/micro/micro_time.cc#L37-L47

and https://github.com/tensorflow/tflite-micro/blob/2e2dfae3f6ee355649feb6ecc9dd6b13fd60ab21/tensorflow/lite/micro/debug_log.cc#L43-L50

since there is no `micro/cortex_m_qemu` directory. We can fine tune what we want micro_time to be doing for qemu later, and fprintf worked just fine with qemu + semihosting.

Here is the logic if you're interested: https://github.com/tensorflow/tflite-micro/blob/2e2dfae3f6ee355649feb6ecc9dd6b13fd60ab21/tensorflow/lite/micro/tools/make/Makefile#L584-L588

advaitjain (changes requested):
Can I also request updating the PR description with the set of commands that you ran manually to confirm that things are working as intended?

.github/workflows/ci.yml
mbernico: I think this was from a previous merge.  I shouldn't be making any changes here now.  I'll revert ci.yaml to the current upstream.

ci/Dockerfile.hexagon
mbernico: Done.

ci/Dockerfile.hexagon
```diff
@@ -27,6 +27,9 @@ RUN \
 RUN \
   pip3 install numpy
 
+# Install QEMU requirements emulation based tests.
+RUN apt-get install -y libglib2.0-dev libfdt-dev libpixman-1-dev zlib1g-dev ninja-build
+
```
mbernico, line 32: This can be reverted

ci/Dockerfile.micro
```diff
@@ -29,6 +29,9 @@ RUN ln -s /usr/bin/clang++-16 /usr/bin/clang++
 RUN apt-get install clang-format-16
 RUN ln -s /usr/bin/clang-format-16 /usr/bin/clang-format
 
+# Install QEMU requirements emulation based tests.
+RUN apt-get install -y libglib2.0-dev libfdt-dev libpixman-1-dev zlib1g-dev ninja-build
+
```
mbernico, line 34: This can be reverted for now.

ci/Dockerfile.micro
mbernico: Oops, yes.

tensorflow/lite/micro/cortex_m_generic/debug_log.cc
mbernico: Yes, that's right.

tensorflow/lite/micro/cortex_m_generic/micro_time.cc
mbernico: Done

tensorflow/lite/micro/cortex_m_generic/micro_time.cc
mbernico: Okay, SGTM

tensorflow/lite/micro/docs/qemu.md
mbernico: Updated.

tensorflow/lite/micro/testing/test_emulated_binary.sh
mbernico: Yes, thanks

tensorflow/lite/micro/tools/ci_build/test_cortex_m_qemu.sh
mbernico: Done

mbernico:
@advaitjain I believe all changes should be made.  Sorry, I should have reviewed this a bit better before getting your code review.

advaitjain (approved):
I pushed a commit with some minor cleanup. Approving the PR.

PR #1895, opened 2023-04-11
Add a script to measure a Cortex-M binary size.

shlmregev:
BUG=277659730


tensorflow/lite/micro/tools/make/targets/cortex_m_generic_makefile.inc
```diff
@@ -204,3 +204,7 @@ ifneq ($(TARGET_ARCH), project_generation)
     $(TENSORFLOW_ROOT)tensorflow/lite/micro/examples/micro_speech/Makefile.inc
   MICRO_LITE_EXAMPLE_TESTS := $(filter-out $(EXCLUDED_EXAMPLE_TESTS), $(MICRO_LITE_EXAMPLE_TESTS))
 endif
+
+TEST_SCRIPT := $(TENSORFLOW_ROOT)tensorflow/lite/micro/testing/test_with_qemu.sh arm $(TARGET_ARCH)
+SIZE_SCRIPT := $(TENSORFLOW_ROOT)tensorflow/lite/micro/testing/size_cortex_m_binary.sh
```
advaitjain, line 209: can these be moved to [cortex_m_qemu_makefile](https://github.com/tensorflow/tflite-micro/blob/095d270fc16dc150c8547362b0a4c86fc1d065e0/tensorflow/lite/micro/tools/make/targets/cortex_m_qemu_makefile.inc)?

I would like to keep the cortex_m_generic_makefile independent of qemu.

tensorflow/lite/micro/testing/size_cortex_m_binary.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash -e
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
advaitjain, line 2: 2023

tensorflow/lite/micro/testing/size_cortex_m_binary.sh
```diff
@@ -0,0 +1,60 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+#
+# Neasures the size of a Qualcomm Hexagon binary by parsing the output of
+# hexgaon-size. If an optional list of symbols is provided, the symbols' sizes
+# are excluded from the total. This is useful when the binary contains symbols
+# that are only used during testing.
```
advaitjain, line 20: update the comment

advaitjain (changes requested):

tensorflow/lite/micro/tools/make/targets/cortex_m_generic_makefile.inc
mbernico: No opinions either way here, but I'm responsible for creating the cortex_m_qemu_makefile so if we decide we need to expand its functionality I'm happy to help.  I know that @advaitjain previously noted we need to make sure that testing can use -nostdlib before we deprecated renode.

advaitjain: @shlmregev, is there something missing in cortex_m_qemu_makefile today for your use-case that cortex_m_generic has?

cortex_m_generic was created to have a way to build the tflm static lib (no tests or binaries) and it can be used for various integrations of TFLM into other frameworks.

For that reason, and consistent with https://github.com/tensorflow/tflite-micro/pull/1892#issuecomment-1503042781, I would really like to keep it free from additional things like qemu.

PR #1907, opened 2023-04-17
Invoke failure throws ValueError.

tanmaydas82:
BUG=http://b/273622999

### DIFF
--- a/tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
+++ b/tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
@@ -253,12 +253,13 @@ InterpreterWrapper::InterpreterWrapper(
 
 void InterpreterWrapper::PrintAllocations() { allocator_->PrintAllocations(); }
 
-void InterpreterWrapper::Invoke() {
+int InterpreterWrapper::Invoke() {
   if (interpreter_->Invoke() != kTfLiteOk) {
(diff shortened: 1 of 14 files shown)


tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
```diff
@@ -253,7 +253,14 @@ InterpreterWrapper::InterpreterWrapper(
 
 void InterpreterWrapper::PrintAllocations() { allocator_->PrintAllocations(); }
 
-int InterpreterWrapper::Invoke() { return interpreter_->Invoke(); }
+int InterpreterWrapper::Invoke() {
+  if (interpreter_->Invoke() != kTfLiteOk) {
+    char err_strbuf[128];
```
advaitjain, line 258: This check should be `!= kTfLiteError` similar to https://github.com/tensorflow/tflite-micro/blob/7330d1cf7a720e5c539080189ae47760a127f941/tensorflow/lite/micro/micro_graph.cc#L181

This is needed because we don't want an exception for the case where we return say `kTfLiteAbort` which is a return value that doesn't signify an error.

tanmaydas82: Thank you for the review. Change the code accordingly.

advaitjain (approved):

PR #1919, opened 2023-04-20
Return the correct TfLiteStatus from interpreter->Invoke()

advaitjain:
We would like to throw and exception on kTfLiteError (http://b/273622999) and pass-through all the other TfLiteStatus returned by calling interpreter->Invoke()

https://github.com/tensorflow/tflite-micro/pull/1907 changed the behavior to throw an exception on kTfLiteError but did not pass through all the other status.

In particular, this is important because streaming implementation with circular_buffers relies on kTfLiteAbort being appropriately propagated.

BUG=http://b/279055226


tanmaydas82 (approved):
LGTM

PR #1921, opened 2023-04-20
Manually ran sync_from_upstream_tf to add tensorflow/lite/core/macros.h.

tanmaydas82:
BUG=http://b/279053959
NO_CHECK_TFLITE_FILES=manual sync from upstream


advaitjain (approved):
thanks for the fix!

advaitjain (approved):

PR #1934, opened 2023-04-24
Fix "fatal: detected dubious ownership in repository at '/github/workspace'"

advaitjain:
Putting the git config command as part of running the docker container does the trick (tested on my fork of tflite-micro).

Why adding the `git config` command to the Dockerfile.micro (from https://github.com/tensorflow/tflite-micro/pull/1886) does not solve the issue is not clear at this time.

BUG=http://b/275010053


advaitjain:
Merging without running the CI since that does not catch changes to the yaml file. Pair-debugged this issue with @tanmaydas82.

PR #1939, opened 2023-04-26
Replace snprintf with std::string in interpreter_wrapper.

tanmaydas82:
BUG=http://b/279061457

### DIFF
--- a/tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
+++ b/tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
@@ -231,10 +231,9 @@ InterpreterWrapper::InterpreterWrapper(
 
   for (const auto& registerer : registerers_by_name) {
     if (!AddCustomOpRegistererByName(registerer.c_str(), &all_ops_resolver_)) {
-      char strbuf[128];
-      snprintf(strbuf, sizeof(strbuf),
-               "TFLM could not register custom op via %s", registerer.c_str());
-      ThrowRuntimeError(strbuf);
+      std::string strbuf;
+      strbuf = "TFLM could not register custom op via " + registerer;
+      ThrowRuntimeError(strbuf.c_str());
     }
   }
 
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
```diff
@@ -231,10 +231,9 @@ InterpreterWrapper::InterpreterWrapper(
 
   for (const auto& registerer : registerers_by_name) {
     if (!AddCustomOpRegistererByName(registerer.c_str(), &all_ops_resolver_)) {
-      char strbuf[128];
-      snprintf(strbuf, sizeof(strbuf),
-               "TFLM could not register custom op via %s", registerer.c_str());
-      ThrowRuntimeError(strbuf);
+      std::string strbuf;
+      strbuf = "TFLM could not register custom op via " + registerer;
+      ThrowRuntimeError(strbuf.c_str());
```
advaitjain, line 236: this can be collapsed into a single line:
```cc
ThrowRuntimeError(("TFLM could not register custom op via " + registerer).c_str());
```

tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
```diff
@@ -231,10 +231,9 @@ InterpreterWrapper::InterpreterWrapper(
 
   for (const auto& registerer : registerers_by_name) {
```
advaitjain, line 232: unrelated to the current PR but we should not be using auto here since the type can not be inferred from the local context.

let's change to
```cc
const std::string& registerer : registerers_by_name
```

advaitjain (changes requested):

tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
tanmaydas82: done

tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
tanmaydas82: done

tensorflow/lite/micro/python/interpreter/src/interpreter_wrapper.cc
```diff
@@ -287,33 +284,30 @@ void InterpreterWrapper::SetInputTensor(PyObject* data, size_t index) {
     throw pybind11::error_already_set();
   }
 
-  char err_strbuf[128];  // buffer for error message since we can not use
-                         // absl
+  std::string err_str;
```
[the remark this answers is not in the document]
tanmaydas82: done

advaitjain (approved):

PR #1977, opened 2023-05-16
Adding python wrapper for tflite flatbuffer align tool

suleshahid:
The `tflite_flatbuffer_align` tool is currently a C++ script. The functionality of this script is only useful after you have modified the tflite flatbuffer via the Python API, then you would call this script to realign via the C++ API.

It would make more sense to wrap this tool in python, and use it directly in the python script, instead of having a two step procedure.
We leave a py binary with the same bazel target +args structure to keep support for any prior usecases.

BUG=[274637329](http://b/274637329)


tensorflow/lite/micro/tools/BUILD
```diff
@@ -64,13 +65,27 @@ py_test(
...
 
-cc_binary(
-    name = "tflite_flatbuffer_align",
+pybind_extension(
+    name = "tflite_flatbuffer_align_wrapper",  # :tflite_flatbuffer_align_wrapper.so
     srcs = [
-        "tflite_flatbuffer_align.cc",
+        "tflite_flatbuffer_align_wrapper.cc",
     ],
+    visibility = ["//visibility:public"],
```
advaitjain, line 73: let's avoid public visibility.

tensorflow/lite/micro/tools/tflite_flatbuffer_align.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2022 The TensorFlow Authors. All Rights Reserved.
```
advaitjain, line 1: 2023

tensorflow/lite/micro/tools/tflite_flatbuffer_align_wrapper.cc
```diff
@@ -0,0 +1,54 @@
...
+  // ModelT is just the unpacked version of the model file.
+  tflite::ModelT* unpacked_model = model->UnPack();
+  flatbuffers::FlatBufferBuilder fbb;
+  auto new_model = tflite::Model::Pack(fbb, unpacked_model);
+  fbb.Finish(new_model, tflite::ModelIdentifier());
+  flatbuffers::SaveFile(output_file_name,
+                        reinterpret_cast<char*>(fbb.GetBufferPointer()),
+                        fbb.GetSize(), /*binary*/ true);
+  int size = fbb.GetSize();
+  fprintf(stderr, "Aligned model size: %d\n", size);
```
advaitjain, line 45: let's move this out of the function call and perhaps into the py_binary tool. I'm also ok with removing the printf completely.

tensorflow/lite/micro/tools/tflite_flatbuffer_align_wrapper.cc
```diff
@@ -0,0 +1,54 @@
...
+#include "tensorflow/lite/schema/schema_generated.h"
+
+namespace py = pybind11;
+
+void align_tflite_model(const char* input_file_name,
+                        const char* output_file_name) {
+  std::string model_file;
+  // Read the file into a string using the included util API call:
+  flatbuffers::LoadFile(input_file_name, false, &model_file);
+  fprintf(stderr, "Original model size: %lu\n", model_file.size());
```
advaitjain, line 29: let's move the printf out into the py_binary. I'm also ok with removing the printf completely.

advaitjain (changes requested):

tensorflow/lite/micro/tools/tflite_flatbuffer_align_wrapper.cc
suleshahid: Done. Agreeing on removing it, if you are using via cmd line, `ls -l` is just as easy.

tensorflow/lite/micro/tools/tflite_flatbuffer_align.py
suleshahid: Done.

tensorflow/lite/micro/tools/BUILD
suleshahid: Done.

tensorflow/lite/micro/tools/tflite_flatbuffer_align_wrapper.cc
suleshahid: Done.

advaitjain (approved):

advaitjain (approved):

Issue #1826, opened 2023-03-14
CMSIS-NN LSTM int8 unit test is disabled


mansnils:
Disabled test: https://github.com/tensorflow/tflite-micro/blob/dff54f654551c1c5603db9e9abf1b11737e4c52e/tensorflow/lite/micro/kernels/unidirectional_sequence_lstm_test.cc#L138

This is likely because of: https://github.com/tensorflow/tflite-micro/pull/1797/

mansnils:
refresh

PR #1996, opened 2023-05-24
Add intermediate tensors to LSTM unit test to enable CMSIS-NN

mansnils:
Unit test for CMSIS-NN were failing because missing intermediate tensors, which was needed for the old LSTM implementation.
Therefore adding intermediate tensors since to the unit test since it should not affect the new reference implementation.

BUG=https://github.com/tensorflow/tflite-micro/issues/1826


advaitjain (approved):

advaitjain:
@mansnils, looks like there is a conflict that needs to be resolved.

mansnils:
> @mansnils, looks like there is a conflict that needs to be resolved.

Conflict resolved

advaitjain (approved):

PR #1999, opened 2023-05-25
Update visibility of recipes:resource_variable_lib and micro/models

suleshahid:
We need additional visibility into these files for some other tests.

BUG=[274635545](http://b/274635545)

### DIFF
--- a/tensorflow/lite/micro/examples/recipes/BUILD
+++ b/tensorflow/lite/micro/examples/recipes/BUILD
@@ -8,6 +8,7 @@ py_library(
     name = "resource_variables_lib",
     srcs = ["resource_variables_lib.py"],
     srcs_version = "PY3",
+    visibility = ["//visibility:public"],
     deps = [
         requirement("numpy"),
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/examples/recipes/BUILD
```diff
@@ -8,6 +8,7 @@ py_library(
     name = "resource_variables_lib",
     srcs = ["resource_variables_lib.py"],
     srcs_version = "PY3",
+    visibility = ["//visibility:public"],
```
advaitjain, line 11: can we make this ["//:__subpackages__"] ?

advaitjain (changes requested):

tensorflow/lite/micro/examples/recipes/BUILD
suleshahid: Done.

advaitjain (dismissed):

advaitjain (approved):

PR #2015, opened 2023-06-01
Update SELECT_V2 to match TFLM kernel code

### LINKED ISSUE #2014: SELECT_V2 Eval not using TfLiteEvalTensor
@tensorflow/micro 

The SELECT_V2 kernel has a non-compliant implementation.

1. Eval phase is not using TfLiteEvalTensor
2. The `has_low_rank_input_condition` variable is not used in SELECT_V2 and therefore should not be allocated storage in the arena
3. All code other than the registration should be in the anonymous namespace

ddavis-2015:
@tensorflow/micro

Eval method now uses TfLiteEvalTensor.

The `has_low_rank_input_condition` variable has been removed from OpData as SELECT_V2 does not use this variable.

All code outside of the registration has been placed into the anonymous namespace.

bug=fixes #2014

### DIFF
--- a/tensorflow/lite/micro/kernels/select.cc
+++ b/tensorflow/lite/micro/kernels/select.cc
@@ -24,6 +24,7 @@ limitations under the License.
 #include "tensorflow/lite/micro/micro_log.h"
 
 namespace tflite {
+namespace {
 
 constexpr int kInputTensorCondition = 0;
 constexpr int kInputTensorX = 1;
@@ -32,9 +33,6 @@ constexpr int kOutputTensor = 0;
 
 struct OpData {
   bool requires_broadcast;
-  // True if input condition is scalar or input condition has rank one and
-  // matches the first dimension of other inputs.
-  bool has_low_rank_input_condition;
 };
 
 void* SelectInit(TfLiteContext* context, const char* buffer, size_t length) {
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/kernels/select.cc
```diff
@@ -164,25 +163,20 @@ TfLiteStatus SelectEval(TfLiteContext* context, TfLiteNode* node) {
...
   }
 
-  if (data->has_low_rank_input_condition) {
-    MicroPrintf("Not yet implemented.");
-    return kTfLiteError;
-  } else if (data->requires_broadcast) {
+  if (data->requires_broadcast) {
     TF_LITE_SWITCH(input_x->type, BroadcastSelect5DSlow);
   } else {
     TF_LITE_SWITCH(input_x->type, Select);
```
advaitjain, line 169: Let's remove these macros and instead directly make the function call.

The macros hurt readability and we have moved away from them in the other TFLM kernels as well.

advaitjain (changes requested):

tensorflow/lite/micro/kernels/select.cc
ddavis-2015: done

advaitjain (approved):

PR #2018, opened 2023-06-01
Manual sync from upstream (and additional changes)

advaitjain:
Changes in strided slice from http://cl/536336953 required updates to the kernel test.

Note that the TFLM strided\_slice kernel does not support the new offset parameter from http://cl/536336953. See http://b/285340415 for additional context.

NO_CHECK_TFLITE_FILES=manual sync from upstream.
BUG=http://b/285340415


tensorflow/lite/micro/kernels/strided_slice_test.cc
```diff
@@ -289,7 +289,7 @@ TF_LITE_MICRO_TEST(In1D_BeginMask) {
   float golden[] = {1, 2, 3};
   float output_data[4];
 
-  TfLiteStridedSliceParams builtin_data = {1, 0, 0, 0, 0};
+  TfLiteStridedSliceParams builtin_data = {1, 0, 0, 0, 0, 0};
```
rascani, line 292: Should these be `false` instead of `0` since the added field `offset` is a bool?

rascani (dismissed):

tensorflow/lite/micro/kernels/strided_slice_test.cc
advaitjain: yes, I completely missed that it was a bool. will fix in a bit.

advaitjain:
ready for review again.

rascani (approved):

PR #2025, opened 2023-06-07
TFLM Model Transforms Tool for additional memory savings

suleshahid:
Introduces the `tflm_model_transforms` tool which is a last step in the model deployment journey for additional memory savings of the tflite flatbuffer binary.

BUG=[285021986](http://b/285021986)


tensorflow/lite/micro/tools/BUILD
```diff
@@ -88,3 +99,76 @@ py_binary(
...
+py_library(
+    name = "model_transforms_utils",
+    srcs = ["model_transforms_utils.py"],
+    data = [
+        ":tflite_flatbuffer_align_wrapper.so",
+    ],
+    srcs_version = "PY3",
+    visibility = [
+        ":tflm_tools",
+        # TODO(b/285021986): properly fix visibility
```
advaitjain, line 112: the bug linked from here doesn't give me additional context on what is needed for the visibility fix. If you can add a comment there about exactly what the issue with the current visibility is, and link to that specific comment from here, that would be great.

```py
#TODO(b/285021986): see http://b/285021986#comment2 for details on what is needed to fix the visibility.
```

advaitjain (changes requested):
one small nit. other than that, looks good to me.

tensorflow/lite/micro/tools/BUILD
suleshahid: Done.

advaitjain (approved):

Issue #1484, opened 2022-10-19
Create a Python distribution package


rkuester:
# Create a Python distribution package

This is a proposal for creating a TFLM Python package.

## Vision

The Python bindings of the TFLM interpreter are difficult to use in situ, from
their location in the TFLM source tree. By building a package, we allow users to
install and then use TFLM via Python in any virtual or real environment,
independently of the TFLM source tree, the same way they would use any package,
including Tensorflow and Tensorflow Lite.

While TFLM's target application is running on microcontrollers in C, of course,
running in Python on x86_64 is useful during development. Running in Python is
useful for quick prototyping and testing in its own right, and especially
powerful because it allows TFLM to be used directly alongside Tensorflow and
Tensorflow Lite within a Python environment. E.g., you can create a model with
Keras, run it through TFLM and Lite in parallel, and compare the outputs---all
from one short Python script.

Beyond the usefulness of installation from a local source tree, additional use
cases can be enabled by publishing the package to PyPI, including use from from
Colab and use on x86_64 without first building from Bazel.

## Design

Name the distribution package `tflite-micro` to match our project naming in
GitHub and elsewhere. In Python, that translates to the package name
`tflite_micro`, note the underscore.

Distribute only one package `tflite-micro` containing both code designed for
runtime like the TFLM interpreter and development-only code like tools for
manipulating models. Perhaps a reason to seperate the use cases into distinct
packages will emerge in the future, but for now, keep it simple and follow in
the footsteps of Tensorflow proper, which releases one omnibus package.

Within the source tree, add a `//python/` directory at the root and begin
refactoring our packagable Python code into it. For compatibility with the
Python tools, the directory structure underneath should reflect the physical
structure of the distributable package---i.e., a `//python/tflite_micro/`
directory, representing our [implict namespace
package](https://peps.python.org/pep-0420/) `tflite_micro`, containing
subpackages and modules within, such as `//python/tflite_micro/runtime.py`.

        └── python/
            └── tflite_micro/
                ├── examples/
                │   └── hello.py
                ├── runtime.py     # containing class Interpreter
                └── signal/        # pre- and post-porcessing layers
                    └── rfft.py
    
Where it makes more sense to locate distributable Python code elsewhere in the
source tree, e.g., in proximity to C code it wraps, the Python code can be
collected underneath the packaging directory at build time.

This organization leads to names in Python such as `tflite_micro.runtime.Interpreter`.

## Publishing to PyPI

[...]

PR #2028, opened 2023-06-08
build: remove intermediate pybind_library target

rkuester:
Remove the pybind_library() used only as a dependency to the
pybind_extension(), and instead simply make the pybind_extension() directly. A
pybind_library() may be useful if used in other palces in addition to a
pybind_extension(), but with no such use in our tree, it just adds
complication.

BUG=part of #1484


advaitjain (approved):
Approving and will then test to confirm that this doesn't trip anything up internally.

PR #2030, opened 2023-06-08
python: rename package tflm_runtime to runtime

rkuester:
Rename the Python package `tflm_runtime` to simply `runtime` in preparation for adding it under the new namespace package `tflite_micro`. Its full name will then be `tflite_micro.runtime`.

We have kept the `tflm_runtime` target as an alias in order to stage this change. More details in http://b/286456378

BUG=part of #1484


advaitjain (approved):

advaitjain (commented):
this will need to be staged since there is code outside of the current repository that depend on the tflm_runtime target. Let me think about how we would do that,

advaitjain:
> this will need to be staged since there is code outside of the current repository that depend on the tflm_runtime target. Let me think about how we would do that,

Here is the change to continue to keep a `tflm_runtime` target: https://github.com/tensorflow/tflite-micro/pull/2030/files/7740c965a46dcb4644aceddf890ea66dfb039f16..d5d9e032afcb7efc9f9eb3ea5e5013d9b789311e

advaitjain (approved):

PR #2044, opened 2023-06-13
Update KISSFFT Bazel version to match Makefile version

suleshahid:
In order to maintain a single version for our Makefile and Bazel builds, we should use the same version of KISSFFT. In addition, the Makefile version is properly setup to support multiple resolutions with the patch, so we apply that to the Bazel version as well.

BUG=[201319430](http://b/201319430)


third_party/kissfft/kissfft.patch
suleshahid, line 1: You can technically also use the original version at tensorflow/lite/micro/tools/make/kissfft.patch, but we need to add a BUILD file to recognize that directory as a path.

advaitjain: Let's avoid two copies of the patch file. How about we move the patch file to third_party/kisssfft/kissfft.patch and then use it from the Makefile as well.

advaitjain (changes requested):

third_party/kissfft/kissfft.patch
suleshahid: Done.

tensorflow/lite/micro/tools/make/kissfft_download.sh
```diff
@@ -59,7 +59,7 @@ else
 
   pushd ${DOWNLOADED_KISSFFT_PATH} > /dev/null
   create_git_repo ./
-  apply_patch_to_folder ./ ../../kissfft.patch "TFLM patch"
+  apply_patch_to_folder ./ ../../../../../../../third_party/kissfft/kissfft.patch "TFLM patch"
```
suleshahid, line 62: Kinda awkward with the multiple ../? I tried with the ${2} root but it didn't work.

advaitjain: yeah, that is just what it is right now.

advaitjain (approved):

PR #2046, opened 2023-06-14
Codegenerator for CWrapper API, Adds back in generate_<project_name>_make_project, adds additional target platform makefiles

cdknorow:


cdknorow:
Just testing the CI pipelines. I will break this up and add more descriptions to it for a better PR

PR #2053, opened 2023-06-14
Sync from upstream TF.

advaitjain:
Steps:
 * modified ci/tflite_files.txt
 * ```./ci/sync_from_upstream_tf.sh```
 * additional changes needed to fix the build

There are some assumptions that we had been making that have been
somewhat broken by the changes from http://cl/539690111

More details to follow. This PR is currently to test that all the TFLM
builds are green.

BUG=http://b/287269351
NO_CHECK_TFLITE_FILES=manual sync + fixes


tensorflow/lite/core/macros.h
```diff
@@ -65,4 +65,14 @@ limitations under the License.
 #define TFLITE_HAS_ATTRIBUTE_WEAK 0
 #endif
 
+#ifndef TF_LITE_STATIC_MEMORY
```
rascani, line 68: Do we use this flag to gate any differences between TFLite & TFLM? The naming is kind of non-intuitive in this context, as its guarding static data.

tensorflow/lite/kernels/kernel_util.cc
rascani, line 1: From the changes to this file, is the process to basically fix the sync for TFLM then send the changes upstream to TF before the next sync?

tensorflow/lite/micro/build_def.bzl
```diff
@@ -4,6 +4,7 @@ def micro_copts():
         "-Werror",
         "-Wnon-virtual-dtor",
         "-DFLATBUFFERS_LOCALE_INDEPENDENT=0",
+        "-std=c++17",
```
rascani, line 7: Did you mean to leave this in here? I thought we needed to stick at C++11.

tensorflow/lite/core/macros.h
advaitjain: it is indeed confusing. http://b/145762662 and http://b/160955687 has some historical context.

rascani (approved):

PR #2067, opened 2023-06-19
refactor(runtime.py): count resource variables in native code

rkuester:
Move the counting of resource variables from the Python code to the native code
of the soon-to-be-public Python module tflm_micro.runtime. This avoids what
would be a large, complex dependency in the public module on the third-party
packages tensorflow, tflite-runtime, and flatbuffers (or vendored parts
thereof).

This code is covered by the existing tests.

BUG=part of #1484

### DIFF
--- a/python/tflite_micro/BUILD
+++ b/python/tflite_micro/BUILD
@@ -75,7 +75,6 @@ py_library(
     visibility = ["//visibility:public"],
     deps = [
         requirement("numpy"),
-        "//tensorflow/lite/tools:flatbuffer_utils",
     ],
 )
 
--- a/python/tflite_micro/_runtime.cc
+++ b/python/tflite_micro/_runtime.cc
@@ -28,10 +28,9 @@ PYBIND11_MODULE(_runtime, m) {
   py::class_<InterpreterWrapper>(m, "InterpreterWrapper")
       .def(py::init([](const py::bytes& data,
                        const std::vector<std::string>& registerers_by_name,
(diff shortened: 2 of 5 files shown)


advaitjain (commented):
We will want to do flatbuffer manipulations from the tflm pypi package so I'm assuming that the dependencies being removed here will still creep in. 

If at all possible, I would like to keep the interpreter wrapper thin.

I think it's worth a chat to understand how big an issue you expect these to be.

rkuester:
> We will want to do flatbuffer manipulations from the tflm pypi package so I'm assuming that the dependencies being removed here will still creep in.

I don't foresee a problem with depending on tensorflow and lite if we must.

As originally implemented, we would need to package up some of the lite code we vendor in from upstream (which I presume is already published in the `tflite-runtime` package) and declare a runtime dependency on the `tensorflow` package. That should work, it was just a complicated way to implement this particular, small detail (counting the number of resource variables).

I agree with goal of keeping things thin. My sense of thin here was: keep these Python bindings (including the Python dependencies they drag in) small and delegate down to our native code ASAP. True I added a little code to the native part of the bindings, but I traded that off as better than dragging in all that stuff at the Python layer.

I hear what you're saying—maybe the bigger picture of other things we'll want to do in the Python layer make the dependencies unavoidable and reverse the tradeoff I was trying to make.

advaitjain:
Thanks, I better understand where you were coming from. We will be doing a lot more in Python so following the same pattern here would be the way to go from my perspective.

PR #2076, opened 2023-06-22
Download the CMSIS-DSP source code from the Github repo

shlmregev:
As part of the cmsis_nn optimized external library. Also added some #define options for RFFT code.

BUG=288422234


tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
```diff
@@ -30,6 +30,14 @@ ifeq ($(CMSIS_NN_PATH), $(CMSIS_NN_DEFAULT_DOWNLOAD_PATH))
     $(error Something went wrong with the CMSIS-NN download: $(DOWNLOAD_RESULT))
   endif
 endif
+CMSIS_DSP_DEFAULT_DOWNLOAD_PATH := $(DOWNLOADS_DIR)/cmsis_dsp
```
advaitjain, line 33: nit: add newline above

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
```diff
@@ -56,5 +72,28 @@ THIRD_PARTY_CC_HDRS += \
...
+CMSIS_NN_FLAGS := \
+  -DMIN_RFFT_Q15_LEN=$(MIN_RFFT_Q15_LEN) \
+  -DMAX_RFFT_Q15_LEN=$(MAX_RFFT_Q15_LEN) \
+  -DMIN_RFFT_Q31_LEN=$(MIN_RFFT_Q31_LEN) \
+  -DMAX_RFFT_Q31_LEN=$(MAX_RFFT_Q31_LEN) \
+  -DMIN_RFFT_FLOAT_LEN=$(MIN_RFFT_FLOAT_LEN) \
+  -DMAX_RFFT_FLOAT_LEN=$(MAX_RFFT_FLOAT_LEN)
+
+CCFLAGS += $(CMSIS_NN_FLAGS)
+CXXFLAGS += $(CMSIS_NN_FLAGS)
```
advaitjain, line 99: I'm lookng at https://github.com/ARM-software/CMSIS-DSP to see what the different compiler flags are.

Do you have any profiling information that tells us the impact of `-Ofast`, `-O2`, `-DARM_MATH_LOOPUNROLL` etc. also, is it worth disabling f16 (`-DDISABLEFLOAT16`)?

advaitjain (changes requested):
@mansnils, this is a first step towards using optimized implementations for https://github.com/tensorflow/tflite-micro/blob/main/signal/micro/kernels/rfft.cc

also tagging @suleshahid, @rascani

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
shlmregev: I haven't experimented with flags.

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
```diff
@@ -38,12 +46,20 @@ else
 endif
 THIRD_PARTY_CC_HDRS += $(shell find $(CMSIS_NN_PATH)/Include -name "*.h")
 
+ifeq (,$(CMSIS_DSP_LIBS))
+  THIRD_PARTY_KERNEL_CC_SRCS += $(shell find $(CMSIS_DSP_PATH)/Source -name "arm_*.c")
```
rascani, line 50: Not sure if this matters, but this will also grab all the f16 sources as well, which would not be gated on `-DDISABLEFLOAT16`. It might be a safer bet to just list the SRCS from here: https://github.com/ARM-software/CMSIS-DSP/blob/main/Source/Makefile#L35

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
shlmregev: Done.

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
shlmregev: Is there a way for us to specify flags that will affect cmsis-dsp sources but not other sources?

rascani: We have a way for specifying the optimization level for third_party kernel sources:
https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/tools/make/Makefile#L706

It might be good to extend that with a new flag for `THIRD_PARTY_KERNEL_CCFLAGS`.

shlmregev: I'm hesitant to define -Ofast, -O2 even for third party kernels. If possible, I'd like to land this PR with the default flags (-O2).

mansnils: These are the recommended options to use: https://github.com/ARM-software/CMSIS-DSP#options-to-use
So for performance:
-Ofast 
-DLOOPUNROLL=ON
If you go with the default for kernels -O2 there will probably not be a huge performance diff. 
If using GCC and Helium (e.g. Arm(R) Cortex(R)-M55) this will be faster but may require some additional tweaks to get integrated: https://github.com/ARM-software/EndpointAI/tree/master/Kernels/ARM-Optimized/DSP/Source
About F16, it can be useful for Helium: https://github.com/ARM-software/CMSIS-DSP#half-float-support

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
mansnils: Perhaps it would makes sense to create a new file cmsis_dsp.inc?

mansnils: Ok, I guess it is because we can only specify one optimized kernel dir e.g. OPTIMIZED_KERNELS=cmsis_nn..

shlmregev: That's correct.

tensorflow/lite/micro/tools/make/ext_libs/cmsis_nn.inc
shlmregev: Thanks @mansnils. @advaitjain, how should we proceed? Regarding -Ofast, since there's not much difference from -O2, I'd rather not s/-O2/-Ofast for fear of breaking/bloating the rest of the code. Ideally, we should have flags that are specific to each external library, but since we don't, I'd rather punt. 
Regarding f16 and loop unroll, we can control them with make variables.

PR #2086, opened 2023-06-26
IF kernel Eval not using TfLiteEvalTensor

### LINKED ISSUE #2043: IF kernel Eval not using TfLiteEvalTensor
@tensorflow/micro 

The IF kernel has a non-compliant implementation.

1. Eval phase is not using TfLiteEvalTensor
2. The `cond_value` variable should be declared `const`

ddavis-2015:
@tensorflow/micro

The IF kernel has a non-compliant implementation. Eval phase is not using TfLiteEvalTensor.

The `cond_value` variable should be declared `const`

bug=fixes #2043

### DIFF
--- a/tensorflow/lite/micro/kernels/if.cc
+++ b/tensorflow/lite/micro/kernels/if.cc
@@ -1,4 +1,4 @@
-/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2023 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -31,6 +31,8 @@ namespace tflite {
 
 namespace {
 
+constexpr int kCondTensor = 0;
+
 struct OpData {
   int then_subgraph_index;
   int else_subgraph_index;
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/kernels/if.cc
```diff
@@ -86,11 +89,10 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
   const OpData* op_data = reinterpret_cast<OpData*>(node->user_data);
 
   tflite::MicroContext* micro_context = tflite::GetMicroContext(context);
-  TfLiteTensor* cond = micro_context->AllocateTempInputTensor(node, 0);
-
+  const TfLiteEvalTensor* cond =
+      tflite::micro::GetEvalInput(context, node, kCondTensor);
   TF_LITE_ENSURE(context, cond != nullptr);
-  bool cond_value = cond->data.b[0];
-  micro_context->DeallocateTempTfLiteTensor(cond);
```
rascani, line 93: Don't we still need to deallocate the `cond` Tensor that was allocated?

ddavis-2015: TfLiteEvalTensor is a permanent allocation that exists for the lifetime of the MicroInterpreter.

rascani: Ah sorry, I misread. I got it confused with the `cond` allocated at line 58.

rascani (approved):

PR #2095, opened 2023-06-30
Update docker containers for Xtensa workflows.

advaitjain:
As part of reducing the size of the docker containers for the Xtensa tests, we have split them up based on different versions of the toolchain.

This PR changes the github workflows to use the smaller and more targeted docker containers.

Couple of things to note:
 * This PR is not pinning to a specific version of the docker container. This will make it easier to push updates if needed.
 * the names of the containers might change again. We want to get somethign going quickly so that CI is green again and PRs can be merged.
 * This PR will be merged bypassing the branch protection rules since the Xtensa workflows run as pull_request_target. We will only be able to test the changes in this PR after the merge.

BUG=http://b/289098887


advaitjain:
Tagging @suleshahid

PR #2104, opened 2023-07-05
sync

turbotoribio:


signal/micro/kernels/framer.cc
```diff
@@ -0,0 +1,192 @@
...
+  params->n_frames = innermost_dim / params->frame_step;
+
+  params->state_buffers =
+      static_cast<int8_t**>(context->AllocatePersistentBuffer(
+          context, params->outer_dims * sizeof(int8_t*)));
+  params->circular_buffers = static_cast<tflm_signal::CircularBuffer**>(
+      context->AllocatePersistentBuffer(
+          context, params->outer_dims * sizeof(tflm_signal::CircularBuffer*)));
+  for (int i = 0; i < params->outer_dims; i++) {
+    // Calculate the capcity of the circular buffer. Round up the frame size to
```
rascani, line 114: Typo: capacity*

signal/micro/kernels/framer_test.cc
```diff
@@ -0,0 +1,249 @@
...
+==============================================================================*/
+#include <cstdint>
+#include <cstring>
+
+#include "signal/micro/kernels/framer_flexbuffers_generated_data.h"
+#include "tensorflow/lite/micro/kernels/kernel_runner.h"
+#include "tensorflow/lite/micro/test_helpers.h"
+#include "tensorflow/lite/micro/testing/micro_test.h"
+
+namespace {
```
rascani, line 20: We should probably have this in the tflite namespace so we don't have to specify that everywhere.
```
namespace tflite {
namespace {
```

signal/src/circular_buffer.h
```diff
@@ -0,0 +1,119 @@
...
+==============================================================================*/
+
+#ifndef SIGNAL_SRC_CIRCULAR_BUFFER_H_
+#define SIGNAL_SRC_CIRCULAR_BUFFER_H_
+
+#include <stddef.h>
+#include <stdint.h>
+
+namespace tflm_signal {
+// TODO(b/286250473): remove namespace once de-duped libraries above
```
rascani, line 23: We should probably still put this in a namespace. Does the naming conflict with the CircularBuffer op? We might want to consider renaming this.

signal/src/circular_buffer.h
```diff
@@ -0,0 +1,119 @@
...
+// Returns the size of the memory that the circular buffer needs
+// in order to hold `capacity` items.
+size_t CircularBufferGetNeededMemory(size_t capacity);
+
+// Initialize an instance of the circular buffer that holds `capacity` items.
+// `state` points to a memory allocation of size `state_size`. The size
+//  should be greater or equal to the value returned by
+//  CircularBufferGetNeededMemory(capacity). Fails if it isn't.
+//  On success, returns a pointer to the circular buffer's object.
+CircularBuffer *CircularBufferInit(size_t capacity,
```
rascani, line 48: Pointer should go next to the type:

```
CircularBuffer* CircularBufferInit(size_t capacity,
                                   void* state,
                                   size_t state_size);
```

signal/src/circular_buffer.cc
```diff
@@ -0,0 +1,290 @@
...
+limitations under the License.
+==============================================================================*/
+
+#include "signal/src/circular_buffer.h"
+
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define ASSERT assert
```
rascani, line 22: Any reason we can't use the TFLITE_DCHECK macros for asserts here?

rascani (commented):
Looks good, just a few minor comments.

signal/src/circular_buffer.cc
turbotoribio: no idea , all of this was written by Speech Team from my understanding they might have a reason I'm not sure of. Probably good question for Shlomi. I'd prefer to not change this till all the code is ported and we can see if it breaks anyone.

signal/src/circular_buffer.h
turbotoribio: It doesn't conflict with CircularBuffer , I believe this has Audio specific stuff in it and isn't an op. It only clashes with the same file in AudioFrontend, would prefer not renaming anything till everything is ported over and we swap users to this package.

signal/micro/kernels/framer_test.cc
turbotoribio: done !

signal/micro/kernels/framer.cc
turbotoribio: leftover typo from original file , fixed !

signal/micro/kernels/framer.cc
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+}
+
+void* Init(TfLiteContext* context, const char* buffer, size_t length) {
+  const uint8_t* buffer_t = reinterpret_cast<const uint8_t*>(buffer);
+
+  auto* params =
+      static_cast<TFLMSignalFramerParams*>(context->AllocatePersistentBuffer(
+          context, sizeof(TFLMSignalFramerParams)));
```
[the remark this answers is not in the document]
turbotoribio: added a dcheck but not sure if this good/the best way to do this .

signal/micro/kernels/framer_test.cc
turbotoribio: done !

signal/micro/kernels/framer.cc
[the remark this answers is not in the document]
rascani: Other kernels seem to simply return the result of `AllocatePersistentBuffer`. The difference here is that we need to populate a couple of the fields, but I'd generally prefer to stick with that same pattern of returning nullptr if `AllocatePersistentBuffer` fails. So I'd suggest just a nullptr check with early return:

```
if (params == nullptr) {
  return nullptr;
}
```

signal/micro/kernels/framer_test.cc
```diff
@@ -160,6 +160,7 @@ void TestFramerReset(int* input_dims_data, int16_t* input_data,
 }
 
 }  // namespace
+}  // namespace tflite
```
rascani, line 163: Do `TF_LITE_MICRO_TEST`s need to be in the global namespace? Can this just go all the way at the end of the file?

signal/src/circular_buffer.h
rascani: I'm fine with avoiding renaming, but I think we should put this in the `tflite` namespace like everything else. Temporarily `tflite::tflm_signal` and eventually `tflite`.

signal/micro/kernels/framer_test.cc
turbotoribio: TF_LITE_MICRO_TEST is just a macro for main(){} and all of our test are formatted like that  currently.

rascani (approved):

PR #2114, opened 2023-07-11
Fix formatting in lstm_eval_hifi.cc

rascani:
The lack of whitespace around bitwise & operations started failing clang-format checks. The code didn't change, so presumably this was on account of clang-format updates.

BUG=formatcleanup


rascani:
Looks like there is a disagreement between local machine & CI. We should stick with what is in CI.

PR #2127, opened 2023-07-14
Explicit load for py_library.

advaitjain:
BUG=http://b/291306662


python/tflite_micro/signal/tflm_signal.bzl
```diff
@@ -1,5 +1,7 @@
 """Build rule for wrapping a custom TF OP from .cc to python."""
 
+load("@rules_python//python:defs.bzl", "py_library")
```
rascani, line 3: Do we also need to remove the `native.` from the `py_library` usage down at line 64?

advaitjain: thanks. fixed.

rascani (approved):

PR #2128, opened 2023-07-14
Add RISC-V as a requirement for the issue on error.

advaitjain:
This will (hopefully) ensure that RISC-V failures result in a github issue being created.

BUG=http://b/291335234


rascani (approved):

advaitjain:
merged without waiting for checks since this workflow file is pull_request_target and can not be tested until it it merged.

advaitjain:
The post_tests workflow from this PR appears to confirm that the change is correct: https://github.com/tensorflow/tflite-micro/actions/runs/5558888228?pr=2128


image

advaitjain:
Prior to this change, the graph looked like: 
image

PR #2134, opened 2023-07-18
Adds FFT Auto Scale Op

suleshahid:
This PR adds additional FFT op functionality in the Signal library, namely adding the FFT Auto Scale operation.
Testing added in the original `fft_test.cc` and `fft_ops_test.py`.
BUG=[287346710](http://b/287346710)


python/tflite_micro/signal/ops/fft_ops_test.py
```diff
@@ -33,6 +33,31 @@ def GetResource(self, filepath):
...
+    input_size = len(lines[0].split())
+    concrete_function = func.get_concrete_function(
+        tf.TensorSpec(input_size, dtype=tf.int16))
+    interpreter = util.get_tflm_interpreter(concrete_function, func)
+    i = 0
+    while i < len(lines):
+      in_frame = np.array([int(j) for j in lines[i].split()], dtype=np.int16)
+      out_frame_exp = [int(j) for j in lines[i + 1].split()]
+      scale_exp = [int(j) for j in lines[i + 2].split()]
+      # TFLite
```
rascani, line 48: TFLM?

signal/src/fft_auto_scale.cc
```diff
@@ -0,0 +1,40 @@
...
+#include "signal/src/fft_auto_scale.h"
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include "signal/src/max_abs.h"
+#include "signal/src/msb.h"
+
+// TODO(b/286250473): remove namespace once de-duped libraries
+namespace tflm_signal {
```
rascani, line 25: Can we put this in `tflite::tflm_signal`? When we remove the namespace, we don't want the symbol to wind up in the global namespace.

signal/src/fft_auto_scale.cc
```diff
@@ -0,0 +1,40 @@
...
+
+int FftAutoScale(const int16_t *input, int size, int16_t *output) {
+  const int16_t max = tflm_signal::MaxAbs16(input, size);
```
rascani, line 28: The `tflm_signal::` prefix shouldn't be necessary here or line 29, as we're already in that namespace.

signal/src/max_abs.cc
```diff
@@ -0,0 +1,33 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "signal/src/max_abs.h"
+
+// TODO(b/286250473): remove namespace once de-duped libraries
+namespace tflm_signal {
```
rascani, line 19: Same comment here about putting it in the `tflite::tflm_signal` namespace.

python/tflite_micro/signal/ops/fft_ops_test.py
suleshahid: Done. Changed to TFLM, but we have a TODO elsewhere to revisit these tests in general when we get the chance.

signal/src/fft_auto_scale.cc
suleshahid: Done. I wonder if we can remove the tflm_signal anyways then, or just reword it to tflite when we de-dupe?

signal/src/fft_auto_scale.cc
suleshahid: Done.

signal/src/max_abs.cc
suleshahid: Done.

PR #2140, opened 2023-07-20
Add Q16x8 Depthwise Conv Support

mbrooksx:
Adds support for 16-bit activations + 8-bit weights for depthwise convolution in the reference kernel. Uses 64-bit bias to match TFLite. Also adds passthrough to the q16x8 reference kernel for Xtensa, CEVA, and ARC (CMSIS already has it's own implementation).

Tested:
depthwise_conv_test

BUG=2141


tensorflow/lite/micro/kernels/xtensa/depthwise_conv.cc
```diff
@@ -48,6 +48,16 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
 
 TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
   TF_LITE_ENSURE_OK(context, DepthwiseConvPrepare(context, node));
+  // Use only the default depthwise convolution for int16 input.
```
rascani, line 51: Can we expand this comment a bit to explain why we only need to do the default for int16? Specifically, we should mention that we fall back to the reference kernels for int16, so there's no need to call the Xtensa-specific Prepare methods.

rascani (commented):
LGTM, just one minor comment clarification and will approve.

tensorflow/lite/micro/kernels/xtensa/depthwise_conv.cc
mbrooksx: Thanks for the review. Updated. I did once again forget the bug on this commit but it should work overall with a squash commit (with the top-level message above).

rascani (approved):

PR #2151, opened 2023-07-31
feat: build and test a Python distribution package `tflite_micro`

rkuester:
Add the build configuration and integrated test to generate a Python
distribution package named `tflite_micro` for publishing the tflm interpreter
as a Python module with a native extension.

Use the build tools provided in @rules_python, augmented by a custom rule
`py_namespace` for the reasons documented in `python/py_namespace.bzl`.

Provide an integration test at `//python/tflite_micro:whl_test`. Use a .tflite
model copied from the hello_world example. (Copied to avoid creating a
dependency.)

BUG=part of #1484


python/py_namespace.bzl
```diff
@@ -0,0 +1,141 @@
...
+
+The stock @rules_python py_library() -> py_package() -> py_wheel() BUILD file
+workflow packages files at Python package paths set to the source paths of the
+files relative to the workspace root. This has a several problems. Firstly, it
+implies that files must be located underneath a source directory with the same
+name as the desired Python namespace package. ( py_wheel.strip_path_prefixes
+can remove path components, but cannot add them.) This is not always feasible
+or desirable.
+
+Secondly, this path naming is maddingly incompatible with the PYTHONPATH set by
```
advaitjain, line 12: nit: "this path naming is maddingly incompatible" --> "this path naming is incompatible"

advaitjain (commented):
lgtm with a small nit.

python/py_namespace.bzl
rkuester: Fixed in https://github.com/tensorflow/tflite-micro/pull/2151/commits/b6f04b387ad7466cb50d5463ceb6ce35ed713c7a.

advaitjain (approved):

PR #2155, opened 2023-08-07
Fix armclang build issue

mansnils:
BUG=armclang has build issues with https://github.com/eyalroz/printf


rascani (approved):

mansnils:
Actually we prefer to use the standard library here for the cortex_m_generic target. The new third party library cause some additional linking issue with armclang (sorry for not spotting this earlier). It can be worked around but I think we can go with the standard library implementation of vsnprintf here. So will update the PR.

PR #2162, opened 2023-08-10
Add code generator experiment

rascani:
This PR creates the initial code generator scaffolding for performing inference without an interpreter. Currently, this does nothing other create a header and source file from Mako templates. Mako was chosen as a template engine due to existing dependency.

BUG=b/295076487


tensorflow/lite/micro/tools/codegen/inference_generator.py
```diff
@@ -0,0 +1,60 @@
...
+
+def _generate_source(source_path: pathlib.Path, model_data: ModelData) -> None:
+  _render(source_path, _SOURCE_TEMPLATE, model_data)
+
+
+def generate(output_dir: str, output_name: str) -> None:
+  """ Generate C/C++ inference code. """
+  header_file = f"{output_name}.h"
+  model_data: ModelData = {
+      'header_file': header_file,
```
advaitjain, line 51: is this needed?

advaitjain (changes requested):
What do you think of making a top-level codegen directory instead of tensorflow/lite/micro/tools/codegen?

tensorflow/lite/micro/tools/codegen/inference_generator.py
rascani: Yeah, the generated source file will need to know where it's corresponding header is. Probably could have inferred it in the template from the model name, but that seemed like it could drift if we start including paths.

rascani:
Top level codegen directory sounds good to me, lot less typing. :)

tensorflow/lite/micro/tools/codegen/inference_generator.py
rascani: Added header_file to the ModelData TypedDict

advaitjain (approved):

PR #2165, opened 2023-08-11
Remove unused and combine similar loads

rascani:
BUG=cleanup


tensorflow/lite/micro/examples/person_detection/BUILD
```diff
@@ -1,6 +1,5 @@
 # Description:
 #   TensorFlow Lite for Microcontrollers Vision Example.
-load("@rules_python//python:defs.bzl", "py_binary", "py_library", "py_test")
```
turbotoribio, line 3: I believe previously https://critique.corp.google.com/cl/556860566 had these changes as well, but they have already removed  it so I can't double check . Just making a note of it if stuff breaks in the future related to person_detect

turbotoribio (approved):

PR #2167, opened 2023-08-14
fix(py): remove leftover debugging print()

rkuester:
Remove a leftover print() presumably added for debugging purposes. It prints
the number of resource variables used in a model whenever a TFLM interpreter is
constructed, causing unwanted output in Python programs using the interpreter
(including those using the Python distribution package).

BUG=see description

### DIFF
--- a/python/tflite_micro/README.md
+++ b/python/tflite_micro/README.md
@@ -128,8 +128,9 @@ TFLite's custom op support.
 
 ## Print Allocations
 
-The Python interpreter can also be used to print memory arena allocations. This
-is very helpful to figure out actual memory arena usage.
+The Python interpreter can also be used to print memory arena allocations and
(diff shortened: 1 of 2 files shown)


advaitjain:
You're correct about this being an undesired output in a lot of cases. Let's still keep it for now.

The direction that we want to go in is to print out all the model-specific parameters from Python that can then be used when a user writes some C++ code for inference. We have also considered going down the path of generating some C++ code for the `MicroMutableOpResolver`, number of resource variables ... Not sure exactly what that API should look like right now.

@rascani can help figure out when we might want to put some thought into this.

Some more context: https://github.com/tensorflow/tflite-micro/pull/1738 and http://b/251851084

rkuester:
How do you feel about 862df7a, moving it to print_allocations()? @suleshahid @advaitjain

PR #2168, opened 2023-08-14
feat(py): give package a developmental version label

rkuester:
Use Bazel's workspace status mechanism, designed for "stamping" builds with
identifying information from the build environment, to dynamically generate the
version label of the Python distribution package. Generate stamps when Bazel
runs via the --workspace_status_command option and command script. Then use
these stamps in the version label.

Guidelines for Python version labels are given in [PEP 440][]. TFLM does not
currently tag and publish what PEP 440 calls "final releases". Instead,
distribution packages will periodically be published from the tip of the main
branch, and users will be expected to use the latest version or pin to a
historical version of their choice. To facilitate such use, the build system
needs to generate a unique, ascending version label for each commit on the main
branch.

Guided by [PEP 440][], use developmental version labels of the form
*major[.minor].dev<time>*, for example:

    0.dev20090103181505

Use a release segment with a major number of 0 (via the standard Bazel
BUILD_EMBED_LABEL stamp), because TFLM does not currently tag and publish
releases. This leaves room in the version space for making semver-style final
releases in the future. Add a developmental release segment with a date and
time stamp.

Packages should be traceable to the exact source from which they were built.
[PEP 440][] does not allow Git hashes in version labels published to PyPI;
however, the Git hash can be embedded in the package's runtime-visible
attribute `tflite_micro.__version__`, and in the package's description, which
is displayed on PyPI.

[PEP 440]: https://peps.python.org/pep-0440

BUG=part of #1484


tools/expand_stamp_vars_test.py
```diff
@@ -0,0 +1,49 @@
...
+import sys
+
+if __name__ == "__main__":
+  stamps = """
+BUILD_STAMP_ONE value_one
+BUILD_STAMP_TWO value_two
+"""
+
+  input = "This is {BUILD_STAMP_TWO}. This is {BUILD_STAMP_ONE}."
+  golden = "This is value_two. This is value_one.."
```
rascani, line 32: Seems like there is an extra `.` at the end of the `golden` value.

tools/expand_stamp_vars_test.py
```diff
@@ -0,0 +1,49 @@
...
+# ----
+
+# A test for the filter that expands Bazel workspace stamp variables.
+
+from tools import expand_stamp_vars
+
+import io
+import sys
+
+if __name__ == "__main__":
```
rascani, line 25: Is it possible to write this test in the form of a python unittest?

rascani (commented):
This largely looks good to me. Just a couple of testing comments.

tools/expand_stamp_vars_test.py
rkuester: Sure thing; although as I do, please help me articulate a general approach.

The only precedent I saw for Python-language unit tests in our tree is the use of `tensorflow.python.test`, from the `tensorflow` package. For example:

https://github.com/tensorflow/tflite-micro/blob/46ccaf8fb3687d0f6b579f726400f81ab5fa7a81/python/tflite_micro/runtime_test.py#L27-L28
https://github.com/tensorflow/tflite-micro/blob/46ccaf8fb3687d0f6b579f726400f81ab5fa7a81/python/tflite_micro/runtime_test.py#L33-L34
https://github.com/tensorflow/tflite-micro/blob/46ccaf8fb3687d0f6b579f726400f81ab5fa7a81/python/tflite_micro/runtime_test.py#L268-L269

The layers here are something like:

1. `tensorflow.python.test` [wraps](https://github.com/tensorflow/tensorflow/blob/f52dc5dafc43fabc594f1c1bf744c0ae585ee4bc/tensorflow/python/platform/test.py#L24)
2. `absl.testing.absltest` ([via](https://github.com/tensorflow/tensorflow/blob/f52dc5dafc43fabc594f1c1bf744c0ae585ee4bc/tensorflow/python/platform/googletest.py#L25) the name `testing.pybase.googletest`), which [wraps](https://github.com/abseil/abseil-py/blob/976413320682f8fd1c05c36f24a9475050702699/absl/testing/absltest.py#L564)
3. the standard library's `unittest`. All run as a
4. Bazel `py_test()`, which like all Bazel tests, simply expects an executable that returns success or failure.

Because what I'm testing here is just a scrap of generic build tooling, it didn't seem right to intertwine a heavy dependency like `tensorflow`. It seemed so lite, in fact—it's trivially short, I'm not testing multiple cases, I'm not aware of any time we run things like this from outside Bazel (where it might benefit from the discoverability of `unittest`), etc.—I just jumped in at no. 4 there, the lowest layer, and wrote a straightforward script that returns success or failure.

I imagine the rationale for using `unittest` being: it's standard so it will cue the reader that this is a test—even though I can't think of any features of `unittest` from which it will benefit. I myself have some room for low-level, Bazel-ly test executables that simply pass or fail, like I see in other Bazel-adjacent projects like [bazelbuild/rules_python](https://github.com/bazelbuild/rules_python/tree/main).

Would there be reasons for going even higher up the stack? If one accepts the argument for `unittest`, I could imagine an argument for `absltest` as The Google Way to do `unittest`. I strongly dislike the idea of bringing in `tensorflow` for this. Again, remember this is just a small, stand-alone test of some very Bazel-centric glue code.

Maybe there are some motivations or requirements I don't see?

rkuester: Refactored to use `unittest` in dd643ed5.

rascani: The motivation is commonality. Ideally, tests are written in one framework so that developers don't have to be familiar with more than one. Perhaps I should have suggested `tensorflow.python.test` to get closer to the norm (I'm still new around here). While I agree this is a pretty small test and that adding dependencies should be done thoughtfully, there is still value in commonality.

rascani (approved):

PR #2172, opened 2023-08-16
build(py): add target that publishes to PyPI

rkuester:
Add a Bazel target `//python/tflite_micro:whl.publish` that publishes the
Python distribution package to PyPI. Require an authorization token in
the environment. See code comments for usage.

BUG=part of #1484


rkuester:
@rascani, it looks like `bazel run` only allows running a single target[^1], so there's no need to hide the publishing target from being caught by a wildcard.

[^1]: `bazel run ...` gives the message "ERROR: Only a single target can be run. [....]"

rascani (approved):

PR #2176, opened 2023-08-17
Generate op table and subgraph invoke functions

rascani:
As the next step in the codegen experiment, we want to generate the invoke calls for each layer. This is slightly challenging with the existing sources, as kernels only expose a registration function, not their individual Eval functions. In an effort to keep the code churn to a minimum, this PR introduces an inference only registration structure and function. It includes just two function pointers: invoke and reset. For this CL, we've only introduced it for FullyConnected.

In the code generator, this PR creates a new op_table array in the generated source, with an enum for lookup. It also generates an invoke function for each subgraph, that calls each operator's invoke function.

BUG=295174388


advaitjain (dismissed):

advaitjain (approved):

rascani:
@Mergifyio refresh

PR #2186, opened 2023-08-24
Break dependency of kernels on MicroGraph

rascani:
The control flow and resource kernels make calls into MicroGraph by using the MicroContext::graph() accessor. For interpreter-less invoke, we don't need a full MicroGraph, only the subset of functionality that the kernels use. This PR makes MicroGraph an abstract base class, which MicroInterpreterGraph now implements. Existing MicroGraph functionality was all moved to MicroInterpreterGraph. The kernels only depend on MicroGraph, but the MicroContext will still have a full MicroInterpreterGraph.

BUG=b/295174086


tensorflow/lite/micro/mock_micro_graph.h
```diff
@@ -18,15 +18,15 @@ limitations under the License.
...
-#include "tensorflow/lite/micro/micro_graph.h"
+#include "tensorflow/lite/micro/micro_interpreter_graph.h"
 #include "tensorflow/lite/schema/schema_generated.h"
 
 namespace tflite {
 
 // MockMicroGraph stubs out all MicroGraph methods used during invoke. A count
 // of the number of calls to invoke for each subgraph is maintained for
 // validation of control flow operators.
-class MockMicroGraph : public MicroGraph {
```
rascani, line 29: I'd like to fix this to be dependent on the abstract type. I'm gonna try to fix that before landing.

rascani:
Closing in favor of #2202

PR #2193, opened 2023-08-28
Generate static TfLiteEvalTensors

rascani:
This PR generates all of the TfLiteEvalTensors for the graph. It also generates all of the static buffers used for tensors that were present in the flatbuffer.

BUG=b/295077140


codegen/utils.py
```diff
@@ -30,6 +31,17 @@ def bool_to_c_str(b: bool) -> str:
...
 
+def split_into_chunks(
+    data: Iterable[Any],
+    chunk_size: int) -> Generator[Tuple[Any, ...], None, None]:
+  """Splits an iterable into chunks of a given size."""
+  data_iterator = iter(data)
+  chunk = tuple(itertools.islice(data_iterator, chunk_size))
+  while chunk:
+    yield chunk
+    chunk = tuple(itertools.islice(data_iterator, chunk_size))
```
advaitjain, line 42: this reads a bit funny.

Is there any way to write this where the loop looks something like:
```py
while <not sure what the condition would be>:
    chunk = tuple(itertools.islice(data_iterator, chunk_size))
    yield chunk
```

advaitjain (changes requested):
one nit question. Other than that lgtm.

codegen/utils.py
rascani: WDYT of this:

```
while True:
  chunk = tuple(itertools.islice(data_iterator, chunk_size))
  if not chunk:
    break
  yield chunk
```

advaitjain: +1. that seems  more readable to me.

rascani: Done

advaitjain (approved):

PR #2197, opened 2023-08-29
Update ethos_u_core_platform_download.sh to accommodate larger models

mansnils:
BUG=https://github.com/tensorflow/tflite-micro/issues/2189

### DIFF
--- a/tensorflow/lite/micro/tools/make/ethos_u_core_platform_download.sh
+++ b/tensorflow/lite/micro/tools/make/ethos_u_core_platform_download.sh
@@ -60,7 +60,7 @@ else
   rm -rf .git
   create_git_repo ./
 
-  apply_patch_to_folder ./ ../../increase-stack-size-and-switch-DTCM-SRAM.patch "TFLM patch"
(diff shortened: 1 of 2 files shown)


rascani (commented):
Thanks for taking a look at this @mansnils. As an alternative approach, do you see any issue with greatly increasing the SRAM segment size? Obviously it would be unrealistic on target, but we're only concerned with using it in simulation here. My concern is that we'll have to keep updating this patch to pick out individual symbols to relocate them into DDR.

mansnils:
> Thanks for taking a look at this @mansnils. As an alternative approach, do you see any issue with greatly increasing the SRAM segment size? Obviously it would be unrealistic on target, but we're only concerned with using it in simulation here. My concern is that we'll have to keep updating this patch to pick out individual symbols to relocate them into DDR.

I have no concern in increasing SRAM as long as it works. I saw your first comment a bit to late, so if you already have this kind of patch feel free to update this PR or replace it with another PR.

mansnils:
Closing in favor of: https://github.com/tensorflow/tflite-micro/pull/2201

PR #2202, opened 2023-08-30
Make MicroContext and MicroGraph abstract

rascani:
MicroContext and MicroGraph serve as the primary API surface between the interpreter and the kernels. For interpreter-less execution, we need to be able to substitute different MicroContext and MicroGraph implementations. This PR makes MicroContext and MicroGraph abstract and moves their existing implementations to MicroInterpreterContext and MicroInterpreterGraph, respectively. This will allow us to substitute an inference-only implementation of each for codegen.

FakeMicroContext and MockMicroGraph were also updated to implement the base classes.

BUG=b/295174086


suleshahid (approved):
LGTM

PR #2212, opened 2023-09-07
Create codegen_preprocessor

rascani:
This PR creates the initial scaffolding for the codegen preprocessor. The preprocessor is a target-specific binary that will take a model, load it into a TFLM Interpreter, perform the Init & Prepare stages, then serialize the resulting data structures to an output file. Currently, all this binary does is load the model file and write an output file that simply contains the source model path in it. This will be expanded as we expose the data.

BUG=b/295076056

### DIFF
--- a/codegen/BUILD
+++ b/codegen/BUILD
@@ -64,6 +64,7 @@ py_binary(
     deps = [
         ":graph",
         ":inference_generator",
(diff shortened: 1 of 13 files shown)


rascani:
Closing in favor of #2219

PR #2218, opened 2023-09-12
Implement MicroContext/Graph for codegen

rascani:
The TFLM kernels use the MicroContext and MicroGraph interfaces to fetch
eval tensors and access other parts of the graph. Since codegen does not
have the MicroInterpreter and related objects to serve those, this PR
introduces a new MicroCodegenContext class that serves as both the
MicroContext and MicroGraph.

The MicroCodegenContext is configured with a span of Subgraphs, each of
which includes the inputs, outputs, nodes, tensors and an invocation
function. The code generator will create the data and functions needed
for each Subgraph and initialize the MicroCodegenContext with it. By
having the re-usable MicroCodegenContext, the code generator won't have
to generate nearly as much code.

BUG=b/295174086


suleshahid (approved):
LGTM

PR #2222, opened 2023-09-14
Create Make helpers for running codegen

rascani:
The codegen process is a multi-step process that requires compiling,
executing code under simulation, and executing python scripts. To
simplify this workflow, this commit adds Make helper functions for
generating inference source code from a model and creating a binary with
it.

It also updates the hello world example to use these helpers and adds an
update script for keeping the checked in generated source in sync.

BUG=cleanup


turbotoribio (approved):

rascani:
@Mergifyio refresh

PR #2236, opened 2023-09-20
feat(py): add scripts for building and uploading to PyPI

rkuester:
Add mechanism and scripts for building and uploading the Python distribution
package `tflite_micro` to PyPI. These scripts are intended mainly for use by CI
when generating packages for distribution via for PyPI, and won't be used by
most developers. Building a package for local use is still done via a normal
Bazel build.

Heavily comment the scripts with rationale and technical details of the
implementation.

Make significant updates to python/tflite_micro/README.md which explain
building, installing, and uploading the package to PyPI. Leave some cleanup of
the existing text for later.

Add a build setting `--//python/tflite_micro:compatibility_tag` for setting
:whl's platform compatibility tag. Unfortunately, it cannot derived
automatically from the execution environment by the current implementation of
@rules_python.

BUG=part of #1484

### DIFF
--- a/python/tflite_micro/README.md
+++ b/python/tflite_micro/README.md
@@ -16,12 +16,12 @@ build is instead driven by its C/C++ build system Bazel.
 The Bazel target `//python/tflite_micro:whl.dist` builds a `tflite_micro`
 Python *.whl* under the output directory `bazel-bin/python/tflite_micro/whl_dist`. For example:
 ```
-% bazel build //python/tflite_micro:whl.dist                                                                   ~/wetu/tflite-micro
+% bazel build //python/tflite_micro:whl.dist
 ....
 Target //python/tflite_micro:whl.dist up-to-date:
   bazel-bin/python/tflite_micro/whl_dist
 
-% tree bazel-bin/python/tflite_micro/whl_dist                                                                ~/wetu/tflite-micro
+% tree bazel-bin/python/tflite_micro/whl_dist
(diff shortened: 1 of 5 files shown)


python/tflite_micro/README.md
```diff
@@ -1,25 +1,127 @@
...
+## Building and installing locally
 
-### Bazel
+### Building
 
-#### Build
+The Bazel target `//python/tflite_micro:whl.dist` builds a `tflite_micro`
+Python *.whl* under the output directory `bazel-bin/python/tflite_micro/whl_dist`. For example:
+```
+% bazel build //python/tflite_micro:whl.dist                                                                   ~/wetu/tflite-micro
```
rascani, line 19: Do we need that trailing path?

python/tflite_micro/README.md
```diff
@@ -1,25 +1,127 @@
...
+....
+Target //python/tflite_micro:whl.dist up-to-date:
+  bazel-bin/python/tflite_micro/whl_dist
+
+% tree bazel-bin/python/tflite_micro/whl_dist                                                                ~/wetu/tflite-micro
```
rascani, line 24: Or this one?

python/tflite_micro/README.md
rkuester: Doh, that's a copy-and-paste error I didn't catch. Will remove.

rascani (approved):

PR #2239, opened 2023-09-22
xtensa: Add BUILD file and workspace for nnlib_hifi4

afoxley:
The optimized kernel selection expects a lib target to be present in third_party/xtensa/nnlib_hifi4 so provide an alias there that maps to an external repository.

Provide a build file to be used as the root of the nnlib_hifi4 repository and add to WORKSPACE. No build is done using this lib yet as there is no xtensa toolchain available in bazel just yet.

BUG=None

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -16,6 +16,7 @@
 workspace(name = "tflite_micro")
 
 load("//tensorflow:workspace.bzl", "workspace")
+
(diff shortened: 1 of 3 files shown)


rascani (approved):

rascani:
@Mergifyio refresh

PR #2247, opened 2023-09-26
Avoid shadow parameter naming in lambda functions

cdknorow:
the lambda function has the same variable names as the function it is encapsulated by, for some compilers this is being caught as an error in the default settings for no shadow parameters.

Example:
```c++
  BroadcastSubCommon<T>(
      **params**, input1_shape, input1_data, input2_shape, input2_data,
      output_shape, output_data,
      [](T input1_val, T input2_val, const ArithmeticParams& params)
```

suggest changing to  **lambda_params** or **input_params**

```c++
  BroadcastSubCommon<T>(
      params, input1_shape, input1_data, input2_shape, input2_data,
      output_shape, output_data,
      [](T input1_val, T input2_val, const ArithmeticParams& input_params)
```

### DIFF
--- a/tensorflow/lite/kernels/internal/reference/mul.h
+++ b/tensorflow/lite/kernels/internal/reference/mul.h
@@ -147,18 +147,18 @@ inline void BroadcastMul6DSlow(const ArithmeticParams& params,
   BroadcastMulRecursiveDimensions(
       params, 0, input1_data, input2_data, output_data, &input1_offset,
       &input2_offset, &output_offset, desc1, desc2, extended_output_shape_dims,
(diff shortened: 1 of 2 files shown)


rascani:
Thanks for the PR @cdknorow. Unfortunately, the source of truth for these files is in the upstream tensorflow repository.  Would you mind re-submitting there?

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/internal/reference/sub.h
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/internal/reference/mul.h

PR #2266, opened 2023-10-09
Update op resolver generate script

mansnils:
Handle batch mat mul operator correctly.

BUG=Mutable op resolver generate script can not handle BatchMatMult.


rascani (approved):

mansnils:
@mergifyio refresh

Issue #2183, opened 2023-08-23
flatbuffers library hits overflow warning in GCC-13


hcindyl:
GCC 13 is the default in debian testing
```
INFO: From Compiling src/reflection.cpp [for host]:
In file included from /usr/include/c++/13/string:51,
                 from bazel-out/host/bin/external/flatbuffers/src/_virtual_includes/flatbuffers/flatbuffers/base.h:41,
                 from bazel-out/host/bin/external/flatbuffers/src/_virtual_includes/flatbuffers/flatbuffers/array.h:20,
                 from bazel-out/host/bin/external/flatbuffers/src/_virtual_includes/flatbuffers/flatbuffers/flatbuffers.h:22,
                 from bazel-out/host/bin/external/flatbuffers/src/_virtual_includes/flatbuffers/flatbuffers/reflection_generated.h:7,
                 from bazel-out/host/bin/external/flatbuffers/src/_virtual_includes/flatbuffers/flatbuffers/reflection.h:25,
                 from external/flatbuffers/src/reflection.cpp:17:
In static member function 'static _Up* std::__copy_move<_IsMove, true, std::random_access_iterator_tag>::__copy_m(_Tp*, _Tp*, _Up*) [with _Tp = const unsigned char; _Up = unsigned char; bool _IsMove = false]',
    inlined from '_OI std::__copy_move_a2(_II, _II, _OI) [with bool _IsMove = false; _II = const unsigned char*; _OI = unsigned char*]' at /usr/include/c++/13/bits/stl_algobase.h:506:30,
    inlined from '_OI std::__copy_move_a1(_II, _II, _OI) [with bool _IsMove = false; _II = const unsigned char*; _OI = unsigned char*]' at /usr/include/c++/13/bits/stl_algobase.h:533:42,
    inlined from '_OI std::__copy_move_a(_II, _II, _OI) [with bool _IsMove = false; _II = const unsigned char*; _OI = unsigned char*]' at /usr/include/c++/13/bits/stl_algobase.h:540:31,
    inlined from '_OI std::copy(_II, _II, _OI) [with _II = const unsigned char*; _OI = unsigned char*]' at /usr/include/c++/13/bits/stl_algobase.h:633:7,
[...]

PR #2274, opened 2023-10-13
build: update flatbuffers dependency to v23.5.26

rkuester:
Update the third_party flatbuffers library to v23.5.26, the
current version in upstream TF. Synchronize the override BUILD
and build_defs.bzl files with those from upstream TF at
e4485c98eae.

Also update the Makefile build, which downloads flatbuffers
separately. Rebase the patch applied to the download.

Regenerate the generated-and-checked-in schemas (see
ci/sync_from_upstream_tf.sh and codegen/preprocessor/
update_schema.sh), because they are stamped with the version of
the flatbuffers library, and fail a static_assert if they are
built with a different version of flatbuffers than they were
generated with.

BUG=unsuccessful attempt to fix warning in #2183


rascani:
We also need to update the version downloaded as part of the make build: https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/tools/make/flatbuffers_download.sh#L57

rascani (approved):

rkuester:
@rascani, the merge is failing for a reason I've not encountered before: https://github.com/tensorflow/tflite-micro/pull/2274/checks?check_run_id=17757924691. Can you please take a look?

rascani:
@Mergifyio refresh

rascani:
@Mergifyio refresh

PR #2284, opened 2023-10-20
`Layer by Layer TFLM vs TfLite python debug tool`

turbotoribio:
- Introduces a tool that simplifies debugging by doing automatic layer by layer comparison between TFLM and TfLite  and allowing for TFLM comparison between x86 reference implementations and  optimized implementations.

BUG=[b/288141725](https://b.corp.google.com/issues/288141725)

### DIFF
--- a/tensorflow/lite/micro/kernels/xtensa/lstm_eval_hifi.cc
+++ b/tensorflow/lite/micro/kernels/xtensa/lstm_eval_hifi.cc
@@ -114,7 +114,7 @@ void calc_cell_state_without_cifg(int16_t* cell_state,
 // residue iterations
 #pragma concurrent
 #pragma loop_count max = 7
-    for (i = 0; i < ((num_elms) & 7); i++) {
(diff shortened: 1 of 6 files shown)


rascani (approved):

tensorflow/lite/micro/kernels/xtensa/lstm_eval_hifi.cc
rascani, line 1: You'll probably need to revert the changes in this file. These show up with newer versions of clang-format than what is installed on the docker image. It'll likely fail CI.

rascani (dismissed):

tensorflow/lite/micro/kernels/xtensa/lstm_eval_hifi.cc
turbotoribio: yeah every time I run the script to format everything it gets added . I thought I excluded it when pushing the format changes , will revert right now !

rascani (approved):

PR #2287, opened 2023-10-23
Adds optimized kernels for Signal ops

suleshahid:
Adds optimized directory structure for Signal OPs.
This PR targets optimizations for xtensa for FftAutoScale and FilterBankSquareRoot.

BUG=[287346710](https://b.corp.google.com/issues/287346710)


signal/micro/kernels/xtensa/fft_auto_scale.cc
```diff
@@ -0,0 +1,101 @@
...
+#include "tensorflow/lite/kernels/internal/tensor_ctypes.h"
+#include "tensorflow/lite/kernels/kernel_util.h"
+#include "tensorflow/lite/micro/kernels/kernel_util.h"
+#include "tensorflow/lite/micro/micro_context.h"
+
+#if XCHAL_HAVE_HIFI3
+#include <xtensa/tie/xt_hifi3.h>
+// Implementation for DSPs that support the Hifi3 ISA. Bit exact with the
+// portable version below.
+int FftAutoScale(const int16_t* input, int size, int16_t* output) {
```
rascani, line 34: We should still put this in an anonymous namespace. I would suggest using the same one as Eval below, but name this XtensaFftAutoScale to ensure no lookup issues.

signal/micro/kernels/xtensa/fft_auto_scale.cc
```diff
@@ -0,0 +1,101 @@
...
+
+#include "signal/micro/kernels/fft_auto_scale.h"
+#include "signal/src/max_abs.h"
+#include "signal/src/msb.h"
+#include "tensorflow/lite/kernels/internal/tensor_ctypes.h"
+#include "tensorflow/lite/kernels/kernel_util.h"
+#include "tensorflow/lite/micro/kernels/kernel_util.h"
+#include "tensorflow/lite/micro/micro_context.h"
+
+#if XCHAL_HAVE_HIFI3
```
rascani, line 30: We've typically used our own defines (HIFI3/4/5). Any reason to prefer the XCHAL versions?

Seems like it would simplify our life to just entirely switch over to the XCHAL defines, but I'm not sure what that might break.

signal/src/max_abs.h
```diff
@@ -22,6 +22,10 @@ limitations under the License.
 namespace tflite {
 namespace tflm_signal {
 
+#if defined(XTENSA)
+int16_t XtensaMaxAbs16(const int16_t* input, int size);
```
rascani, line 26: Can we just put this within MaxAbs16 so callers can just use that? They don't need to know they're calling an optimized version.

I'm thinking something like:

```
int16_t MaxAbs16(const int16_t* input, int size) {
#if defined(XTENSA)
  return XtensaMaxAbs16(input, size);
#else
  int16_t max = 0;
  ...
  return max;
#endif
}
```

WDYT?

signal/micro/kernels/xtensa/fft_auto_scale.cc
suleshahid: Done.

signal/micro/kernels/xtensa/fft_auto_scale.cc
suleshahid: Following the source here. It does seem simpler this way.

signal/src/max_abs.h
suleshahid: Yep that looks better, don't need to add a new function in the .h now too. Still keeping the name and separate function so its a bit cleaner.

rascani (dismissed):

rascani (approved):

PR #2317, opened 2023-11-15
PyPI whl build and upload to PyPI test server nightly job

jwithers:
This script builds the python 3.10 and 3.11 wheels using python/tflite_micro/pypi_build.sh and then uploads to the test PyPI server with python/tflite_micro/pypi_upload.sh. It is scheduled to run at 1300 UTC each day (0600 or 0700 PST). The runs require approximately 11 minutes.

This job can also be run manually by going to Actions > PyPI Build. There is an option to run only the whl build without upload for troubleshooting.

It also should be relatively obvious how to add a release build option if the live server key was added to repo secrets and the script env block and another optional run block added with the -p <live server key> switch added to a copy of the upload script line.

I have tested the build block, but can't test the upload section since I don't have access to the PYPI_API_KEY secret. 

BUG=request from rj

### DIFF
--- a/.github/workflows/pypi_build.yml
+++ b/.github/workflows/pypi_build.yml
@@ -43,7 +43,7 @@ jobs:
           ls -l bazel-pypi-out
       - name: upload to pypi
         if: |
-          github.event.inputs.upload-type == 'test-pypi' ||
+          github.event.inputs.upload-type == 'pypi' ||
           github.event_name == 'schedule'
         run: |
           python/tflite_micro/pypi_upload.sh \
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi_build.yml
```diff
@@ -0,0 +1,50 @@
...
+    - cron: '0 13 * * *'
+
+  workflow_dispatch:
+    inputs:
+      upload-type:
+        description: 'Upload type'
+        required: true
+        default: 'test-pypi'
+        type: choice
+        options:
```
rascani, line 22: Can we add an option for `'pypi'` and have that be the default? It would need to drop the `--test-pypi` flag as well.

I think we'd need to use a different TWINE_PASSWORD as well.

rascani (approved):

PR #2345, opened 2023-12-07
Provide a way for kernels to request an optional scratch buffer size

mansnils:
A more optimized implementation may require more scratch buffer usage. This enables such an operator to have a fallback if tensor arena size is not enough,

### DIFF
--- a/tensorflow/lite/core/c/common.h
+++ b/tensorflow/lite/core/c/common.h
@@ -1,4 +1,4 @@
-/* Copyright 2023 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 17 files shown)


rascani:
Just a couple initial thoughts:

1. From a user's perspective, determining the right size for the arena becomes a little complicated now. How do we communicate to the user that "if you provide X amount of memory, you'll improve latency"? 
2. On the implementation, we should probably attach the actual scratch buffer size to the `GetScratchBuffer` call rather than flagging it in the node to use the fallback. Each node can have multiple scratch buffers, so we'll need to associate those with the scratch handles.

rascani:
@mansnils - I think I like the idea of a build flag, but I think that might beg the question: how useful is the "both" option? The code logic becomes very simple with a build flag: if optimizing for size, scratch buffer is X, else Y. The "both" option brings more complex logic and I'm not sure I see when it would be useful to the user. Maybe it makes sense for multi-tenant situations where multiple models are sharing the same arena space (thus the arena size is the max of the model set)?

mansnils:
@rascani - perhaps the added complexity that comes with the "both" option can not be justified. Considering that the simpler build option proposal kind of solves the original problem, the "both" option can be seen more as nice to have. I am happy to update the draft PR with the new proposal.

mansnils:
To give an example of a multi-tenant use case, it could be running a test suite with multiple models for a given target.
The simpler build flag proposal could still solve it by splitting the test suite in two and building two binaries.

mansnils:
Closing in favor of https://github.com/tensorflow/tflite-micro/pull/2408

Issue #2338, opened 2023-11-30
Shape assertion and unintialized memory in Conv2D


cosw0t:
I have a simple keras model made of a Conv2D with dilation_rate of 2 - converted to tflite and int8-quantized

the first issue is in main and the error is reading the shapes in, the assertion here:
https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/kernels/internal/types.h
called from here:
https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/kernels/internal/reference/integer_ops/conv.h#L52C4

Here's the branch to reproduce: https://github.com/cosw0t/tflite-micro/tree/bug_report/conv2d_main
Command:
```
bazel run tensorflow/lite/micro/examples/bug_report_conv2d:bug_report_conv2d_test
```

It asserts because it has stored input shape: `[4, 1, 49, 3]` and output shape: `[1, 1, 45, 30]`, but the input shape is wrong, it should be [1, 1, 49, 3]. The shape is correctly reported by Netron as `[1, 1, 49, 3]`.

Second issue:
Following the discussion in #2336 I have rebased onto https://github.com/rascani/tflite-micro/tree/initialize_values that fixes #2337, however it doesn't include #2335 that is instead in main.

Msan still complains about uninitialized memory. The model is the same.

Here's the branch to reproduce: https://github.com/cosw0t/tflite-micro/tree/bug_report/conv2d_2337
Command:
```
bazel run --config msan tensorflow/lite/micro/examples/bug_report_conv2d:bug_report_conv2d_test
```

output:

```
==637541==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x7f9d3ceca0da in int gemmlowp::SaturatingRoundingDoublingHighMul<int>(int, int) /proc/self/cwd/external/gemmlowp/fixedpoint/fixedpoint.h:342:26
    #1 0x7f9d3ceca0da in tflite::MultiplyByQuantizedMultiplierSmallerThanOneExp(int, int, int) /proc/self/cwd/./tensorflow/lite/kernels/internal/common.h:359:7
    #2 0x7f9d3ceca0da in tflite::reference_integer_ops::AddFunc(signed char, signed char, tflite::ArithmeticParams const&) /proc/self/cwd/./tensorflow/lite/kernels/internal/reference/integer_ops/add.h:186:7
[...]

Issue #2319, opened 2023-11-16
Dilated Convolutional Layers Produce Wrong Output


apage224:
Issue: Models containing dilated convolutional layers produce incorrect output. This is true for 1D and 2D convolutional layers, all tested dilation_rate sizes, and for both fp32 and int8 quantized models. This issue has been brought up in #1646, #1629, #1231 which had no response nor resolution and closed to due being stale.

Reproduce: Running the following script will show that TF and TFLite produce correct results while TFLM does not.

```python

import numpy as np
import tensorflow as tf
import tflite_micro as tflm

INPUT_SHAPE = (8, 1)
DILATION_RATE = 2  # <- If 1 then assertions pass

# Create example dilated convolutional model
inputs = tf.keras.layers.Input(shape=INPUT_SHAPE)
y = inputs
y = tf.keras.layers.Convolution1D(1, 3, dilation_rate=DILATION_RATE, padding='same')(y)
model = tf.keras.Model(inputs=inputs, outputs=y)


# Convert to TF Lite model
converter = tf.lite.TFLiteConverter.from_keras_model(model=model)
tflite_model = converter.convert()

# Create TFLite and TFLM interpreter
tfl_interpreter = tf.lite.Interpreter(model_content=tflite_model)
tfl_interpreter.allocate_tensors()

tflm_interpreter = tflm.runtime.Interpreter.from_bytes(tflite_model)

# Create input tensor of 1s
x = np.ones((1,)+INPUT_SHAPE, dtype=np.float32)

# Invoke TF, TFLite and TFLM models using same input
y_tf = model(x)

tfl_interpreter.set_tensor(tfl_interpreter.get_input_details()[0]["index"], x)
tfl_interpreter.invoke()
y_tfl = tfl_interpreter.get_tensor(tfl_interpreter.get_output_details()[0]["index"])

tflm_interpreter.set_input(x, 0)
tflm_interpreter.invoke()
y_tflm = tflm_interpreter.get_output(0)

# Below passes
np.testing.assert_almost_equal(y_tf, y_tfl, decimal=3, err_msg="TF and TFLite output do not match.")

# Below fails
np.testing.assert_almost_equal(y_tf, y_tflm, decimal=3, err_msg="TF and TFLM output do not match.")


```

ddavis-2015:
@apage224 

Thank you for your interest in TFLM.  Your reporting of this issue is much appreciated.  This issue is currently under investigation.

apage224:
[...]

Issue #1646, opened 2023-01-05
Dilated convolution produces wrong output


silabs-Vladimir:
**Issue**: TFLite Micro produces incorrect results compared to TFLite when running the same model, containing dilated convolution, on the same input.

**Steps to Reproduce**: Create a model containing a single dilated convolution, then execute it with an input of ones using the TFLite interpreter and the TFLite micro interpreter and compare the results.

**Cause**: The `space_to_batch_nd` and `batch_to_space_nd` ops in TFLite Micro generate incorrect padding values.

**Possible Solution**: Pass the correct quantization offset value in the `space_to_batch_nd.h `file.

___
### Details:

I saw that TFLite, instead of generating ops containing dilation, actually generates the following ops:
`space_to_batch_nd(with padding) -> conv2d(regular, no dilation) -> batch_to_space_nd(with cropping)`

Since I know that `conv2d` is functioning correctly, I investigated the `batch_to_space/space_to_batch` ops and found that they generate different paddings on TFLite Micro and TFLite. It is supposed to be zero-padding, and that is what TFLite does. TFLite Micro, on the other hand, produces paddings with seemingly random values. 

I played around and found that https://github.com/tensorflow/tflite-micro/blob/172aa57aac9121752b701f8206a6f5897999a4b4/tensorflow/lite/kernels/internal/reference/space_to_batch_nd.h#L77-L78
seems to be the culprit, as the offset value it gets seems wrong. At least when testing `space_to_batch_nd` in isolation.

This issue seems to be related to #1629.

Issue #1629, opened 2022-12-19
Conv1D dilation is not working.


KukhoHeo:
Hi, im using tflite-micro latest version.
my specific version is follow:
Python:  3.7.9 (default, Aug 31 2020, 17:10:11) [MSC v.1916 64 bit (AMD64)]
Keras:  2.10.0
Tensorflow:  2.11.0-dev20220812

input_layer = Input(shape=(input_shape ))

#Series of temporal convolutional layers with dilations increasing by powers of 2.
conv_1 = Conv1D(filters=128, kernel_size=2, dilation_rate=2,
                padding='causal', strides=1,input_shape=input_shape,
                kernel_regularizer=regularizers.l2(0.01),
                activation='relu')(input_layer)

#Dropout layer after each 1D-convolutional layer
drop_1 = SpatialDropout1D(0.05)(conv_1)

#Flatten layer to feed into the output layer
flat = Flatten()(drop_1)

output_layer = Dense(2, activation='softmax')(flat)

TCN = Model(inputs=input_layer, outputs=output_layer)
TCN.compile(loss='mean_squared_error',
              optimizer=optimizers.Adam(lr=0.002),
              metrics=['mae', 'accuracy'])
TCN.fit(x_train, y_train,
          batch_size=128,
          epochs=10,
          verbose=1,
          validation_data=(x_test, y_test),
          callbacks = [TensorBoard(log_dir='../../logs/TCN1')])

When the following model is used, the values of the tflite result and the tflite-micro result are different.

but, if the dilation_rate set to 1, the results are same value.

So, i think dilation is not work on tflite-micro.

What should I do to get same result?

PR #2356, opened 2023-12-15
Compute output shapes for some kernels

### LINKED ISSUE #2368: Compute output shapes for some kernels
@tensorflow/micro 

Update the output tensor shape during prepare phase when the computed shape does not match the shape in the flatbuffer.

Kernels:
- BATCH_TO_SPACE_ND
- SPACE_TO_BATCH_ND
- CONV
- RESHAPE
- EXPAND_DIMS
- DEPTHWISE_CONV

Update CMSIS_NN and ARC_MLI optimized kernels.
Add additional tests from TfLite.
Update existing tests.

ddavis-2015:
@tensorflow/micro

Update the output tensor shape during prepare phase when the computed shape does not match the shape in the flatbuffer.

Kernels:
- BATCH_TO_SPACE_ND
- SPACE_TO_BATCH_ND
- CONV
- RESHAPE
- EXPAND_DIMS
- DEPTHWISE_CONV

Update CMSIS_NN and ARC_MLI optimized kernels.
Add additional tests from TfLite for BATCH_TO_SPACE_ND and SPACE_TO_BATCH_ND.
Update existing tests.
Add tests for Keras model using convolution with dilation > 1.

Update memory_arena_threshold_test to increase total, tail, and persistent allocation sizes:
- Add 20 bytes for CONV output shape
- Add 15 bytes for arena allocation alignment
- x2 convolution layers

Update micro_speech_test arena size as per description in C++ code.

See #2319 for additional details.

Resolves [b/317362237](https://issuetracker.google.com/317362237)

bug=fixes #2368 #1646 #1629 #1231 #2338 #2319


tensorflow/lite/micro/kernels/expand_dims_test.cc
```diff
@@ -187,25 +184,41 @@ TF_LITE_MICRO_TEST(ExpandDimsNegativeAxisTest1) {
...
+  // tensor dimensions are computed correctly.
+  int golden_dims[] = {4, 1, 3, 2, 1};
   int output_dims[] = {4, 1, 3, 1, 2};
 
+  tflite::testing::TestExpandDims<float>(input_dims, input_data, axis_dims,
+                                         axis_data, golden_dims, output_dims,
+                                         golden_data, output_data);
+}
+
+TF_LITE_MICRO_TEST(ExpandDimsAxisOutOfRangeShallFailTest1) {
```
rascani, line 212: Nit: rather than having these tests listed as 1 & 2, can we be more specific with the name about what failure condition it is triggering? Seems like 1 is too high of an axis value, while 2 is too low. Maybe ExpandDimsAxisOutOfRangeHighShallFailTest & ExpandDimsAxisOutOfRangeLowShallFailTest.

rascani (commented):
This LGTM, just one tiny nit. We should probably file follow-up issues for supporting this with CEVA & Xtensa optimized kernels.

ddavis-2015:
> This LGTM, just one tiny nit. We should probably file follow-up issues for supporting this with CEVA & Xtensa optimized kernels.

The CEVA and Xtensa code calls the reference Prepare method.  So no additional support work is required.

tensorflow/lite/micro/kernels/expand_dims_test.cc
ddavis-2015: done.

rascani (approved):

PR #2360, opened 2023-12-15
build(bazel): pin bazel version to 7.0.0

rkuester:
Pin the Bazel version used and executed by Bazelisk to guarantee that
TFLM is built with a specified version of Bazel.

When the latest release of Bazel went from version 6.3 to 7.0, our
builds, which were set to follow the latest release (because there was
no .bazelversion), were disrupted due to incompatibilities. By pinning
the Bazel version, we prevent such surprises in the future.

Revise the developer documentation to mandate the use of Bazelisk.

According to [Bazelisk documentation](https://github.com/bazelbuild/bazelisk/blob/9d3fc7d5e356071d2d40b75e69694a0c124e5e33/README.md#ensuring-that-your-developers-use-bazelisk-rather-than-bazel),
direct installations of bazel typically use a wrapper script which
checks .bazelversion also.

BUG=see description

### DIFF
--- a/.bazelversion
+++ b/.bazelversion
@@ -1,0 +1,4 @@
+# The version of bazel for Bazelisk to download and use. See
+# https://github.com/bazelbuild/bazelisk.
+
+7.0.0
--- a/.bazelversion
+++ b/.bazelversion
@@ -1,4 +1,1 @@
-# The version of bazel for Bazelisk to download and use. See
-# https://github.com/bazelbuild/bazelisk.
-
 7.0.0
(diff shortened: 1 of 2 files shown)


tools/bazel.bzl
```diff
@@ -0,0 +1,35 @@
...
+      required: the exact version string required
+
+    Returns:
+      None: when the current version matches the required version
+      Fails with a message otherwise
+    """
+
+    current = versions.get()
+    if current != required:
+        fail(
```
rascani, line 30: Is there a way we could make this a warning? I think it should generally be okay if users use a different version, but we can only guarantee 7.0 will work. WDYT?

rascani (approved):

rascani (approved):

PR #2451, opened 2024-02-10
Generic Benchmark Tool changes

### LINKED ISSUE #2450: Generic Benchmark Tool changes
@tensorflow/micro 

Add built-in model support, chosen at compile time
Add README for build instructions.

ddavis-2015:
@tensorflow/micro

Add ability to build with a model chosen at compile time. Add README with build instructions.

bug=fixes #2450


tensorflow/lite/micro/tools/benchmarking/generic_model_benchmark.cc
```diff
@@ -45,15 +58,17 @@ namespace {
...
 // Seed used for the random input. Input data shouldn't affect invocation timing
 // so randomness isn't really needed.
 constexpr uint32_t kRandomSeed = 0xFB;
 
+#if !defined(MODEL_HEADER_PATH)
 constexpr size_t kTensorArenaSize = 3e6;
-constexpr int kNumResourceVariable = 100;
 constexpr size_t kModelSize = 2e6;
+#else
+constexpr size_t kTensorArenaSize = TENSOR_ARENA_SIZE;
```
rascani, line 69: Why did you need to make the tensor arena adjustable? I'm guessing the answer is "so it could fit" in one of our targets. I'd generally like to make this as absolutely big as possible, as the tool's output will then tell you how much it really needed.

Since we know we can fit 3e6 + 2e6 in the command line version, but don't need the 2e6 model space in the builtin version, maybe we could have the `kTensorArenaSize = 5e6 - g_model_size`?

tensorflow/lite/micro/tools/benchmarking/generic_model_benchmark.cc
```diff
@@ -45,15 +58,17 @@ namespace {
 
 using Profiler = ::tflite::MicroProfiler;
 
-constexpr int kTfLiteAbort = -9;
-
 // Seed used for the random input. Input data shouldn't affect invocation timing
 // so randomness isn't really needed.
 constexpr uint32_t kRandomSeed = 0xFB;
 
+#if !defined(MODEL_HEADER_PATH)
```
rascani, line 65: Can we add a define like `USING_BUILTIN_MODEL` for these conditionals so that they're slightly more understandable at the usage site?

rascani (commented):
This looks great to me, just a minor nit and a question/suggestion.

tensorflow/lite/micro/tools/benchmarking/generic_model_benchmark.cc
ddavis-2015: done

tensorflow/lite/micro/tools/benchmarking/generic_model_benchmark.cc
ddavis-2015: done

rascani (approved):

PR #2468, opened 2024-02-23
Fix array out-of-bounds access in WideDynamicFuncLut

kismeter:
### Problem description:

In the original code, pointer arithmetic of gain_lut and the assignment of gain_lut[4 * interval + 3] could potentially lead to out-of-bounds array access.
On certain architectures (e.g., macOS ARM64), this out-of-bounds access causes the program to crash.
BUG=None, reported issue#2464

### Solution:

Increase the size of the gain_lut_storage array by 1 to provide an extra buffer and prevent overflow during the calculation within the loop.

### Risks and considerations:

Increasing the array size will slightly increase memory usage.
In extremely resource-constrained systems, alternative algorithm implementations may need to be considered.


rascani (approved):
Thank you so much for the bug fix!

rascani:
@Mergifyio refresh

kismeter:
@Mergifyio refresh

PR #2476, opened 2024-02-26
Fix a linker error: undefined kMaxSmallSize

shlmregev:
When building with a recent version of xt-clang and -std=c++11, the linker errs on missing definition of a static constexpr class member. In C++11, static class members still had to be defined in a C++11 file and TFLM code is expected to be compatible with C++11.

BUG=323856831

### DIFF
--- a/tensorflow/lite/kernels/internal/common.cc
+++ b/tensorflow/lite/kernels/internal/common.cc
@@ -17,7 +17,6 @@ limitations under the License.
 
 namespace tflite {
 
-
 // Single-rounding MultiplyByQuantizedMultiplier
 #if TFLITE_SINGLE_ROUNDING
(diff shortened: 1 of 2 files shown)


ddavis-2015:
@rascani @shlmregev 

This PR needs to be submitted in the Tensorflow repository.  The reason is that the file being changed is currently being synched daily from TensorFlow Lite.

FYI, the list of TFLite synched files is [here](https://github.com/tensorflow/tflite-micro/blob/main/ci/tflite_files.txt).

rascani:
Per our OOB conversation, this definition can be defined in runtime_shape.cc, as it is from the runtime_shape.h header. The runtime_shape.h header is not synced from TFLite.

rascani (approved):

PR #2496, opened 2024-03-08
Add meta-data to generic benchmark

### LINKED ISSUE #2495: Add meta-data to generic benchmark
@tensorflow/micro 

Add meta-data to the generic benchmark through built-in strings, which are output each time the binary is run.

[b/310657721](https://issuetracker.google.com/310657721)
[b/310653635](https://issuetracker.google.com/310653635)

ddavis-2015:
@tensorflow/micro

Add meta-data strings to the binary.  Meta-data is output each time binary is executed.  See the README for a sample of the output.

If a Git repo. is not available at ```${TENSORFLOW_ROOT}```, an appropriate message is generated in the meta-data.

Changes to how the CMSISNN and XTENSA NN library download scripts work.

Adds run of generic benchmark with embedded model for default build of x86, Corstone-300, Xtensa to test scripts executed during CI.

bug=fixes #2495


tensorflow/lite/micro/tools/benchmarking/collect_meta_data.sh
```diff
@@ -0,0 +1,159 @@
...
+TARGET_ARCH="${TARGET_ARCH:-x86}"
+OPTIMIZED_KERNEL="${OPTIMIZED_KERNEL:-none}"
+BUILD_TYPE="${BUILD_TYPE:-default}"
+result=$(printf 'TARGET=%s\nTARGET_ARCH=%s\nOPTIMIZATION=%s\nBUILD_TYPE=%s\n' \
+  "${TARGET}" \
+  "${TARGET_ARCH}" \
+  "${OPTIMIZED_KERNEL}" \
+  "${BUILD_TYPE}" \
+)
+if [[ "${XTENSA_CORE}" ]]; then
```
rascani, line 140: We should add more of the XTENSA metadata defined [here](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/tools/make/targets/xtensa_makefile.inc#L17).

Specifically, XTENSA_BASE and XTENSA_TOOLS_VERSION would be super useful. It'd probably also be good to figure add in the nnlib url/md5, but that's definitely more complicated than the CMSIS one.

rascani (approved):

tensorflow/lite/micro/tools/benchmarking/collect_meta_data.sh
ddavis-2015: done

rascani (approved):

PR #2514, opened 2024-03-21
Add GNU Make rule to print GENDIR

shlmregev:
Useful when the TFLM makefile is called recursively and build artifacts used.

BUG=b/327609938


rascani (approved):

rascani:
@Mergifyio refresh

PR #2539, opened 2024-04-11
Add big-endian support for specific target

liamsun2019:
Big-endian support is added for specific target. The codes are conditionally compiled by corresponding macros.


liamsun2019:
CLA failed.

PR #2544, opened 2024-04-16
Quantize, dequantize optimizations for HiFi targets

cad-audio:
BUG=none

### DIFF
--- a/tensorflow/lite/micro/kernels/xtensa/quantize.cc
+++ b/tensorflow/lite/micro/kernels/xtensa/quantize.cc
@@ -75,7 +75,6 @@ TfLiteStatus EvalXtensa(TfLiteContext* context, TfLiteNode* node) {
 
         case kTfLiteInt8: {
           int size = ElementCount(*input->dims);
-#if defined(HIFI5) || defined(HIFI4)
           int32_t zero_point = op_data->quantization_params.zero_point;
           const int8_t* input_data_ptr;
           int8_t* output_data_ptr;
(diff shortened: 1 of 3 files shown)


tensorflow/lite/micro/kernels/xtensa/quantize.cc
```diff
@@ -135,6 +151,17 @@ TfLiteStatus EvalXtensa(TfLiteContext* context, TfLiteNode* node) {
...
+          TF_LITE_ENSURE_EQ(
+              context,
+              xa_nn_elm_requantize_asym16s_asym8s(
+                  tflite::micro::GetTensorData<int8_t>(output),
+                  tflite::micro::GetTensorData<int16_t>(input),
+                  op_data->input_zero_point, op_data->quantization_params.zero_point,
+                  op_data->requantize_output_shift,
+                  op_data->requantize_output_multiplier, size),
+              0);
+#else // defined(HIFI5) || defined(HIFI4) || defined(HIFI4_INTERNAL)
```
rascani, line 164: This #else codeblock is the same as the #if. It seems like this code could have just been left as is?

tensorflow/lite/micro/kernels/xtensa/quantize.cc
```diff
@@ -135,6 +151,17 @@ TfLiteStatus EvalXtensa(TfLiteContext* context, TfLiteNode* node) {
       switch (output->type) {
         case kTfLiteInt8: {
           int size = ElementCount(*input->dims);
+#if defined(HIFI5) || defined(HIFI4) || defined(HIFI4_INTERNAL)
```
rascani, line 154: Note to self: Remove the HIFI4_INTERNAL conditional, that is deprecated.

tensorflow/lite/micro/kernels/xtensa/quantize.cc
rascani: else block should fall back to reference op

rascani: Actually, looks like none of these #ifdefs are needed. In the Eval function, we already do a check for HIFI3, 4, & 5.  If those aren't defined, we'll use a reference implementation of Eval. I think the only #ifdef check still needed is the HIFI_VFPU for the float quantizations. I'll add a commit to this effect. Please let me know if I get anything wrong.

rascani (approved):

rascani:
@Suyash20012000 - Can you sign the CLA? https://github.com/tensorflow/tflite-micro/pull/2544/checks?check_run_id=23948620573

Suyash20012000:
@rascani I have signed the CLA. Requesting you to check.

Issue #1973, opened 2023-05-16
import error tflite_micro


aryanxk02:
I am trying to run `generate_micro_mutable_op_resolver_from_model.py` file via bazel. But seems like I am getting an import error.

```
ModuleNotFoundError: No module named 'tflite_micro'
```

Reference: https://github.com/aryanxk02/tflite-micro/blob/main/tensorflow/lite/micro/tools/gen_micro_mutable_op_resolver/generate_micro_mutable_op_resolver_from_model.py#L26

aryanxk02:
Okay some tweaks here. I changed the import statement from
```
from tflite_micro.tensorflow.lite.tools import visualize as visualize
```
to
```
from tensorflow.lite.tools import visualize as visualize
```

It works now.

vikramdattu:
cc @advaitjain I think this is a bug and import should have been `tensorflow.lite.tools` and not `tflite_micro.`

PR #2545, opened 2024-04-17
gen_micro_mutable script: update import path

mansnils:
BUG=https://github.com/tensorflow/tflite-micro/issues/1973

### DIFF
--- a/tensorflow/lite/micro/tools/gen_micro_mutable_op_resolver/generate_micro_mutable_op_resolver_from_model.py
+++ b/tensorflow/lite/micro/tools/gen_micro_mutable_op_resolver/generate_micro_mutable_op_resolver_from_model.py
@@ -1,4 +1,4 @@
-# Copyright 2023 The TensorFlow Authors. All Rights Reserved.
+# Copyright 2024 The TensorFlow Authors. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -23,7 +23,7 @@ from absl import app
 from absl import flags
 from mako import template
 
-from tflite_micro.tensorflow.lite.tools import visualize as visualize
+from tensorflow.lite.tools import visualize as visualize
 
 TEMPLATE_DIR = os.path.join(os.path.dirname(__file__), 'templates')
 TEMPLATE_DIR = os.path.abspath(TEMPLATE_DIR)


rascani (approved):

rascani:
FYI @mansnils - I needed to revert this due to internal breakage. From my understanding, I believe either `from tflite_micro.tensorflow.lite` or `from tensorflow.lite` *should* work fine. The former should pick up `visualize` from the [TFLM copy](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/tools/visualize.py), whereas the latter should pick up `visualize` from the pip installed version of `tensorflow`. I think its better to use the former.

Were you seeing an issue with that? The linked issue didn't provide repro steps.

mansnils:
> FYI @mansnils - I needed to revert this due to internal breakage. From my understanding, I believe either `from tflite_micro.tensorflow.lite` or `from tensorflow.lite` _should_ work fine. The former should pick up `visualize` from the [TFLM copy](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/tools/visualize.py), whereas the latter should pick up `visualize` from the pip installed version of `tensorflow`. I think its better to use the former.
> 
> Were you seeing an issue with that? The linked issue didn't provide repro steps.

Thanks @RJ - the issue the PR tried to fix has been seen randomly in seemingly similar environments to where it has not been seen but the fix always seemed to work in both environments. That is why I didn't provide the PR immediately. So not sure how to reproduce it.

mansnils:
> > FYI @mansnils - I needed to revert this due to internal breakage. From my understanding, I believe either `from tflite_micro.tensorflow.lite` or `from tensorflow.lite` _should_ work fine. The former should pick up `visualize` from the [TFLM copy](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/tools/visualize.py), whereas the latter should pick up `visualize` from the pip installed version of `tensorflow`. I think its better to use the former.
> > Were you seeing an issue with that? The linked issue didn't provide repro steps.
> 
> Thanks @RJ - the issue the PR tried to fix has been seen randomly in seemingly similar environments to where it has not been seen but the fix always seemed to work in both environments. That is why I didn't provide the PR immediately. So not sure how to reproduce it.

I think the problem might be that from` tflite_micro.tensorflow.lite.tools import visualize as visualize` as it is now works when the tflite-micro wheel is not installed. If it is installed it seems bazel tries to import `visualize` from the actual pip installed tflite-micro package and in fact there is no visualize in that package. For example it seems to work if creating a virtual environment before running the bazel command. I don't know if there is a way to tell bazel to just look in the directory structure.

PR #2557, opened 2024-04-29
CMSIS-NN: Add int32 bias support int16xint8 conv

mansnils:
BUG=CMSIS-NN glue is updated to support int32bias for int16x8


mansnils:
@Mergifyio refresh

PR #2582, opened 2024-05-18
Fix test_cortex_m_corstone_300 ci job

ArmRyan:
* Update path for new corstone300 downloads
 * Add missing benchmark arena size to generic benchmark test
 
BUG=Corstone300 Ci Job fails to run
Author: Ryan O'Shea <ryan.oshea3@arm.com>

### DIFF
--- a/tensorflow/lite/micro/tools/ci_build/test_cortex_m_corstone_300.sh
+++ b/tensorflow/lite/micro/tools/ci_build/test_cortex_m_corstone_300.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright 2022 The TensorFlow Authors. All Rights Reserved.
+# Copyright 2024 The TensorFlow Authors. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -51,4 +51,5 @@ readable_run make -j$(nproc) -f tensorflow/lite/micro/tools/make/Makefile \
   TARGET_ARCH=${TARGET_ARCH} \
   TOOLCHAIN=${TOOLCHAIN} \
   GENERIC_BENCHMARK_MODEL_PATH=tensorflow/lite/micro/models/person_detect.tflite \
+  GENERIC_BENCHMARK_ARENA_SIZE=`expr 150 \* 1024` \
   run_tflm_benchmark
--- a/tensorflow/lite/micro/tools/make/targets/cortex_m_corstone_300_makefile.inc
+++ b/tensorflow/lite/micro/tools/make/targets/cortex_m_corstone_300_makefile.inc
@@ -16,7 +16,13 @@
 # ARM Cortex M makefile targeted for a FVP based on Arm Corstone-300 software.
 # For more info see: tensorflow/lite/micro/cortex_m_corstone_300/README.md
 
-export PATH := $(DOWNLOADS_DIR)/corstone300/models/Linux64_GCC-6.4:$(PATH)
+UNAME_M := $(shell uname -m)
+ifeq ($(UNAME_M), aarch64)
+  export PATH := $(DOWNLOADS_DIR)/corstone300/models/Linux64_armv8l_GCC-9.3:$(PATH)
+else
+  export PATH := $(DOWNLOADS_DIR)/corstone300/models/Linux64_GCC-9.3:$(PATH)
+endif
+
 DOWNLOAD_RESULT := $(shell $(MAKEFILE_DIR)/corstone_300_download.sh $(DOWNLOADS_DIR) $(TENSORFLOW_ROOT))
 ifneq ($(DOWNLOAD_RESULT), SUCCESS)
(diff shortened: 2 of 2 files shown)


ArmRyan:
Hmm, I just noticed there was a patch put up to remove ethos u as coprocessor, but this patch fixes that issue if it was intended to use ethos u?

tensorflow/lite/micro/tools/ci_build/test_cortex_m_corstone_300.sh
```diff
@@ -51,4 +51,5 @@ readable_run make -j$(nproc) -f tensorflow/lite/micro/tools/make/Makefile \
   TARGET_ARCH=${TARGET_ARCH} \
   TOOLCHAIN=${TOOLCHAIN} \
   GENERIC_BENCHMARK_MODEL_PATH=tensorflow/lite/micro/models/person_detect.tflite \
+  GENERIC_BENCHMARK_ARENA_SIZE=`expr 150 \* 1024` \
```
ddavis-2015, line 54: Good catch with `GENERIC_BENCHMARK_ARENA_SIZE`.  I failed to test locally without it, and it turns out the simulator just hangs (arena size is larger than mapped address space?).

tensorflow/lite/micro/tools/ci_build/test_cortex_m_corstone_300.sh
```diff
@@ -51,4 +51,5 @@ readable_run make -j$(nproc) -f tensorflow/lite/micro/tools/make/Makefile \
   TARGET_ARCH=${TARGET_ARCH} \
```
ddavis-2015, line 51: The `CO_PROCESSOR=ethos_u` line needs to be removed.  With the co-processor specified, the simulator does not make the instruction cycle counter available.  I have tested this locally.

tensorflow/lite/micro/tools/make/targets/cortex_m_corstone_300_makefile.inc
ddavis-2015, line 1: Improves on my changes (PR #2579).  👍

ddavis-2015 (changes requested):

tensorflow/lite/micro/tools/ci_build/test_cortex_m_corstone_300.sh
ArmRyan: yes it allocated 3e6 which is way bigger than memory in the platform linker script

tensorflow/lite/micro/tools/ci_build/test_cortex_m_corstone_300.sh
ArmRyan: cpu cycle count on fvp is not accurate, and is not good for benchmarking! Should be able to get the count with ethos u as that is what the fvp was made to be cycle accurate to

ddavis-2015: Understood, however I just get zeros from the `GetCurrentTimeTicks` function with the co-processor defined.

mansnils: Please refer to this - [run-on-fvp-based-on-arm-corstone-300-software]( https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/benchmarks/README.md#run-on-fvp-based-on-arm-corstone-300-software ) - for why  only `CO_PROCESSOR=ethos_u` makes sense when running on Corstone-300 FVP and you care about cycle counter.
However supplying` CO_PROCESSOR=ethos_u` is not enough. The supplied tflite model must first be optimized so it contains the Ethos-U custom op. Within the examples that comes with tflite-micro there are only one such example and that is the person detect model that can be run with the network tester example or the benchmarks.
So the rationale for test_cortex_m_corstone_300.sh is that here we don't care about cycle measurements. We could then remove `CO_PROCESSOR=ethos_u.` However the rationale for keeping is to compile test that code path.

rascani: If I understand correctly:

* If we have `CO_PROCESSOR=ethos_u` defined, we'll read the NPU's cycle counts for operations.
* The model we're using for test here doesn't have the ethos_u custom op in the model, so regardless of `CO_PROCESSOR`, we'll be executing on CPU.

The way I see it, we have three options (feel free to suggest alternatives):
1. Use the modified person detect model with Ethos-U custom op and enable co-processor. We should get proper benchmark numbers for person detect on Ethos-U.
2. Keep test as is, where we're running person detect on CPU and reading the inaccurate CPU cycle counts. Add a disclaimer that these are not accurate results and should not be relied upon.
3. Remove the test to not be misleading. Downside here is we won't have continuous testing of tflm_benchmark utility on ARM, but maybe that's okay if we can't trust it.

I would lean towards 1 or 2. @mansnils @ArmRyan WDYT?

rascani (approved):
Going to move ahead with landing this, we can sort out the benchmark test in a separate PR.

ArmRyan:
I would prefer option 1 as its very easy to skip the disclaimer and look straight at the numbers

PR #2603, opened 2024-05-29
Update new_platform_support.md

StrugglingBunny:
Change link of system_setup.h

BUG=cleanup


rascani (approved):
Thank you for the fix!

Issue #2593, opened 2024-05-22
Passing custom/additional data to kernels 


DanieleParravicini-Synthara:
Hello, I have the following use case that I would like to cover.

How can I provide an efficient kernel for a layer without breaking the compatibility of the model?

Consider:

1. a layer of a NN model (e.g. a linear layer) that could be accelerated by a dedicated hardware module. 
E.g. imagine you have M cores and you break the linear layers in M chunks.

2.  the custom hardware module could benefit from additional information (e.g. how many elements each core could use).

Here follows some ideas:

1. use custom operator

    1. for each layer that you want to accelerate create a custom operator
    2. modify the NN model replacing the original layer with a custom operator that is functionally equivalent to the first one
    3. retrain the NN model
    4. convert the model to a TensorFlow Lite Model adding the additional custom operator (and possibly passing additional info to the new custom layer using `custom_options` field of Operator of flatbuffer)
    5. provide a implementation of the operator to the interpreter that reads the `custom_options` field and execute the layer accordingly

4. keep the same operator and use a custom conversion tools that adds information on the `custom_options` field. In this way I can avoid the first three steps of the previous list and do the following
    1.    convert the model to a TensorFlow Lite Model adding the additional custom operator and passing additional info to the new custom layer using `custom_options` field of Operator of flatbuffer.
       e.g. like this
```
import np
import tensorflow as tf
from tensorflow.lite.python import schema_py_generated as schema_fb

def load_model(save_path: str):
    with open(save_path, "rb") as f:
       return f.read()

tflite_quantized = load_model("models/mlp_int8.tflite")

aModel = schema_fb.ModelT.InitFromPackedBuf(tflite_quantized, 0)
     

def BuiltinCodeToName(code):
  """Converts a builtin op code enum to a readable name."""
  for name, value in schema_fb.BuiltinOperator.__dict__.items():
    if value == code:
      return name
  return None
     

for i, op in enumerate(aModel.subgraphs[0].operators):
    op_code = aModel.operatorCodes[op.opcodeIndex].builtinCode
    print(f"[{i}] :  {BuiltinCodeToName(op_code)} ({op_code})")

     
### FROM HERE
custo = np.ones(10,dtype=np.uint8)
aModel.subgraphs[0].operators[4].customOptions = custo
### TO HERE

from tflite_support import flatbuffers
b = flatbuffers.Builder(0)

b.Finish(
    aModel.Pack(b)
)
model_buff = b.Output()

[...]

PR #2605, opened 2024-06-04
Adds GetCurrentOperatorIndex to micro_interpreter

DanieleParravicini-Synthara:
In this way an AOT can add metadata that could then be accessed with the couple GetCurrentSubgraphIndex, GetCurrentOperatorIndex.
These can be accessed from tflite::GetMicroContext(context)->graph()

Refers to #2593
BUG=#2593


DanieleParravicini-Synthara:
Hello @rascani, 
Hope that everything is right.
What do you think, is it a good starting point?

On a side note, I see that the CLA is in place, but the CI fails. However, I am not so versed in github actions, but I guess it is missing the label to go further right?

rascani (dismissed):

rascani:
@DanieleParravicini-Synthara - Looks like the only failing check is the code style check. I can fix that up if I have permission to modify your branch, or you can do it by running:
```
tensorflow/lite/micro/tools/ci_build/test_code_style.sh --fix_formatting
```

rascani (approved):

PR #2626, opened 2024-07-14
Update Ethos-U driver.

### LINKED ISSUE #2619: Feature request: Update Ethos-U driver ?
Hi! It seems that the Ethos-U core driver currently used by tflite-micro is somewhat outdated. It's missing the `ethosu_address_remap` callback, which was introduced around release v23.02 (the current release is 24.05). This function is crucial; without it, invocation will fail if the NPU core cannot use addresses directly. 

Additionally, could you please consider allowing configuration arguments (such as the ones defined in `src/ethosu_config_uxx.h`) to be passed to the driver from the command line when building the library?

Thanks!

iabdalkader:
* Update Ethos-U driver to the latest (v24.05)
* Allow passing extra C flags to the driver from the command line.
* Fixes #2619

Note I tested with `TARGET=cortex_m_corstone_300 test_network_tester_test` and on real hardware as well built with:
```bash
make -j12 -f tensorflow/lite/micro/tools/make/Makefile \
TARGET=cortex_m_generic TARGET_ARCH=cortex-m55 \
CO_PROCESSOR=ethos_u ETHOSU_ARCH=u55 OPTIMIZED_KERNEL_DIR=ethos_u  \
CORE_OPTIMIZATION_LEVEL=-O2 KERNEL_OPTIMIZATION_LEVEL=-O2 \
THIRD_PARTY_KERNEL_OPTIMIZATION_LEVEL=-O2 \
TARGET_TOOLCHAIN_ROOT=/opt/arm-none-eabi/bin/ \ 
TARGET_TOOLCHAIN_PREFIX=arm-none-eabi- BUILD_TYPE=release microlite
```

BUG=#2619

### DIFF
--- a/tensorflow/lite/micro/tools/make/third_party_downloads.inc
+++ b/tensorflow/lite/micro/tools/make/third_party_downloads.inc
@@ -40,5 +40,3 @@ EMBARC_MLI_PRE_COMPILED_URL := "https://github.com/foss-for-synopsys-dwc-arc-pro
 EMBARC_MLI_PRE_COMPILED_MD5 := "173990c2dde4efef6a2c95b92d1f0244"
 
 # Skip md5sum-check since ethos-u-core-driver download link is non-deterministic, see  https://github.com/google/gitiles/issues/84
-ETHOSU_URL := "https://review.mlplatform.org/plugins/gitiles/ml/ethos-u/ethos-u-core-driver/+archive/24455eedb9e8939f8a28ca0101a6f2d171e1b2f9.tar.gz"
-ETHOSU_MD5 := "SKIP_MD5_CHECK"
--- a/tensorflow/lite/micro/tools/make/ethos_u_core_driver_download.sh
+++ b/tensorflow/lite/micro/tools/make/ethos_u_core_driver_download.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+# Copyright 2024 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+#
+# Called with following arguments:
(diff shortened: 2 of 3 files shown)


iabdalkader:
Rebased.. unsure how to fix that CI entry test error.

tensorflow/lite/micro/tools/make/third_party_downloads.inc
```diff
@@ -40,5 +40,3 @@ EMBARC_MLI_PRE_COMPILED_URL := "https://github.com/foss-for-synopsys-dwc-arc-pro
 EMBARC_MLI_PRE_COMPILED_MD5 := "173990c2dde4efef6a2c95b92d1f0244"
 
 # Skip md5sum-check since ethos-u-core-driver download link is non-deterministic, see  https://github.com/google/gitiles/issues/84
-ETHOSU_URL := "https://review.mlplatform.org/plugins/gitiles/ml/ethos-u/ethos-u-core-driver/+archive/24455eedb9e8939f8a28ca0101a6f2d171e1b2f9.tar.gz"
```
rascani, line 43: Did you have a particular reason for migrating from using a download link to using git clone? The URL approach would continue to work by simply updating the link to: 
```
https://review.mlplatform.org/plugins/gitiles/ml/ethos-u/ethos-u-core-driver/+archive/9622608a5cc318c0933bcce720b59737d03bfb6f.tar.gz
```

Regardless of HTTP or git download, it is important that we version based on the SHA, not the tag. Tags can be moved, which can result in our scripts breaking (this has happened before). While SHAs can technically be erased too, that is significantly less likely.

iabdalkader: > Did you have a particular reason for migrating from using a download link to using git clone?

I was just following the existing pattern for the other Ethos platform package, which is also a git repo. It's also easier to tell which specific version of the driver is being used from a tag or SHA than from a download URL or even the downloaded package. Finally, probably only important for development, but if you want to check out a different commit for testing something, you can't do that with a downloaded package.

I've switched to commit SHA and retested the build, and retested on hardware.

rascani (approved):

iabdalkader:
@rascani Seems my patch might have broken some test here: https://github.com/tensorflow/tflite-micro/actions/runs/10087978372/job/27892985767

I'll try to fix it.

rascani:
Thanks @iabdalkader

mansnils:
Thanks for this @iabdalkader

iabdalkader:
You're welcome. Note that there were some issue that I've missed because the workflow runs on schedule. The workflow for Cortex-M doesn't take much time, just a little over 10 minutes, so perhaps it's best to change it to trigger on PRs. This could save a lot of time in the future.

mansnils:
Yes agreed. We are looking into that.

Issue #2636, opened 2024-07-24
feature: model compression


rkuester:
An issue to track the implementation of model compression.

**Merge Queue**
1. [PRs from rkuester/feat-compression](https://github.com/tensorflow/tflite-micro/pulls?q=+is%3Apr+author%3Arkuester+head%3Afeat-compression) are open for approval and merging. Because of our one-commit-per-PR policy, there is typically only one PR open at a time.
2. [Commits along the branch rkuester/feat-compression-next](https://github.com/tensorflow/tflite-micro/compare/main...rkuester:tflite-micro:feat-compression-next) are queued, so to speak, for submission as PRs. Looking at this branch might give context for the open PR in *feat-compression*, see above. Be aware—this branch is rebased often.
3. The branch [rkuester/compress-testing]() typically contains the final result once all PRs for the model compression feature are merged. This is the Oort cloud from which commits along *feat-compression-next*, which turn into PRs from *feat-compression*, are formed. [Explore](https://github.com/rkuester/tflite-micro/blob/compress-testing/README.md) this branch or [compare it to main](https://github.com/tensorflow/tflite-micro/compare/main...rkuester:tflite-micro:compress-testing) to see the full model compression feature implementation and understand the queued PRs in context.

rkuester:
This task remains open; PRs and issues link to it for tracking purposes regularly.

tinskip:
Hi. Was this change intended for whether TFLM compression is or or off, or should it have been made conditional as well?

https://github.com/tensorflow/tflite-micro/blame/main/tensorflow/lite/micro/kernels/concatenation.cc#L189

ddavis-2015:
> Hi. Was this change intended for whether TFLM compression is or or off, or should it have been made conditional as well?
> 
> https://github.com/tensorflow/tflite-micro/blame/main/tensorflow/lite/micro/kernels/concatenation.cc#L189

@tinskip This change is intentional and is there regardless of RTLM conditional compression compilation.  The change updates the LRTM (TFLM) code to match the LiteRT (TfLite) reference implementation.

PR #2642, opened 2024-07-25
feat: add fixed-capacity, statically-allocated type tflite::StaticVector

rkuester:
Add a type, tflite::StaticVector, which behaves like std::vector, but
which avoids heap memory allocation.

BUG=#2636

### DIFF
--- a/tensorflow/lite/micro/static_vector.h
+++ b/tensorflow/lite/micro/static_vector.h
@@ -1,0 +1,81 @@
+// Copyright 2024 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef TENSORFLOW_LITE_MICRO_STATIC_VECTOR_H_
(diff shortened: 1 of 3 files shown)


rkuester:
To my knowledge, we don't have any type like this yet—something that behaves like a std::vector, but can be statically allocated. Am I wrong?

This is used in upcoming PRs.

tensorflow/lite/micro/static_vector.h
```diff
@@ -0,0 +1,90 @@
+// Copyright 2023 The TensorFlow Authors. All Rights Reserved.
```
rascani, line 1: 2024

tensorflow/lite/micro/static_vector.h
```diff
@@ -0,0 +1,90 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef TENSORFLOW_LITE_MICRO_STATIC_VECTOR_H_
+#define TENSORFLOW_LITE_MICRO_STATIC_VECTOR_H_
+
+#include <array>
+#include <cassert>
+#include <cstddef>
+#include <ostream>
```
rascani, line 21: Can we avoid ostream usage? I have found just the inclusion of the i/o stream headers results in binary bloat.

rascani (commented):
Generally LGTM, just a couple minor comments.

tensorflow/lite/micro/static_vector.h
rkuester: Sure, that's only there for the operator<<() supplied as a convenience at the bottom. I'll axe it, and either put that in a separate header or just inline something in the one place that used operator<<().

(That lightweight operator<<() is a little perilous anyhow; it doesn't handle all the possible side-effects that using other operator<<()s within its implementation can have on the ostream state.)

rascani (approved):

rkuester:
These changes required #2648.

rascani (approved):

PR #2643, opened 2024-07-25
Fix Ethos-U driver build.

iabdalkader:
The updated Ethos-U driver v24.05 (introduced in #2626), contains a second set of device functions for the U85. This can cause the linker to link the functions of the U85, instead of the U55-U65, if they're both included in the driver's sources files. Therefore a glob pattern can no longer be used to collect the driver's source files in the Makefile.

BUG=#2639


rascani (approved):

iabdalkader:
@rascani I'm just curious, how long does the Cortex-M workflow take to finish ? I triggered it manually in my fork, on this patch, and it's been running for an hour.

iabdalkader:
Just a heads up there might still be an issue with the FPV, it seems to get stuck. I'm not sure I can fix this one, but I did try a newer platform (23.x and 24.05), with your patches, but that didn't fix it. If there's a way to debug it or get some output out of it, let me know I could give it another shot. I did try telnet but there's no output.

iabdalkader:
@rascani I found the issue and will send another fix.

Issue #2646, opened 2024-07-29
TFLM compression changes


ddavis-2015:
@tensorflow/micro 

Initial header file changes/additions

PR #2647, opened 2024-07-29
TFLM compression changes

ddavis-2015:
@tensorflow/micro

initial header file changes/additions.

bug=#2646

### DIFF
--- a/tensorflow/lite/micro/compression.h
+++ b/tensorflow/lite/micro/compression.h
@@ -33,7 +33,7 @@ enum class CompressionScheme : uint8_t {
 };
 
 // TODO(ddavis-2015): pack struct
-struct BinQuantData {
+struct LookupTableData {
   static constexpr size_t kMaxBitWidth = 7;
   static constexpr size_t kMaxValueTableChannelStride = 128;
 
@@ -47,7 +47,7 @@ struct BinQuantData {
 };
 
 union CompressionData {
-  BinQuantData bin_quant;
+  LookupTableData* lut_table;
 };
 
 // TODO(ddavis-2015): pack struct
--- a/tensorflow/lite/micro/tools/layer_by_layer_debugger.py
+++ b/tensorflow/lite/micro/tools/layer_by_layer_debugger.py
@@ -101,7 +101,7 @@ def numpy_from_tensor_type(tensor_type_idx):
       layer_schema_fb.TensorTypes.INT64:
       np.int64,
       layer_schema_fb.TensorTypes.STRING:
-      np.string_,
+      np.bytes_,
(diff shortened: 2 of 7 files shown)


tensorflow/lite/micro/compression.h
```diff
@@ -0,0 +1,70 @@
...
+
+static constexpr const char* kCompressionMetadataString = "TFLM_COMPRESSION";
+
+enum class CompressionScheme : uint8_t {
+  kBinQuant,
+};
+
+// TODO(ddavis-2015): pack struct
+struct BinQuantData {
+  static constexpr size_t kMaxBitWidth = 7;
```
rascani, line 37: Is there any reason we actually have to constrain this to 7 bits? Theoretically, we could do 8 bit LUTs for f32s?

tensorflow/lite/micro/compression.h
```diff
@@ -0,0 +1,70 @@
...
+//
+
+static constexpr const char* kCompressionMetadataString = "TFLM_COMPRESSION";
+
+enum class CompressionScheme : uint8_t {
+  kBinQuant,
+};
+
+// TODO(ddavis-2015): pack struct
+struct BinQuantData {
```
rascani, line 36: Lets just call it LookupTableData. The binning is just a method for determining the lookup table.

tensorflow/lite/micro/compression.h
ddavis-2015: The index bit width is unrelated to the value table element size.  The associated literature does not use more than 20 bins (5 bit indices).  I'm not currently seeing a use case for a bit width greater than 5.

tensorflow/lite/micro/compression.h
```diff
@@ -0,0 +1,70 @@
...
+  const void* value_table;             // Pointer into FlatBuffer Values.
+  uint8_t value_table_channel_stride;  // elements per channel
+  uint8_t compressed_bit_width : 3;    // 1 to 7 bits
+  bool is_per_channel_quantized : 1;   // tensor is per-channel quantized
+  bool use_alternate_axis : 1;         // shape default channel:
+                                       // 0 = first, 1 = last
+  uint8_t reserved : 3;
+};
+
+union CompressionData {
```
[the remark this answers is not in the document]
ddavis-2015: I will change this to a union of pointers.  More compatible with IDEs and debuggers without any explicit type-casting.

tensorflow/lite/micro/compression.h
ddavis-2015: done

rascani (approved):

PR #2648, opened 2024-07-29
build: begin using -std=c++17 and c17 in Makefile builds

rkuester:
Begin using -std=c++17 and -std=c17 in Makefile builds on
all platforms. Bazel builds have been using C++17 since 52c9568.

Set `-stdlib=libc++ on xt-clang` on Xtensa to add C++17 library
support in addition to compiler support. From the xt-clang docs:

    Starting with the RI-2019.1 release, XT-CLANG has included support
    for C++14 and C++17 language features. The compiler support for
    C++14 and C++17 is accompanied by the C++ library from the LLVM
    project. This library can be selected with the -stdlib=libc++
    option, and this is strongly recommended when compiling with
    -std=c++14 or - std=c++17. Starting with the RI-2021.6 release, two
    additional versions of this C++ library are provided— one that
    excludes support for exception handling, and one that excludes both
    exception handling and run-time type identification. These libraries
    can be selected with -stdlib=libc ++-e and -stdlib=libc++-re options
    respectively.

Based on the `docker run` command in
tflite-micro/.github/workflows/xtensa_presubmit.yml, our CI is
currently using RI-2020.4.

Refactor the make/ext_libs/xtensa_download.sh script to make it eaiser
to patch downloads for all Xtensa platforms. The old script made overly
strict assumptions about the name of the patch.

Patch the Xtensa vision_p6 platform download xi_tflmlib_vision_p6 for
compatibility with the C++ library standard. Use the header <climits> to
access constants such as INT_MAX.

BUG=#2650


rkuester:
Closing and punting, for now, to issue #2650.

rascani (approved):

PR #2669, opened 2024-08-20
Add CMSIS-NN int8 and int16 batch matmul

ArmRyan:
* Moves some common functions with ref to new header file
 * Creates new cmsis_nn batch_matmul.cc

Authored-by:    Ryan O'Shea <ryan.oshea3@arm.com>
Co-authored-by: Adrian Lundell <adrian.lundell@arm.com>

BUG=Add BatchMatmul kernels to cmsisnn

### DIFF
--- a/tensorflow/lite/micro/kernels/batch_matmul.cc
+++ b/tensorflow/lite/micro/kernels/batch_matmul.cc
@@ -1,4 +1,4 @@
-/* Copyright 2023 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2024 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


tensorflow/lite/micro/kernels/batch_matmul.cc
```diff
@@ -271,22 +204,14 @@ TfLiteStatus TransposeRowsColumns(const TfLiteEvalTensor& tensor_in,
   return kTfLiteError;
```
suleshahid, line 204: `TransposeRowsColumns` and `TransposeRowsColumnsImpl` is moved to the header file right? We should de-dupe from here?

ArmRyan: Nice catch, should be fixed now 😄

suleshahid (approved):

Issue #2681, opened 2024-09-09
Add int16 support to RELU6


andresovela:
This is already supported in TF Lite:

https://github.com/tensorflow/tensorflow/blob/53df8c0156cfb3c3a318cfe70bb24f44b5ff8f9e/tensorflow/lite/kernels/activations.cc#L1003-L1005

PR #2682, opened 2024-09-09
Add int16 support to OP_RELU6

andresovela:
This PR adds int16 support to OP_RELU6. This is already supported in TF Lite.

bug=#2681


tensorflow/lite/micro/kernels/activations.cc
```diff
@@ -70,12 +70,15 @@ TfLiteStatus ReluEval(TfLiteContext* context, TfLiteNode* node) {
 
 void* Relu6Init(TfLiteContext* context, const char* buffer, size_t length) {
   TFLITE_DCHECK(context->AllocatePersistentBuffer != nullptr);
-  return context->AllocatePersistentBuffer(context, sizeof(Relu6OpData));
+  // Allocate enough memory to hold Relu6OpData<int16_t>, this is big enough for
+  // int8_t too
+  // TODO: How to improve this? Maybe make Relu6OpData hold int16s by default?
```
rascani, line 75: I would suggest eliminating the template on Relu6OpData and just using the largest type we'd expect. In this case, int32_t or just int is probably fine to use for `zero` and `six`. Thus the allocation is always the same size, regardless of the type. 

Since we'll be using a larger type to store the values, we should also add an ENSURE check on the output of `FloatToQuantizedType` to ensure that it remains within the bounds of int16_t or int8_t. That should allow us to safely narrow cast before we use it.

rascani (commented):
Thank you so much for the PR! Sorry for the delayed review.

tensorflow/lite/micro/kernels/activations.cc
andresovela: Done

rascani (approved):
Thanks again!

PR #2691, opened 2024-09-19
update the tensorflow.lite to use ai-edge-litert for all python scripts

ecalubaquib:
to update all usage from tensorflow.lite to ai-edge-litert
which has similar functionality with tflite_runtime

bug=2691

### DIFF
--- a/third_party/python_requirements.txt
+++ b/third_party/python_requirements.txt
@@ -11,6 +11,9 @@ absl-py==2.0.0 \
     #   keras
     #   tensorboard
     #   tensorflow
+ai-edge-litert==1.0.1 \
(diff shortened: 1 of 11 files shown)


third_party/python_requirements.txt
rascani, line 1: `ai-edge-litert` also needs to be added to `third_party/python_requirements.in`. Please add that, then run `bazel run //third_party:python_requirements.update`

rascani (commented):
Thanks for the PR!

third_party/python_requirements.txt
ecalubaquib: thanks for review. I added ai-edge-litert to requirements.in and ran the bazel update above. Committed the update as well

rascani (dismissed):

rascani (approved):

PR #1835, opened 2023-03-20
Increase kMaxEvents and error out when it's exceeded

paulinesho:
The current behavior silently resets and overwrites the internal list, this PR changes it to error out instead.

BUG=b/264555999

PR #2724, opened 2024-10-18
Profiler improvements

andresovela:
This PR adds a few improvements to the `MicroProfiler` class.

bug=#2725

First, I made it a template so that it's possible to adjust the maximum number of events the profiler can handle via a template parameter.

The motivation for this change is to save memory. #1835 increased the maximum number of events from 1024 to 4096, which is a _huge_ waste of RAM. The `MicroProfiler` class has these buffers:

```cpp
const char* tags_[kMaxEvents];
uint32_t start_ticks_[kMaxEvents];
uint32_t end_ticks_[kMaxEvents];

struct TicksPerTag {
  const char* tag;
  uint32_t ticks;
};

TicksPerTag total_ticks_per_tag[kMaxEvents];
```

We need 20 bytes per event, so an instance of `MicroProfiler` uses 80 KB of RAM, even if the model you're profiling has only a few nodes.

This PR also adds the following enum to specify the logging format and remove some duplicated code and clean the API.

```cpp
enum class MicroProfilerLogFormat {
  HumanReadable,
  Csv,
};
```

I replaced `LogTicksPerTagCsv()` with `LogGrouped()`, which does essentially the same thing, but also prints in a human readable format which looks like this:

```
Cumulative event times:
Count    Tag                              Ticks        Time
6        DEPTHWISE_CONV_2D                3085702      5.194 ms
2        SUM                              1281206      2.156 ms
66       FULLY_CONNECTED                  934142       1.572 ms
10       CONV_2D                          765357       1.288 ms
67       QUANTIZE                         731923       1.232 ms
20       CONCATENATION                    362212       0.609 ms
76       STRIDED_SLICE                    335828       0.565 ms
3        RELU                             245643       0.413 ms
4        DEQUANTIZE                       156682       0.263 ms
6        PAD                              145834       0.245 ms
29       ADD                              130229       0.219 ms
65       RESHAPE                          49593        0.083 ms
13       MUL                              48852        0.082 ms
4        SUB                              16371        0.027 ms
6        LOGISTIC                         10336        0.017 ms
3        TANH                             4761         0.008 ms
3        UNPACK                           2738         0.004 ms

Total time: 13.985 ms (8307409 ticks)
```

Finally I also added `TicksToUs()` to micro_time.h for convenience.


tensorflow/lite/micro/micro_profiler.h
```diff
@@ -16,15 +16,28 @@ limitations under the License.
...
+  HumanReadable,
+  Csv,
+};
+
 // MicroProfiler creates a common way to gain fine-grained insight into runtime
-// performance. Bottleck operators can be identified along with slow code
+// performance. Bottleneck operators can be identified along with slow code
 // sections. This can be used in conjunction with running the relevant micro
 // benchmark to evaluate end-to-end performance.
+template <int MAX_EVENTS>
```
ddavis-2015, line 40: This should be `template <int MAX_EVENTS = 4096>` so that all existing applications using the profiler do not need to be changed.

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -33,67 +46,185 @@ class MicroProfiler : public MicroProfilerInterface {
...
+        for (int i = 0; i < num_events_; ++i) {
+#if defined(HEXAGON) || defined(CMSIS_NN)
+          int ticks = end_ticks_[i] - start_ticks_[i];
+          MicroPrintf("%d,%s,%u,%d", i, tags_[i], TicksToMs(ticks), ticks);
+#else
+          uint32_t ticks = end_ticks_[i] - start_ticks_[i];
+          uint64_t us = TicksToUs(ticks);
+          MicroPrintf("%d,%s,%" PRIu64 ".%" PRIu64 ",%" PRIu32, i, tags_[i],
+                      us / 1000, us % 1000, ticks);
+#endif
```
ddavis-2015, line 118: Changing the CSV format probably won't pass review, as it most likely breaks countless internal Google, and external third-party scripts.

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -33,67 +46,185 @@ class MicroProfiler : public MicroProfilerInterface {
...
+  uint32_t start_ticks_[MAX_EVENTS];
+  uint32_t end_ticks_[MAX_EVENTS];
   int num_events_ = 0;
+  int num_tag_groups_ = 0;
 
-  struct TicksPerTag {
+  struct TagGroup {
     const char* tag;
     uint32_t ticks;
+    uint32_t tag_count;
```
ddavis-2015, line 182: Only used by the "human readable" log format.

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -33,67 +46,185 @@ class MicroProfiler : public MicroProfilerInterface {
...
+    for (int i = 0; i < num_tag_groups_ - 1; ++i) {
+      for (int j = i + 1; j < num_tag_groups_; ++j) {
+        if (tag_groups_[j].ticks > tag_groups_[i].ticks) {
+          TagGroup temp_tag_group = tag_groups_[i];
+          tag_groups_[i] = tag_groups_[j];
+          tag_groups_[j] = temp_tag_group;
+        }
+      }
+    }
+  }
```
ddavis-2015, line 219: Please use standard C++ library algorithms (as long as they don't require memory allocation).  Here `std::sort` should be used.

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -33,67 +46,185 @@ class MicroProfiler : public MicroProfilerInterface {
...
   // Marks the end of an event associated with event_handle. It is the
   // responsibility of the caller to ensure than EndEvent is called once and
   // only once per event_handle.
   //
   // If EndEvent is called more than once for the same event_handle, the last
-  // call will be used as the end of event marker. If EndEvent is called 0 times
+  // call will be used as the end of event marker.If EndEvent is called 0 times
   // for a particular event_handle, the duration of that event will be 0 ticks.
-  virtual void EndEvent(uint32_t event_handle) override;
+  void EndEvent(uint32_t event_handle) {
```
ddavis-2015, line 71: This is an override of the abstract base class, and should remain as such.

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -33,67 +46,185 @@ class MicroProfiler : public MicroProfilerInterface {
...
+    for (int i = 0; i < num_events_; ++i) {
+      ticks += end_ticks_[i] - start_ticks_[i];
+    }
+    return ticks;
+  }
 
   // Prints the profiling information of each of the events in human readable
   // form.
-  void Log() const;
+  void Log(MicroProfilerLogFormat format) const {
```
ddavis-2015, line 95: Should have a default argument of `MicroProfilerLogFormt::Csv` so that all existing applications continue to compile without modification.

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -33,67 +46,185 @@ class MicroProfiler : public MicroProfilerInterface {
...
+          uint32_t ticks = end_ticks_[i] - start_ticks_[i];
+          uint64_t us = TicksToUs(ticks);
+          MicroPrintf("%s took %" PRIu64 ".%" PRIu64 " ms (%u ticks)", tags_[i],
+                      us / 1000, us % 1000, ticks);
+        }
+        break;
 
-  // Prints the profiling information of each of the events in CSV (Comma
-  // Separated Value) form.
-  void LogCsv() const;
```
ddavis-2015, line 61: `LogCsv` should be retained, and should call `Log(MicroProfilerFormat::Csv)`, so that existing applications do not require modification.

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -33,67 +46,185 @@ class MicroProfiler : public MicroProfilerInterface {
...
+                    us / 1000, us % 1000, total_ticks);
+        break;
+      }
+      case MicroProfilerLogFormat::Csv: {
+        MicroPrintf("\"Tag\",\"Total ms\",\"Total ticks\"");
+        for (int i = 0; i < num_tag_groups_; ++i) {
+          uint64_t us = TicksToUs(tag_groups_[i].ticks);
+          MicroPrintf("%s, %" PRIu64 ".%" PRIu64 ", %u", tag_groups_[i].tag,
+                      us / 1000, us % 1000, tag_groups_[i].ticks);
+        }
```
ddavis-2015, line 165: Changing the CSV format probably won't pass review, as it most likely breaks countless internal Google, and external third-party scripts.

ddavis-2015:
@andresovela 

Thanks for your PR submission (and all the spelling corrections!).

This will require some discussion with @suleshahid.  I like the template change for sizing the timing data.

tensorflow/lite/micro/micro_profiler.h
andresovela: What should I do about that? #1290 introduced

```
struct TicksPerTag {
    const char* tag;
    uint32_t ticks;
  };
  // In practice, the number of tags will be much lower than the number of
  // events. But it is theoretically possible that each event to be unique and
  // hence we allow total_ticks_per_tag to have kMaxEvents entries.
  TicksPerTag total_ticks_per_tag[kMaxEvents];
```

which was "only" used by the CSV log format.

ddavis-2015: @andresovela 

The comment is really just there for @suleshahid to see.

tensorflow/lite/micro/micro_profiler.h
andresovela: Why CSV though? This method logs in human readable by default?

tensorflow/lite/micro/micro_profiler.h
andresovela: Done

tensorflow/lite/micro/micro_profiler.h
andresovela: Removed the `Ms` column

tensorflow/lite/micro/micro_profiler.h
andresovela: Done

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -33,67 +46,185 @@ class MicroProfiler : public MicroProfilerInterface {
   // Marks the start of a new event and returns an event handle that can be used
   // to mark the end of the event via EndEvent. The lifetime of the tag
   // parameter must exceed that of the MicroProfiler.
-  virtual uint32_t BeginEvent(const char* tag) override;
+  uint32_t BeginEvent(const char* tag) {
```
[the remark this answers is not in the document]
andresovela: Done

tensorflow/lite/micro/micro_profiler.h
andresovela: Done

tensorflow/lite/micro/micro_profiler.h
andresovela: I added `HumanReadable` as default argument since that's what would keep the existing behavior while continuing to compile

tensorflow/lite/micro/micro_profiler.h
andresovela: Done

tensorflow/lite/micro/micro_profiler.h
```diff
@@ -33,67 +47,174 @@ class MicroProfiler : public MicroProfilerInterface {
...
+      uint32_t ticks = end_ticks_[i] - start_ticks_[i];
+      tag_group.tag = tags_[i];
+      tag_group.ticks += ticks;
+      tag_group.tag_count++;
+    }
 
-  // Prints  total ticks for each unique tag in CSV format.
-  // Output will have one row for each unique tag along with the
-  // total ticks summed across all events with that particular tag.
-  void LogTicksPerTagCsv();
```
ddavis-2015, line 66: This method cannot be removed as it breaks existing applications.

andresovela: Added it back

andresovela: Is there a policy to _never_ break stuff like this? IMO duplicating APIs like this is also not great.

ddavis-2015: @andresovela Let's just say, that at the present time, no one is allowed to break anything. :slightly_smiling_face:

andresovela:
Oops, closed by mistake.

andresovela:
Waiting for review

andresovela:
Waiting for review

ddavis-2015:
@andresovela Thank you for your interest in TFLM (RTLM) and for submitting this PR!
Also, thank you for your persistence and perseverance.

While I think this is a good change to the profiler, it is still in internal discussion.

andresovela:
Updated

ddavis-2015:
While this PR has some very nice changes, a decision on a different approach to the MicroProfiler has been reached.

Issue #2740, opened 2024-10-30
CMSIS-NN Batch Matmul does not transpose shapes correctly with params set


ArmRyan:
CMSIS-NN and Tflite expect shapes to be transposed differently when adj_y is set. There should be a different case when the reference op is used as a fallback.

PR #2741, opened 2024-10-30
Fix issue with transposing shape in CMSIS-NN batch matmul

ArmRyan:
BUG=#2740


suleshahid:
Is there any test to catch the issue? If no existing, could we add one, that fails without the fix?

mansnils:
Some more context.. This was actually caught by one of the unit tests. It was an out of bound read but with GCC it was actually passing by "luck". However not when using armclang. So it was actually seen during testing of this PR and without the fix this new tests will fail: https://github.com/tensorflow/tflite-micro/pull/2750

suleshahid:
Ah I see. Thanks for the context.

suleshahid (approved):

PR #2742, opened 2024-10-30
build(codegen): suppress noise in console output

rkuester:
build(codegen): suppress noise in console output

Redirect output from the code_generator binary so that when it's
used within the build system, it doesn't print unexpected,
distracting noise to the console. Generally, compiler or
generator commands don't print output unless there's an error.

This reverts "build(codegen): suppress noise in console output
(#2708)", commit d2495773, which attempted to fix the same
problem by adding a --quiet flag. A subsequent upgrade to the
Tensorflow Python package caused new noise that wasn't possible
to suppress with the flag.

BUG=description


rkuester:
A motivation here is that I'm running a lot of unit tests locally as I made changes, and I'm trying to reduce spurious warnings so that the important ones stand out.

suleshahid:
Could you take a look at the failing Bazel CI?

rkuester:
This succeeded without changes by simply kicking off CI a few more times. Perhaps there's a race in the bazel build, now exaggerated by the (slight) changes this PR makes to the codegen genrule's action? Perhaps some dependency is underspecified?

It's hard to say more since the logging of the failure wasn't very good. I've submitted PR #2749 to improve the logging. I suggest we merge that PR, and then try to get this PR to fail by running CI a few more times. If it still fails, we'll get better logs.

rkuester:
@suleshahid This is ready to merge.

The original failure was either some kind of temporary infrastructure problem, or race I can't get to repeat. The logging improvements merged in #2749 should help us catch future failures like this in other PRs.

suleshahid (approved):

suleshahid:
[@Mergifyio](https://github.com/Mergifyio) refresh

suleshahid:
[@Mergifyio](https://github.com/Mergifyio) refresh

PR #2743, opened 2024-10-30
build(bazel): add integrity check to nnlib_hifi4 download

rkuester:
build(bazel): add integrity check to nnlib_hifi4 download

Add an integrity check to the http_archive() download of nnlib_hifi4 in
order to make the build more hermetic, reduce the security risk that a
remote file changes, and silence the noisy warning on the console during
the build:

    DEBUG: Rule 'nnlib_hifi4' indicated that a canonical reproducible
    form can be obtained by modifying arguments integrity[....]

BUG=description


rkuester:
A motivation here is that I'm running a lot of unit tests locally as I made changes, and I'm trying to reduce spurious warnings so that the important ones stand out.

suleshahid (approved):

Issue #2752, opened 2024-11-07
Add CMSIS-NN support for int8 Min/Max


ArmRyan:
Add tflm support for recent min/max functions in cmsis-nn

PR #2753, opened 2024-11-07
CMSIS-NN Min Max int8 support

ArmRyan:
* Moves common functions to new maximum_minimum.h
  * Creates cmsis-nn/maximum_minimum.cc

BUG=#2752
Change-Id: Ifbb3fedf53043b2f8d4c48d73c2ca44c7f0f87ca


mansnils:
CMSIS-NN SHA is already up to date so cmsis_nn_download.sh script should not be updated. Preferably quant specific registration should also be supported.

ArmRyan:
commenting to remove stale

tensorflow/lite/micro/kernels/cmsis_nn/maximum_minimum.cc
```diff
@@ -0,0 +1,247 @@
...
+    default:
+      MicroPrintf("Type %s (%d) is not supported by Maximum/Minimum.",
+                  TfLiteTypeGetName(op_context.output->type),
+                  op_context.output->type);
+      return kTfLiteError;
+  }
+  return kTfLiteOk;
+}
+
+TfLiteStatus EvalMaximumInt8(TfLiteContext* context, TfLiteNode* node) {
```
suleshahid, line 97: Is there any point for the separate int8 functions? Seems like the generic Eval handles the INT8 case as done here as well?

ArmRyan: Yeah the int8 specific registration has a smaller code size, since it wont include the float kernels in the translation units

suleshahid: Ah makes sense, thanks for clarifying.

suleshahid (approved):

PR #2757, opened 2024-11-08
Add support for CMSIS-NN int8 transpose and padding operators 

AdrianLundell:
BUG=New operator support


mansnils:
CMSIS-NN SHA is already up to date so cmsis_nn_download.sh script should not be updated. Also AddTranspose in micro_mutable_op_resolver.h should be updated to support quant specific registration.

mansnils (approved):

tensorflow/lite/micro/kernels/transpose_common.cc
```diff
@@ -0,0 +1,52 @@
...
+#include "tensorflow/lite/kernels/internal/tensor_ctypes.h"
+#include "tensorflow/lite/kernels/internal/types.h"
+#include "tensorflow/lite/kernels/kernel_util.h"
+#include "tensorflow/lite/micro/kernels/kernel_util.h"
+#include "tensorflow/lite/micro/kernels/transpose.h"
+#include "tensorflow/lite/micro/micro_log.h"
+
+namespace tflite {
+
+TfLiteStatus TransposePrepare(TfLiteContext* context, TfLiteNode* node) {
```
suleshahid, line 26: We should remove the duplicate function in transpose.cc and only use this.

tensorflow/lite/micro/kernels/transpose.h
```diff
@@ -0,0 +1,63 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#ifndef TENSORFLOW_LITE_MICRO_KERNELS_TRANSPOSE_H_
+#define TENSORFLOW_LITE_MICRO_KERNELS_TRANSPOSE_H_
+
+#include "tensorflow/lite/c/common.h"
+#include "tensorflow/lite/kernels/internal/reference/transpose.h"
```
suleshahid, line 20: Do we need all these includes? Is there something we duplicate from the internal transpose.h, if so, could we just use it directly from there?

AdrianLundell:
Thanks for the review, should be all good now

tensorflow/lite/micro/kernels/transpose.cc
```diff
@@ -19,6 +19,7 @@ limitations under the License.
 #include "tensorflow/lite/kernels/internal/types.h"
 #include "tensorflow/lite/kernels/kernel_util.h"
 #include "tensorflow/lite/micro/kernels/kernel_util.h"
+#include "tensorflow/lite/micro/kernels/transpose.h"
 #include "tensorflow/lite/micro/micro_log.h"
 
 namespace tflite {
```
suleshahid, line 25: sorry for late review, just catching this now. We should remove the common code below and use the one directly from common.

ddavis-2015: @AdrianLundell I believe @suleshahid is referring to the required removal of the code below in favor of the identical code in the transpose.h file:
```
constexpr int kInputTensor = 0;
constexpr int kPermTensor = 1;
constexpr int kOutputTensor = 0;
struct TransposeContext {
  TransposeContext(TfLiteContext* context, TfLiteNode* node) {
    micro_context = GetMicroContext(context);
    input = micro_context->AllocateTempInputTensor(node, kInputTensor);
    perm = micro_context->AllocateTempInputTensor(node, kPermTensor);
    output = micro_context->AllocateTempOutputTensor(node, kOutputTensor);
  }
  ~TransposeContext() {
    micro_context->DeallocateTempTfLiteTensor(input);
    micro_context->DeallocateTempTfLiteTensor(perm);
    micro_context->DeallocateTempTfLiteTensor(output);
  }
  MicroContext* micro_context;
  TfLiteTensor* input;
  TfLiteTensor* perm;
  TfLiteTensor* output;
};
```

tensorflow/lite/micro/kernels/transpose.cc
ddavis-2015, line 1: Only the tensor indexing constants from transpose.h should be used in this code.

ddavis-2015 (changes requested):

tensorflow/lite/micro/micro_mutable_op_resolver.h
ddavis-2015, line 1: This file didnt actually change, so perhaps it can be removed from the PR?

mansnils: Yes good point

ddavis-2015:
@AdrianLundell CI code style check does not pass

ddavis-2015 (approved):
All good here, on to @suleshahid

suleshahid (approved):

Issue #2761, opened 2024-11-12
Log error when node invocation is unsuccessful


andresovela:
Currently a log is only printed for `kTfLiteError` status codes.

andresovela:
#2762 is still open

andresovela:
#2762 is still open

PR #2762, opened 2024-11-12
Print invoke status on unsuccessful invocation

andresovela:
bug=#2761


tensorflow/lite/micro/micro_interpreter_graph.cc
```diff
@@ -224,12 +224,10 @@ TfLiteStatus MicroInterpreterGraph::InvokeSubgraph(int subgraph_idx) {
     // prepare for the next call.
     allocator_->ResetTempAllocations();
 
-    if (invoke_status == kTfLiteError) {
+    if (invoke_status != kTfLiteOk) {
       MicroPrintf("Node %s (number %d) failed to invoke with status %d",
                   OpNameFromRegistration(registration), current_operator_index_,
                   invoke_status);
```
ddavis-2015, line 230: @andresovela 

This will cause extraneous debug output for the `kTfLiteAbort` status code.  The `kTfLiteAbort` status code is used by the circular buffer kernel, the layer-by-layer debug tool, and the Generic Benchmark application.

I would suggest adding another `if` statement at line 228 to enable debug output only if the status code is not `kTfLiteAbort`.

andresovela: I wanted to fix this but there's no such thing as `kTfLiteAbort` it seems:

https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/core/c/c_api_types.h#L74-L120

andresovela: Ah I found it https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/micro_context.h#L29-L30

A suspicious cast if you ask me, there should be better documentation on why that exists.

andresovela:
Waiting for review

andresovela:
Done

ddavis-2015 (approved):
@suleshahid Everything looks good to me.

suleshahid (approved):

PR #2765, opened 2024-11-13
build(bazel): introduce tflm_cc_* macros, refactoring away micro_copts

rkuester:
Remove micro_copts() by replacing every cc_* target that used
them with a tflm_cc_* equivalent, and setting those common copts
in one place, inside the tflm_cc_* macro.

This is the first of several commits introducing tflm_cc_* macros
in place of cc_binary, cc_library, and cc_test. Motivated by the
upcoming need to support conditional compilation, the objective
is to centralize build configuration rather than requiring (and
remembering that) each cc_* target in the project add the same
common attributes such as compiler options and select()ed

Alternatives such as setting global options on the command line
or in .bazelrc, even if simplified with a --config option, fail
to preserve flags and hooks for configuration in the case TFLM is
used as an external repository by an application project. Nor is
it easy in that case for individual targets to override an
otherwise global setting.

BUG=#2636

### DIFF
--- a/codegen/build_def.bzl
+++ b/codegen/build_def.bzl
@@ -1,6 +1,6 @@
 """ Build rule for generating ML inference code from TFLite model. """
 
-load("//tensorflow/lite/micro:build_def.bzl", "micro_copts")
+load("//tensorflow/lite/micro:build_def.bzl", "tflm_cc_library")
 
 def tflm_inference_library(
         name,
@@ -25,7 +25,7 @@ def tflm_inference_library(
         visibility = ["//visibility:private"],
     )
 
-    native.cc_library(
+    tflm_cc_library(
         name = name,
         hdrs = [name + ".h"],
         srcs = [name + ".cc"],
@@ -39,6 +39,5 @@ def tflm_inference_library(
             "//tensorflow/lite/micro:micro_common",
             "//tensorflow/lite/micro:micro_context",
         ],
-        copts = micro_copts(),
         visibility = visibility,
     )
--- a/codegen/runtime/BUILD
+++ b/codegen/runtime/BUILD
@@ -1,12 +1,11 @@
-load("//tensorflow/lite/micro:build_def.bzl", "micro_copts")
+load("//tensorflow/lite/micro:build_def.bzl", "tflm_cc_library")
 
 package(default_visibility = ["//visibility:public"])
 
-cc_library(
+tflm_cc_library(
     name = "micro_codegen_context",
     srcs = ["micro_codegen_context.cc"],
     hdrs = ["micro_codegen_context.h"],
-    copts = micro_copts(),
     deps = [
         "//tensorflow/lite/c:common",
         "//tensorflow/lite/kernels:op_macros",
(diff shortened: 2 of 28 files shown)


suleshahid:
Could you expand on the conditional compilations? Will we have different tflm_copts and the idea is the user will just change that rather than change all micro_copts?

rkuester:
> Could you expand on the conditional compilations? Will we have different tflm_copts and the idea is the user will just change that rather than change all micro_copts?

Yes. Looking ahead, the conditional complication will be implemented through these macros by a change like 884a234ba6. There's a central place to make a change like that once the macros in this commit (`tflm_cc_*`) are in place.

`micro_copts` just wasn't general enough—it allowed modification of compiler options, but not modification of other attributes of a target. Plus, a developer had to remember to add micro_copts. The historical inconsistency throughout the codebase confirms that wasn't a reasonable expectation. It's much easier to outlaw naked `cc_*` targets, and catch those in review or even with tooling.

suleshahid:
I see. Unfortunately theres some internal code that use `micro_copts` as well, which could take some time to change. What's the alternatives, would we just need to add a flag to BUILDs when you want compression like `USE_TFLM_COMPRESSION`?

rkuester:
> I see. Unfortunately theres some internal code that use `micro_copts` as well, which could take some time to change. What's the alternatives, would we just need to add a flag to BUILDs when you want compression like `USE_TFLM_COMPRESSION`?

The original `micro_copts` lives on as `tflm_copts` (it's named that way to match the naming of the tflm_cc_* macros). I'd suggest the internal code substitute `s/micro_copts/tflm_copts` throughout. I could resurrect the name `micro_copts` for internal code to call, but that's ugly.

If this internal code needs to build with compression turned on---yes, ultimately it needs to have USE_TFLM_COMPRESSION defined somehow. If they don't want to adopt our tflm_cc_* macros wholesale, they could use the upcoming `tflm_defines()` in their define attribute. Or, the ugliest option that might work would be duplication the define in `micro_copts` maintained for just that internal code.

rkuester:
After talking in person, we have decided to resurrect micro_copts() as an alias for out-of-tree code.

suleshahid:
Do we need to keep `tflm_copts` then? Can we just keep the name `micro_copts` as it is and use that.

rkuester:
I'm inclined to leave them both, because:
1. Future commits along this feature branch already use `tflite_copts()`.
2. It's good we maintain a seam along which our internal, TFLM code and the out-of-tree code can differ. Some upcoming code adds to `tflite_copts()`, and this allows us the option of modifying `micro_copts()` differently.
3. The way the defaults can be overridden or augmented in these `tflite_cc_*` macros is to pass arguments which override the default set of options. So, e.g., you'll see upcoming code like `tflm_cc_test(copts = tflm_copts() + ["-foo"], defines = tflm_defines() + ["LOCAL"])`. It's nice if all these calls use the same *tflm_* prefix.
4. It's good we make it obvious in the code that `micro_copts` is deprecated and might diverge from our internal use.

suleshahid (approved):

PR #2768, opened 2024-11-15
build(bazel): replace cc_* with tflm_cc_* in remaining TFLM code

rkuester:
Replace cc_* targets remaining in TFLM code with tflm_cc_*
targets. These are targets which did not formerly use the common
copts. Avoid changing imported TFLite code, if for no other
reason than to avoid merge conflicts during the automatic sync
with upstream TFLite.

BUG=#2636

### DIFF
--- a/tensorflow/lite/micro/BUILD
+++ b/tensorflow/lite/micro/BUILD
@@ -405,7 +405,7 @@ tflm_cc_library(
     ],
 )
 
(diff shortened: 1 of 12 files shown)


suleshahid (approved):

rkuester:
@Mergifyio refresh

PR #2776, opened 2024-11-18
Fix Arm(R) Corstone(TM)-300 FVP CI script

mansnils:
Benchmark target is not supported for armclang.

Also the cortex-m-* workflows are updated to correctly include PR in checkout, if tested from a PR.

BUG=https://github.com/tensorflow/tflite-micro/issues/2775#issuecomment-2481913933


tensorflow/lite/micro/tools/ci_build/test_cortex_m_corstone_300.sh
```diff
@@ -43,13 +43,15 @@ readable_run make -f tensorflow/lite/micro/tools/make/Makefile clean
...
-readable_run make -j$(nproc) -f tensorflow/lite/micro/tools/make/Makefile \
-  CO_PROCESSOR=ethos_u \
-  OPTIMIZED_KERNEL_DIR=${OPTIMIZED_KERNEL_DIR} \
-  TARGET=${TARGET} \
-  TARGET_ARCH=${TARGET_ARCH} \
-  TOOLCHAIN=${TOOLCHAIN} \
-  GENERIC_BENCHMARK_MODEL_PATH=tensorflow/lite/micro/models/person_detect_vela.tflite \
-  GENERIC_BENCHMARK_ARENA_SIZE=`expr 150 \* 1024` \
-  run_tflm_benchmark
+# Run generic benchmark. Not supported for armclang - see comment in target makefile.
```
ddavis-2015, line 46: @mansnils 

You mention a comment in the Generic Benchmark application makefile?  We would be very interested in hearing more about this issue.

ddavis-2015: @mansnils 

Never mind, I found the makefile with the clang comment.  I now understand this change.

ddavis-2015 (approved):
@suleshahid 

Looks good to me.
Also, fixes the CI full-run build.

ddavis-2015 (commented):
@mansnils @suleshahid 

While this fix looks correct, it still fails the ARM Compiler 6 Corstone 300 CI (ci:full_run).

mansnils:
> @mansnils @suleshahid
> 
> While this fix looks correct, it still fails the ARM Compiler 6 Corstone 300 CI (ci:full_run).

That is strange. The fix should make this command pass: `tensorflow/lite/micro/tools/ci_build/test_cortex_m_corstone_300.sh armclang`
Although from the c:full_run log it looks like the original error message is still there.

mansnils:
> > @mansnils @suleshahid
> > While this fix looks correct, it still fails the ARM Compiler 6 Corstone 300 CI (ci:full_run).
> 
> That is strange. The fix should make this command pass: `tensorflow/lite/micro/tools/ci_build/test_cortex_m_corstone_300.sh armclang` Although from the c:full_run log it looks like the original error message is still there.

@Evagan2018 Could it be that the Arm Compiler 6 workflow does not include the SHA (PR) during the test?

MatthiasHertel80:
We are looking into the issue now.

mansnils:
@suleshahid I suggest we merge it and if Arm Compiler 6 workflow still fails we can revert.

suleshahid (approved):

suleshahid:
https://github.com/Mergifyio refresh

suleshahid (approved):

PR #2918, opened 2024-11-20
ci(bazel): simplify scripts and configuration in advance of additions

rkuester:
Simplify the Bazel-related CI scripts and BUILD configuration in
preparation for adding several more for the compression feature.

- Move CC=clang to .bazelrc

- Define CXX in case it's already defined in the environment, in
which case it overrides deferring to CC

- Replace --repo_env, which doesn't affect build, with --action_env

- Move commented-out ubsan invocation to its own CI job. Since
ubsan was commented out, Bazel invocations have moved to their
own jobs so they can run in parallel.

- Consistently test all targets, `...`, since TFLM is no longer
confined to the tensorflow/lite/micro directory. Filter out
inappropriate new targets with no.san tags.

- Don't run `bazel build ...` before `bazel test ...`; the latter
already builds all non-test targets.

- Remove obsolete filter for `no_oss`, there were no more
`no_oss` tags in the tree.

- Stop using readable_run: it has diverged from its original
meaning and now only redirects stderr to stdout. This is not
needed.

BUG=see description

### DIFF
--- a/tensorflow/lite/micro/tools/ci_build/test_bazel.sh
+++ b/tensorflow/lite/micro/tools/ci_build/test_bazel.sh
@@ -17,24 +17,5 @@
 set -e
 set -x
 
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-ROOT_DIR=${SCRIPT_DIR}/../../../../..
-cd "${ROOT_DIR}"
-
-source tensorflow/lite/micro/tools/ci_build/helper_functions.sh
-
-# We are using a bazel build followed by bazel test to make sure that the CI
-# covers non-test binary targets as well. These were previousbly covered by
-# having build_test but that was removed with #194.
-
-CC=clang readable_run bazel build ... \
-  --config=ci \
-  --build_tag_filters=-no_oss
-CC=clang readable_run bazel test ... \
-  --config=ci \
-  --test_tag_filters=-no_oss --build_tag_filters=-no_oss \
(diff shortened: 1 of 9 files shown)


tensorflow/lite/micro/tools/ci_build/test_bazel.sh
```diff
@@ -17,24 +17,5 @@
...
-  --build_tag_filters=-no_oss
-CC=clang readable_run bazel test ... \
-  --config=ci \
-  --test_tag_filters=-no_oss --build_tag_filters=-no_oss \
-  --test_output=errors
-
-# TODO(b/178621680): enable ubsan once bazel + clang + ubsan errors are fixed.
-#CC=clang readable_run bazel test tensorflow/lite/micro/... --config=ubsan --test_tag_filters=-no_oss,-noubsan --build_tag_filters=-no_oss,-noubsan
-
+bazel test ... \
```
suleshahid, line 20: `...` will only do from the current directory and below right? I think we should still include `cd root_dir`?

rkuester: Yes, `...` is from pwd down; however, CI jobs already run from the root_dir.

suleshahid: I would lean toward just leaving the previous code for cd root_dir, its the most clear for a reader, rather than needing to check where CI runs from.

suleshahid (approved):

PR #2980, opened 2024-12-02
refactor: move misplaced TF_LITE_REMOVE_VIRTUAL_DELETEs to private:

rkuester:
Move several TF_LITE_REMOVE_VIRTUAL_DELETE declarations that are
wrongly in a public section of their classes. To have the intended
effect, as documented in t/l/m/compatibility.h, these must be in a
private section.

BUG=see description

### DIFF
--- a/tensorflow/lite/micro/arena_allocator/non_persistent_arena_buffer_allocator.h
+++ b/tensorflow/lite/micro/arena_allocator/non_persistent_arena_buffer_allocator.h
@@ -74,8 +74,6 @@ class NonPersistentArenaBufferAllocator : public INonPersistentBufferAllocator {
   // takes in account any temporary allocations.
   size_t GetAvailableMemory(size_t alignment) const override;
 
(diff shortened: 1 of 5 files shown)


suleshahid:
Seems like these were moved out of private intentionally? https://github.com/tensorflow/tflite-micro/commit/e814105b965315b2123c73b12fbe2a0171176520

rkuester:
If the right fix for whatever that problem was is to make the destructors  `public:` accessible, it certainly breaks the intent, naming, and comments relating to the define TF_LITE_REMOVE_VIRTUAL_DELETE. It might be a sign those classes are called in unintended ways. If the destructors are indeed desired, they should probably be defined without the macro.

This commit is actually from quite a while ago, and I've forgotten now whether I had a need to fix it, or whether I just spotted the inconsistency and cleaned up. I'll drop this PR for now, and resurrect it if this becomes necessary for the compression feature #2636.

PR #2988, opened 2024-12-05
ci(Make): add test jobs for compression

rkuester:
Add GitHub CI jobs that run `make test`, etc. with compression
enabled. Add as new jobs, as opposed to additional steps within
the existing jobs, so they can run in parallel, and so they begin
with a fresh, un-built working directory.

BUG=part of #2636


rkuester:
This PR adds CI jobs.

suleshahid (approved):

PR #3001, opened 2024-12-09
Fixed scratch size calculation for conv for HiFi targets for scenarios when input, filter and output heights are 1.

pramods-cad:
Fixed scratch size calculation for conv for HiFi targets for scenarios when input, filter and output heights are 1.


ddavis-2015:
replaced with PR#3004

PR #3004, opened 2024-12-10
Fixed scratch size calculation for conv for HiFi targets for scenarios when input, filter and output heights are 1.

cad-audio:
Fixed scratch size calculation for conv for HiFi targets for scenarios when input, filter and output heights are 1.
BUG=conv_hifi_scratch_size


suleshahid:
Thanks for this PR, could you run the formatting script, and also add a BUG=something to the description.

suleshahid (approved):

PR #3006, opened 2024-12-11
feat(compression): implement tensor decompression in op fully_connected

rkuester:
Implement tensor decompression in op fully_connected. Extend
tests to validate operation on compressed tensors.

Include Xtensa extensions for decompression, as they are required
by the tests.

Fix the effect on memory_arena_threshold_test by setting a
different expected value for the persistent buffer allocation
when compression is configured in. The allocation was allowed to
vary by 3%; however, compression adds ~10%. Set the expected
value to the measured value when compression is configured in.

BUG=part of #2636


suleshahid:
Hifi compression tests seem to fail for this

rkuester:
> I have updated the compress-testing branch with the correctly merged code. I diffed compress-testing against my own branch and found no other relevant differences.

That still failed. I'll in the process of testing my ultimate destination in CI to see if it's due to something I've left out of this PR's commit.

ddavis-2015:
> > I have updated the compress-testing branch with the correctly merged code. I diffed compress-testing against my own branch and found no other relevant differences.
> 
> That still failed. I'll in the process of testing my ultimate destination in CI to see if it's due to something I've left out of this PR's commit.

@rkuester The test script requires the optimized kernels, but the PR does not have compression support for the optimized kernels.

rkuester:
> The test script requires the optimized kernels, but the PR does not have compression support for the optimized kernels.

I see. That's a surprising way to fail---build and runtime success, but wrong results. Is there an way we can transform this into a failure to build (an extension that should support compression having a new signature or something), or at least a better failure at runtime (an extension noticing the length of the tensor doesn't make sense, etc.)?

tensorflow/lite/micro/kernels/fully_connected.cc
```diff
@@ -115,9 +143,18 @@ TfLiteStatus FullyConnectedEval(TfLiteContext* context, TfLiteNode* node) {
           tflite::micro::GetTensorShape(input),
           tflite::micro::GetTensorData<float>(input),
           tflite::micro::GetTensorShape(filter),
+#ifdef USE_TFLM_COMPRESSION
+          tflite::micro::GetTensorData<float>(micro_context, filter,
```
suleshahid, line 147: Catching this a bit late, but in the new GetTensorData function, we check and return nullptr if tensor is nullptr. However, in the existing code, its a TFLITE_DCHECK, which means we will fail here, since we expected data. I think its better to do it that way, because otherwise we could end up trying to use the nullptr somewhere.

tensorflow/lite/micro/tools/make/targets/xtensa_makefile.inc
```diff
@@ -97,4 +97,10 @@ ifeq ($(OPTIMIZED_KERNEL_DIR), xtensa)
   $(TENSORFLOW_ROOT)tensorflow/lite/micro/kernels/xtensa/lstm_eval.cc \
   $(TENSORFLOW_ROOT)tensorflow/lite/micro/kernels/xtensa/lstm_eval_hifi.cc \
   $(TENSORFLOW_ROOT)tensorflow/lite/micro/kernels/xtensa/unidirectional_sequence_lstm.cc
+
+  # override KERNEL_OPTIMIZATION_LEVEL to enable higher performance
+  # Xtensa intrinsics.
+$(KERNEL_OBJDIR)$(TENSORFLOW_ROOT)tensorflow/lite/micro/kernels/xtensa/decompress.o: $(TENSORFLOW_ROOT)tensorflow/lite/micro/kernels/xtensa/decompress.cc
```
yushhuang, line 103: Curious if it's not sufficient to put this in `MICROLITE_CC_KERNEL_SRCS`? I understand that a dedicated rule was  required when the decompression routine was in `micro_context.cc`, but I'm not sure if it's needed now.

tensorflow/lite/micro/memory_arena_threshold_test.cc
```diff
@@ -87,6 +85,12 @@ uint8_t test_conv_tensor_arena[kTestConvModelArenaSize];
 constexpr int kTestConvModelTensorCount = 15;
 constexpr int kTestConvModelNodeAndRegistrationCount = 7;
 
+#if defined(USE_TFLM_COMPRESSION)
+constexpr int kKeywordModelPersistentBufferDataSize = 920;
```
yushhuang, line 89: Is the buffer size expected to grow by this much, even when the model does not utilize compression? I'm a bit curious where the overhead comes from.

rkuester:
@ddavis-2015, I'll let you comment on these reviews. (Note, I'm mostly passing through the C++ code on David's behalf at this point.)

tensorflow/lite/micro/memory_arena_threshold_test.cc
suleshahid: I believe it comes from the new FCOpData for compression scratch indices. It's behind the compression ifdef. So if you enable compression define, but don't use compression you will see the overhead, but if you don't enable compression build, it should remain same as before. Needs to remain persistent because you need the data at different times (prepare/eval).

tensorflow/lite/micro/kernels/fully_connected.cc
ddavis-2015: This is valid for cases where the bias tensor is optional as input to the operator.

tensorflow/lite/micro/tools/make/targets/xtensa_makefile.inc
ddavis-2015: Cadence requires us to use special compile options just for this one file.  Using -O3 on all kernels do NOT work.

tensorflow/lite/micro/memory_arena_threshold_test.cc
ddavis-2015: @suleshahid Correct.

tensorflow/lite/micro/kernels/fully_connected.cc
suleshahid: In that case I would prefer if we have another override for the optional case. Since this will be basically changing the code even for non-compression execution (in the case globally compression is enabled), it will be safer to have the nullptr DCHECK. 

If you think that will take too long, for now we could just add a single DCHECK for the weights data in the above ifdef compression code.

tensorflow/lite/micro/tools/make/targets/xtensa_makefile.inc
```diff
@@ -97,4 +97,10 @@ ifeq ($(OPTIMIZED_KERNEL_DIR), xtensa)
   $(TENSORFLOW_ROOT)tensorflow/lite/micro/kernels/xtensa/lstm_eval.cc \
   $(TENSORFLOW_ROOT)tensorflow/lite/micro/kernels/xtensa/lstm_eval_hifi.cc \
   $(TENSORFLOW_ROOT)tensorflow/lite/micro/kernels/xtensa/unidirectional_sequence_lstm.cc
+
+  # override KERNEL_OPTIMIZATION_LEVEL to enable higher performance
```
suleshahid, line 101: should we have if compression_enabled here?

ddavis-2015: I don't think so.  I don't believe the primary makefile allows for conditional exclusion of files in the optimized kernel directories.

tensorflow/lite/micro/kernels/fully_connected.cc
suleshahid: Seeing how many files use it, probably easiest to just add the override GetOptionalTensorData.

ddavis-2015: in progress.

suleshahid (approved):

PR #3012, opened 2024-12-13
feat: enhance profiling and benchmarking

rkuester:
Modify `PrintMemoryPlan` in `greedy_memory_planner.cc` for better
handling of tensor indices and scratch buffers.

Fix `total_ticks_per_tag_` usage in `micro_profiler.cc` and add
`ClearEvents` method.

Update `Makefile.inc` and `generic_model_benchmark.cc` to support
alternate memory regions and CRC32 checks for data integrity.

Include compression data in `metrics.cc` allocation records and
handle architecture-specific directives in
`show_meta_data.cc.template`.

BUG=see description


tensorflow/lite/micro/tools/benchmarking/generic_model_benchmark.cc
```diff
@@ -130,39 +154,145 @@ bool ReadFile(const char* file_name, void* buffer, size_t buffer_size) {
...
+  }
+}
+
+uint32_t ComputeCRC32(const uint8_t* data, const size_t data_length) {
+  uint32_t crc32 = ~0U;
+
+  for (size_t i = 0; i < data_length; i++) {
+    // crctab is an array of 256 32-bit constants
+    const uint32_t index = (crc32 ^ data[i]) & 0xFF;
+    crc32 = (crc32 >> 8) ^ crctab[index];
```
suleshahid, line 179: should it be add a check for index out of bounds here?

tensorflow/lite/micro/tools/benchmarking/generic_model_benchmark.cc
```diff
@@ -130,39 +154,145 @@ bool ReadFile(const char* file_name, void* buffer, size_t buffer_size) {
...
+  // For streaming models, the interpreter will return kTfLiteAbort if the
+  // model does not yet have enough data to make an inference. As such, we
+  // need to invoke the interpreter multiple times until we either receive an
+  // error or kTfLiteOk. This loop also works for non-streaming models, as
+  // they'll just return kTfLiteOk after the first invocation.
   uint32_t seed = kRandomSeed;
   while (true) {
     SetRandomInput(seed++, interpreter);
-    TfLiteStatus status = interpreter.Invoke();
+    ShowInputCRC32(&interpreter);
```
suleshahid, line 292: Do we need to see these CRC32? Maybe we can add these in a follow up later standalone, aren't needed for anything compression right?

tensorflow/lite/micro/tools/benchmarking/generic_model_benchmark.cc
rkuester: The index will always be in bounds by virtue of the mask. I've added a constant in 
650e89d to ensure the relationship between the array size and the indexing is clear.

tensorflow/lite/micro/tools/benchmarking/generic_model_benchmark.cc
rkuester: Yes, the CRC32 value is incredibly useful. It's the sole output that enables verification that the output tensor is the same across various benchmark runs.

suleshahid: Gotcha, I would probably leave benchmarking and accuracy validation as separate, but I guess it doesn't hurt to just quickly check it here.

suleshahid (approved):

PR #3014, opened 2024-12-15
feat(compression): implement tensor decompression in op concatenation

rkuester:
Implement tensor decompression in op concatenation. Extend
tests to validate operation on compressed tensors.

BUG=part of #2636

### DIFF
--- a/tensorflow/lite/micro/kernels/concatenation.cc
+++ b/tensorflow/lite/micro/kernels/concatenation.cc
@@ -1,4 +1,4 @@
-/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2024 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -33,6 +33,13 @@ constexpr int kOutputTensor = 0;
 
 struct OpData {
   ConcatenationParams params;
+
+#ifdef USE_TFLM_COMPRESSION
+
+  // scratch buffers for compressed tensors
+  int scratch_indices[kMaxInputNum];
+
+#endif  // USE_TFLM_COMPRESSION
 };
 
 // Handles negative axis index, coerces to positive index value.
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/kernels/concatenation.cc
```diff
@@ -155,62 +176,53 @@ TfLiteStatus ConcatenationPrepare(TfLiteContext* context, TfLiteNode* node) {
...
+    if (input_type == kTfLiteInt8) {
+      // Make sure there is no re-scaling needed for Int8 quantized kernel. This
+      // is a restriction we introduced to Int8 kernels.
+      TF_LITE_ENSURE_EQ(context, static_cast<double>(input->params.scale),
+                        static_cast<double>(output_tensor->params.scale));
+      TF_LITE_ENSURE_EQ(context, input->params.zero_point,
+                        output_tensor->params.zero_point);
+    } else if (input_type == kTfLiteInt16) {
+      // Make sure that all Int16 inputs have a null zero-point.
+      TF_LITE_ENSURE_EQ(context, input->params.zero_point, 0);
```
yushhuang, line 189: Is there a reason why the scale is not checked in the int16 case? AFAICS the (at least the reference) concatenation kernel does not take into account any scaling/quantization at all.

https://github.com/tensorflow/tflite-micro/blob/b2f27186d243d6afc2324c43b3bf9a9f681be474/tensorflow/lite/kernels/internal/reference/concatenation.h#L30

yushhuang (approved):

tensorflow/lite/micro/kernels/concatenation.cc
ddavis-2015: I agree with your conclusion, however this is a copy-paste of the TfLite code.

tensorflow/lite/micro/kernels/concatenation.cc
```diff
@@ -155,62 +176,53 @@ TfLiteStatus ConcatenationPrepare(TfLiteContext* context, TfLiteNode* node) {
...
-        TF_LITE_ENSURE(context, t != nullptr);
-        input_scales[i] = t->params.scale;
-        input_zero_points[i] = t->params.zero_point;
-        micro_context->DeallocateTempTfLiteTensor(t);
-      }
-
-      data->params.input_scale = input_scales;
-      data->params.input_zeropoint = input_zero_points;
-      data->params.output_zeropoint = output->params.zero_point;
-      data->params.output_scale = output->params.scale;
```
suleshahid, line 204: Why do we not need to set these anymore?

suleshahid: Never mind, I followed the code, yea we don't use this, only for the other ConcatWithScaling which TFLM doesn't use.

suleshahid (approved):

tensorflow/lite/micro/kernels/concatenation.cc
tinskip: Yes, also is there are reason why it has to be zero? Seems that concatenation in general shouldn't care about quantization parameters, as long as they match in and out.

ddavis-2015: @tinskip This same check is performed by the LiteRT (TfLite) reference implementation.

PR #3015, opened 2024-12-15
feat(compression): implement tensor decompression in op conv

rkuester:
Implement tensor decompression in op conv. Extend tests to
validate operation on compressed tensors.

BUG=part of #2636

### DIFF
--- a/tensorflow/lite/micro/kernels/conv.cc
+++ b/tensorflow/lite/micro/kernels/conv.cc
@@ -1,4 +1,4 @@
-/* Copyright 2023 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2024 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


yushhuang (approved):

tensorflow/lite/micro/kernels/conv.cc
```diff
@@ -79,9 +108,18 @@ TfLiteStatus ConvEval(TfLiteContext* context, TfLiteNode* node) {
...
+                                                 weights_comp_td,
+                                                 data.weights_scratch_index),
+            tflite::micro::GetTensorShape(bias),
+            tflite::micro::GetTensorData<int64_t>(
+                micro_context, bias, bias_comp_td, data.bias_scratch_index),
+#else   // USE_TFLM_COMPRESSION
             tflite::micro::GetTensorData<int8_t>(filter),
             tflite::micro::GetTensorShape(bias),
-            tflite::micro::GetOptionalTensorData<std::int64_t>(bias),
+            tflite::micro::GetTensorData<std::int64_t>(bias),
```
suleshahid, line 121: Changing this is intentional right?

ddavis-2015: yes, as the bias is checked for nullptr on line 63/83.

tensorflow/lite/micro/kernels/xtensa/conv_hifi.cc
```diff
@@ -177,9 +177,30 @@ TfLiteStatus ConvEvalHifiInt16(TfLiteContext* context, TfLiteNode* node,
...
+#ifdef USE_TFLM_COMPRESSION
+  const int8_t* filter_data = tflite::micro::GetTensorData<int8_t>(
+      micro_context, filter, weights_comp_td,
+      data.reference_op_data.weights_scratch_index);
+  const int64_t* bias_data = tflite::micro::GetOptionalTensorData<int64_t>(
+      micro_context, bias, bias_comp_td,
+      data.reference_op_data.bias_scratch_index);
+#else   // USE_TFLM_COMPRESSION
   const int8_t* filter_data = tflite::micro::GetTensorData<int8_t>(filter);
-  const int64_t* bias_data = tflite::micro::GetTensorData<int64_t>(bias);
```
suleshahid, line 182: should leave this one as it was? the function may be depending on the nullptr check. and we can use the GetTensorData for our compression equivalent as well here to force the same nullptr check.

tensorflow/lite/micro/kernels/xtensa/conv_hifi.cc
```diff
@@ -275,8 +295,26 @@ TfLiteStatus ConvEvalHifiInt8(TfLiteContext* context, TfLiteNode* node,
...
+
+  const CompressionTensorData* weights_comp_td =
+      micro_context->GetTensorCompressionData(node, kConvWeightsTensor);
+  const CompressionTensorData* bias_comp_td =
+      micro_context->GetTensorCompressionData(node, kConvBiasTensor);
+
+#endif  // USE_TFLM_COMPRESSION
+
   const int8_t* input_data = tflite::micro::GetTensorData<int8_t>(input);
-  const int32_t* bias_data = tflite::micro::GetTensorData<int32_t>(bias);
```
suleshahid, line 279: same here, use GetTensorData rather than Optional as it currently is.

tensorflow/lite/micro/kernels/xtensa/conv_hifi.cc
ddavis-2015: The bias tensor is optional, and is not validated inside this eval method.  No DCHECK should be performed, and this is a correction to the original Cadence code.  The NNLIB methods will correctly return error for no bias tensor data, and the eval method will return kTfLiteError regardless of the BUILD_TYPE environ. var.

tensorflow/lite/micro/kernels/xtensa/conv_hifi.cc
ddavis-2015: Please see previous comment.

tensorflow/lite/micro/kernels/xtensa/conv_hifi.cc
suleshahid: I see, thanks.

rkuester:
@mergifyio refresh

PR #3029, opened 2024-12-26
Bugfix: unchecked pointer use

kmnls:
**Description**
Pointer may be used unchecked.
API happens to return NULL if not enough memory but pointer is not checked and used in tensor population.


ddavis-2015:
[internal: fixed in PR #3058]

PR #3064, opened 2025-03-14
Makefile updates for Arm(R) CPU targets

AdrianLundell:
Removing redundant assignments for CMSIS_PATH and adding the possibility to override Device Family Pack (DFP) path with a new CORTEX_DFP_PATH make variable.

This allows flexibility in where the CMSIS and DFP sources reside.

BUG=More flexible build structure


ddavis-2015 (approved):
Looks fine to me.  On to @suleshahid.

Issue #3089, opened 2025-04-07
Preprocessor tests for __ANDROID__ instead of defined(__ANDROID__)


jgoldstn:
Found at https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/core/api/tensor_utils.cc#L36

The rest of the codebase tests to see if `__ANDROID__` is defined, whereas this tests for the value. This is causing some warnings to be errors in my local build and I assume this is in error.

ddavis-2015:
@jgoldstn TFLM (aka LRTM) imports some of its code from TfLite.  `tensor_utils.cc` is one such file.  As such, this issue (and associated PR) need to be submitted in the [tensorflow](https://github.com/tensorflow/tensorflow/issues) repository.

All files imported from TfLite into TFLM can be found [here](https://github.com/tensorflow/tflite-micro/blob/main/ci/tflite_files.txt).

PR #3090, opened 2025-04-07
Update tensor_utils.cc

jgoldstn:
Update preprocessor check for defined as opposed to value, fixing bug #3089


ddavis-2015:
`tensor_utils.cc` is imported from TfLite.

PR #3094, opened 2025-04-14
Remove hexdump_test.cc from source list

shlmregev:
As a unit test, it belongs on MICROLITE_TEST_SRCS, which gets filtered out of MICROLITE_CC_SRCS. Keeping it on the source list results in it being compiled into the tflm archive, where its main() function conflicts with the application's main.
Also, since hexdump.cc calls DebugVsnprintf(), we need to define it in debug_log.cc, also when TF_LITE_STRIP_ERROR_STRINGS is #defined, else compilation fails. Move the #ifdef TF_LITE_STRIP_ERROR_STRINGS to be inside the function implementation, rather around the whole function.

BUG=384562154

### DIFF
--- a/tensorflow/lite/micro/hexdump.cc
+++ b/tensorflow/lite/micro/hexdump.cc
@@ -17,7 +17,7 @@
 #include <algorithm>
 #include <cctype>
 
-#include "tensorflow/lite/micro/debug_log.h"
+#include "tensorflow/lite/micro/micro_log.h"
 #include "tensorflow/lite/micro/static_vector.h"
 
 namespace {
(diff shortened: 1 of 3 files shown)


shlmregev:
> @shlmregev I am thinking that running the tests in RELEASE builds is not something we do, as there is no point in running tests that can't tell you if they fail/succeed. So it seems all the changes to debug_log.cc are not needed.

That's a good point and I reverted the changes to debug_log.cc, but the 'x86 Release' CI check is currently failing because I added hexdump_test to the list of tests and DebugVsnprintf() isn't defined. Even if I define it to return 0, hexdump_test will fail. 
@suleshahid @rkuester any suggestions?

ddavis-2015:
> > @shlmregev I am thinking that running the tests in RELEASE builds is not something we do, as there is no point in running tests that can't tell you if they fail/succeed. So it seems all the changes to debug_log.cc are not needed.
> 
> That's a good point and I reverted the changes to debug_log.cc, but the 'x86 Release' CI check is currently failing because I added hexdump_test to the list of tests and DebugVsnprintf() isn't defined. Even if I define it to return 0, hexdump_test will fail. @suleshahid @rkuester any suggestions?

@shlmregev I believe I see the issue now.  In `hexdump.cc`, the `DebugVsnprintf` should be changed to `MicroVsnprintf` and the include of `debug_log.h` changed to `micro_log.h`.  The `micro_log.h` handles builds with `TF_LITE_STRIP_ERROR_STRINGS` such that they will not fail.  Both the bluepill and x86 release builds should then complete without error.

rkuester:
> I believe that test should only run when Compression is enabled in the config (as we can see Makefile x86 with Compression CI is passing). Is that correct @rkuester ?

Compression uses this feature, but the feature and its unit test are independent of compression. So no, I would not conditionally compile out the unit test. It should always run.

suleshahid (approved):

Issue #2650, opened 2024-07-31
build: C++17 is not enabled on all platforms


rkuester:
Bazel builds have been C++17 since 52c9568; however, Makefile builds remain C++11. Consistent with our webpage documentation and to match TF Lite's language support, it's our desire to support and compile as C++17 on all platforms.

rkuester:
After trying to set -std=c++17 and c17 in PR #2648, the only problem appears to be the age of the Xtensa toolchain. The flatbuffers library includes `<optional>` when it's being compiled under the C++17 standard. The Xtensa toolchain is happy enough to compile with -std=c++17, but the standard library appears to be incomplete—it doesn't have `<optional>`. E.g., from the PR's CI run:

https://github.com/tensorflow/tflite-micro/actions/runs/10152485116/job/28074001989?pr=2648#step:3:367
````
xt-clang++ -std=c++17 [...]
[....]
.../micro/tools/make/downloads/flatbuffers/include/flatbuffers/stl_emulation.h:41:12: fatal error:
'optional' file not found
  #include <optional>
           ^~~~~~~~~~
````
Perhaps upgrading the Xtensa toolchain, which is quite old, will fix this.

For reference, the current toolchain is identified and invoked here: https://github.com/tensorflow/tflite-micro/blob/6b676567b79053d4d275ae834427b4f1259a710d/.github/workflows/xtensa_presubmit.yml#L35

rascani:
I tested this out for at least the hifi3 builds and got it to compile with `-stdlib=libc++`. I think the default stdlib for Xtensa only supports C++14. The downside is that I think that pre-compiled library includes rtti & exception support, so it likely has some bloat. In newer versions of Xtensa toolchain, we can use `libc++-re` to disable those.

rkuester:
Fixed in #2648.

Issue #2058, opened 2023-06-15
c++17-only *inline variable* feature introduced in sync from upstream TF


rkuester:
The sync from upstream TF in PR #2053 brought the c++17-only feature *inline variables* into code widely included in the TFLM kernels: 

https://github.com/tensorflow/tflite-micro/blob/857842e6b184819011cb2cfe59899b7b97e22785/tensorflow/lite/core/macros.h#L68-L76

This causes the following warning in my bazel builds. Using inline variables might lead to build failures with other compilers, or at worst: unexpected behavior if the compiler ignores the intent of `inline` with or without warning. (Although in this case, the variables are const, so maybe it'll just waste space :shrug:.) . BTW, I'm working on turning on warnings into errors a la `-Werror` in #2057 so that a warning like this, subtly introduced in a sync from upstream TF, won't pass CI.

````
ERROR: /home/rkuester/wetu/tflite-micro/tensorflow/lite/micro/kernels/BUILD:146:11: Compiling tensorflow/lite/micro/kernels/micro_tenso
r_utils.cc failed: (Exit 1): gcc failed: error executing command (from target //tensorflow/lite/micro/kernels:micro_tensor_utils) /usr/
bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer '-std=c++
0x' -MD -MF ... (remaining 22 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from ./tensorflow/lite/kernels/internal/common.h:29,
                 from tensorflow/lite/micro/kernels/micro_tensor_utils.cc:25:
./tensorflow/lite/core/macros.h:70:1: error: inline variables are only available with '-std=c++17' or '-std=gnu++17' [-Werror=c++17-ext
ensions]
   70 | inline constexpr unsigned int flatbuffer_size_max = 2147483648;
      | ^~~~~~
./tensorflow/lite/core/macros.h:72:1: error: inline variables are only available with '-std=c++17' or '-std=gnu++17' [-Werror=c++17-ext
ensions]
   72 | inline constexpr char tflite_metadata_buffer_location[] = "buffer_location";
      | ^~~~~~
./tensorflow/lite/core/macros.h:74:1: error: inline variables are only available with '-std=c++17' or '-std=gnu++17' [-Werror=c++17-ext
ensions]
[...]

PR #3095, opened 2025-04-16
Revert back from C/C++17 to C/C++11

shlmregev:
There are older toolchains that don't support C++17 and are still used in shipping products.
Rewrite hexdump.cc to not use C++17 features (std::byte and Class Template Argument Deduction). Also remove forcing of -std=libc for xtensa targets, which was added because stdlib isn't supported with -std=c++17.

BUG=410831256


suleshahid:
@ddavis-2015 It seems like these are really the only files that are needing C++17? The makefile tests are all passing with C++11, do you know what else you would have expected to break?

rkuester:
We may not want to revert back to C++11 and commit to C++11 compliance going forward. That decision is not only about the code uniquely written for TFLM; we have dependencies outside our control---upstream TFLite from which we merge code regularly, libraries like flatbuffers, gemmlowp , ruy, nnlib, etc.

- TFLite [uses with C++17](https://github.com/google-ai-edge/LiteRT/blob/ce68efa039a74b9d25805286939c19d9a0343b48/.bazelrc#L296)
- flatbuffers [seems okay at C++11](https://github.com/google/flatbuffers/blob/609c72ca1a7526f444a109bc04f861ad578e833d/CMakeLists.txt#L62)
- ruy [uses with C++14](https://github.com/google/ruy/blob/83fd40d730feb0804fafbc2d8814bcc19a17b2e5/CMakeLists.txt#L19)
- nnlib seems to be C++11

We took the decision to be a C++17 project in 2023 in #2060, to address an issue #2058 merging C++17 code from TFLite. See the extensive discussion of whether to switch in that issue. Makefile-based builds were caught up in #2648. Issue #2650 has some discussion.

In any event, a commitment to C++11 would be a major shift of policy. It would have project-wide implications for what features we can use in our own code, and what workarounds we're committing to implement when our dependencies use non-C++11 features.

tensorflow/lite/micro/static_vector.h
rkuester, line 1: I'd hate to see `StaticVector` go. We've used it a few times in code that hasn't yet been merged (e.g., unit test refactoring).

suleshahid (approved):

PR #3114, opened 2025-06-02
Fix generate_cc_array treating empty CSV as single-element array

MROS:
Previously, an empty .csv file was incorrectly interpreted as a single-element array due to the behavior of str.split(',') on empty strings returning [''] (length 1). This patch explicitly checks for empty or whitespace-only lines and correctly returns a size of 0 and an empty array definition.


veblush:
Closed in favor of https://github.com/tensorflow/tflite-micro/pull/3423

Issue #3125, opened 2025-06-24
feature: add compression tools to Python package


rkuester:
This feature bug tracks the work of adding compression tools to our Python package, `tflite_micro.`

rkuester:
Noting #3129 was a part of this.

ddavis-2015:
@rkuester is this planned to happen with the DECODE operator changes?

rkuester:
> [@rkuester](https://github.com/rkuester) is this planned to happen with the DECODE operator changes?

Oops, looks like we left this issue open despite the feature becoming completed. This was the addition of the OG compression implementation to our PyPI package.

Yes, compression via the PyPI package must continue to work and not regress as we change over to the DECODE-based implementation. Please file another issue if it breaks.

PR #3126, opened 2025-06-24
fix: resolve TensorFlow keras lazy loader recursion error

rkuester:
Set TF_USE_LEGACY_KERAS=1 to force TensorFlow 2.19.0 to use tf_keras
instead of Keras 3.x if the build environment has both keras 3.10.0 and
tf_keras 2.19.0 installed. This prevents TensorFlow's lazy loader from
entering infinite recursion when trying to determine which Keras version
to use for tf.keras.models.Sequential().

This fixes runtime_test which was failing with:
"RecursionError: maximum recursion depth exceeded in comparison"
at line checking "if self._tfll_keras_version == 'keras_3'"

BUG=#3125


suleshahid:
Also is it not an issue for the CI because we have only one keras version in those testing environments?

rkuester:
Ugh.

Turns out I was wrong about the underlying problem. The workaround does indeed avoid importing Keras, but that "fixes" a different underlying problem. The underlying problem is that Keras imports the module `psutil` for whatever reason, and `psutil` has a crashing bug. See the backtrace below.

`psutil` is imported *if and only if* it happens to be in the environment, and it happened to be in my environment. Uninstalling `psutil` fixes everything without the `TF_USE_LEGACY_KERAS=1` workaround.

There appears to be no problem with Tensorflow coexisting with Keras >= 3. In fact, Keras >= 3 has been in our Bazel Python dependencies alongside Tensorflow 2.18 for some time now without any issues.

I will withdraw this PR, and open or reopen an issue about configuring Bazel's Python toolchain to be hermetic.

*Backtrace:*
```
[....]
  File "/tmp/bazel-working-directory/tflite_micro/bazel-out/k8-fastbuild/bin/python/tflite_micro/runtime_test.runfiles/tflm_pip_deps_keras/site-packages/keras/src/saving/saving_lib.py", line 37, in
 <module>
    import psutil
  File "/home/rkuester/.local/lib/python3.10/site-packages/psutil/__init__.py", line 95, in <module>
    from . import _pslinux as _psplatform
  File "/home/rkuester/.local/lib/python3.10/site-packages/psutil/_pslinux.py", line 25, in <module>
    from . import _psposix
  File "/home/rkuester/.local/lib/python3.10/site-packages/psutil/_psposix.py", line 50, in <module>
    'Negsignal', {x.name: -x.value for x in signal.Signals}
AttributeError: module 'signal' has no attribute 'Signals'
```

PR #3129, opened 2025-07-01
feat(python): add compression module to tflite_micro Python package

rkuester:
feat(python): add compression module to tflite_micro Python package

Integrate the TFLM compression tools into the tflite_micro Python
package, allowing users to compress models directly from Python code
that imports the package.

Usage:
  from tflite_micro import compression

Details:
- Add compress_lib py_library target in compression BUILD
- Create compression package with __init__.py exposing public API
- Include compression module in Python package dependencies
- Add compression dependencies to wheel requirements

BUG=see description

### DIFF
--- a/python/tflite_micro/BUILD
+++ b/python/tflite_micro/BUILD
@@ -130,6 +130,7 @@ py_package(
     # in the tflm tree.
     packages = [
         "python.tflite_micro",
+        "tensorflow.lite.micro.compression",
         "tensorflow.lite.micro.tools.generate_test_for_model",
         "tensorflow.lite.python",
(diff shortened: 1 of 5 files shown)


rkuester:
> Just wondering, would there be any benefit in us having a separate package for compression enabled vs disabled, similar to how we do it for local builds for bazel and make?

I can't imagine any benefit. The reason we provide the option to disable it in C++ builds is to minimize code space for builds that don't utilize compression, which can be meaningful in embedded use cases. Small savings in code space (a few KB) aren't meaningful in the Python-on-a-desktop use case.

suleshahid (approved):

PR #3141, opened 2025-07-11
Add const-tensor checks

### LINKED ISSUE #3140: Add const-tensor checks
@tensorflow/micro 

Add consistent const-tensor checks and error messages across these kernels:
TRANSPOSE
STRIDED_SLICE
FILL
BROADCAST_TO
EXPAND_DIMS

Modify associated kernel unit tests.

ddavis-2015:
@tensorflow/micro

Add consistent const-tensor checks and error messages across these kernels: TRANSPOSE
STRIDED_SLICE
FILL
BROADCAST_TO
EXPAND_DIMS

Modify associated kernel unit tests.

bug=fixes #3140

### DIFF
--- a/python/tflite_micro/python_ops_resolver.cc
+++ b/python/tflite_micro/python_ops_resolver.cc
@@ -99,6 +99,7 @@ PythonOpsResolver::PythonOpsResolver() {
   AddQuantize();
   AddReadVariable();
   AddReduceMax();
+  AddReduceMin();
   AddRelu();
   AddRelu6();
   AddReshape();
--- a/tensorflow/lite/micro/kernels/micro_ops.h
+++ b/tensorflow/lite/micro/kernels/micro_ops.h
@@ -100,6 +100,7 @@ TFLMRegistration Register_PRELU();
 TFLMRegistration Register_QUANTIZE();
 TFLMRegistration Register_READ_VARIABLE();
 TFLMRegistration Register_REDUCE_MAX();
+TFLMRegistration Register_REDUCE_MIN();
 TFLMRegistration Register_RELU();
 TFLMRegistration Register_RELU6();
 TFLMRegistration Register_RESHAPE();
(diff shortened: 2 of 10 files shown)


tensorflow/lite/micro/kernels/fill_test.cc
```diff
@@ -44,6 +44,10 @@ tflite::micro::KernelRunner CreateFillTestRunner(
   static TfLiteTensor tensors[3];
 
   tensors[0] = CreateTensor(dims_data, IntArrayFromInts(dims_shape));
+  if (dims_data != nullptr) {
+    // dims must be a const tensor
+    tensors[0].allocation_type = kTfLiteMmapRo;
+  }
```
rameshkunasi, line 50: Better to have this condition after all the tensor creations are done?

tensorflow/lite/micro/kernels/fill_test.cc
```diff
@@ -172,8 +174,11 @@ TF_LITE_MICRO_TEST(FillInt8NoInputDimsData) {
   int output_shape[] = {3, kDim1, kDim2, kDim3};
   int8_t output_data[kDim1 * kDim2 * kDim3];
 
-  TestFill(dims_shape, dims_data, value_shape, value_data, output_shape,
-           output_data);
+  tflite::micro::KernelRunner runner =
+      CreateFillTestRunner(dims_shape, dims_data, value_shape, value_data,
+                           output_shape, output_data);
```
rameshkunasi, line 179: I just want to understand few things here.
- Is this test was successful previously? 
- Do we want to make this test fail in the prepare stage? 
- Why can't we use TestFill() API here aswell?

tensorflow/lite/micro/kernels/fill_test.cc
ddavis-2015: The code only applies to the `dims` tensor, so it is colocated with the tensor creation.

tensorflow/lite/micro/kernels/fill_test.cc
ddavis-2015: The old kernel allowed the `dims` tensor to have changing values.  This is incorrect and was corrected in the kernel such that the `dims` tensor must be a const-tensor.

As a result, we now want a negative test, and it should fail during the `prepare` phase.

We dont use `TestFill` because it doesn't support negative tests.  We rarely rewrite old unit test code as it is not worth the effort (it would be better to plan an entirely new unit test code base).

Issue #3148, opened 2025-07-15
Update ethos-u links to move from mlplatform to gitlab


ArmRyan:
mlplatform is soon to be deprecated. Gitlab is the new way to access these repos

ddavis-2015:
resolved with PR #3149

PR #3149, opened 2025-07-15
Replace mlplatform links for ethos-u with updated gitlab links (#3148)

ArmRyan:
BUG=#3148
 * Updates all references from mlplatform  to gitlab when referring to ethos-u dependencies


ddavis-2015:
@ArmRyan Nothing on your end.  Google is looking into the problem.

PR #3179, opened 2025-08-19
[Clean-up] Clang-format some existing files

veblush:
- Updated the clang-format to the latest (21)
- Reformatted all existing files using this new clang-format

BUG=SImple-clean-up

### DIFF
--- a/.github/workflows/log_binary_size_pr.yml
+++ b/.github/workflows/log_binary_size_pr.yml
@@ -60,4 +60,4 @@ jobs:
          author: TFLM-bot <tflm-github-bot@google.com>
          body: "BUG=automated binary size log update"
          labels: ci:run
-         reviewers: advaitjain
+         reviewers: veblush
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,4 @@
 *          @tensorflow/micro
 
-/.github/  @advaitjain @rockyrhodes @suleshahid
-/ci/       @advaitjain @rockyrhodes @suleshahid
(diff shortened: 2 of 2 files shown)


tensorflow/lite/micro/kernels/reduce_common.cc
```diff
@@ -120,7 +120,7 @@ enum MinMaxEvalType { kEvalMin, kEvalMax };
 template <typename T>
 struct MinMaxReducerCompare {
   MinMaxReducerCompare() = delete;
-  MinMaxReducerCompare(MinMaxEvalType evalType) : type_(evalType){};
+  MinMaxReducerCompare(MinMaxEvalType evalType) : type_(evalType) {};
```
ddavis-2015, line 123: I believe the Google style rules say that a space is not allowed between the parenthesis and the brace, and my IDE with it's auto formatting agrees.  But the Pigweed style checker says that this is wrong and spacing must be there.  I am much interested in your final resolution of this.  This difference has only appeared in the last year or so.

veblush: I'm sticking with the current format because it follows the rule "There should be a space between the close parenthesis and the open curly brace." ([link](https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions) 

Given that the Pigweed style is implemented by clang-format (following Google conventions), the practical decision is to treat the tool's output as the final authority, overriding minor written style guide disagreements.

suleshahid (approved):

PR #3181, opened 2025-08-20
Detect and fail dynamic tensors

### LINKED ISSUE #3180: Detect and fail dynamic tensors
@tensorflow/micro 

Detect output tensors with dynamic shapes after the `Prepare` method for each operator in a subgraph has been called.  Upon detection of a dynamic tensor, return `kTfLiteError` from the `MicroInterpreter::AllocateTensors` method.

Add unit test for the `MicroInterpreter`.

ddavis-2015:
@tensorflow/micro

Detect output tensors with dynamic shapes after the Prepare method for each operator in a subgraph has been called. Upon detection of a dynamic tensor, return kTfLiteError from the MicroInterpreter::AllocateTensors method.

Add unit test for the MicroInterpreter.

Fix typos in Prepare failure message.

bug=fixes #3180

### DIFF
--- a/tensorflow/lite/micro/micro_interpreter_test.cc
+++ b/tensorflow/lite/micro/micro_interpreter_test.cc
@@ -778,4 +778,35 @@ TF_LITE_MICRO_TEST(TestGetTensorFailsNoLinearMemoryPlanner) {
   TF_LITE_MICRO_EXPECT(interpreter.GetTensor(0) == nullptr);
 }
 
+TF_LITE_MICRO_TEST(TestDynamicTensorFails) {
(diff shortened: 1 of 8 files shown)


tensorflow/lite/micro/micro_interpreter_test.cc
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2024 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2025 The TensorFlow Authors. All Rights Reserved.
```
veblush, line 1: Please keep the original year.

PR #3197, opened 2025-09-21
Support TfLite schema buffer and custom options offsets

### LINKED ISSUE #3196: Support TfLite schema buffer and custom options offsets
@tensorflow/micro 

Allow for models >2Gb (and less than 4Gb) in size, as generated by the TfLite converter.
Parse TfLite schema `Buffer` tables where the `offset` and `size` fields are active.
Parse TfLite schema `Operator` tables where the `large_custom_options_offset` and `large_custom_options_size` fields are active.
Correctly process the Offline Memory Planner metadata buffer.
Correctly process the compression metadata buffer.
Add unit tests for all of the above.

ddavis-2015:
@tensorflow/micro

Allow for models >2Gb (and less than 4Gb) in size, as generated by the TfLite converter. Parse TfLite schema Buffer tables where the offset and size fields are active. Parse TfLite schema Operator tables where the large_custom_options_offset and large_custom_options_size fields are active. Correctly process the Offline Memory Planner metadata buffer. Correctly process the compression metadata buffer. Add unit tests for all of the above.

bug=fixes #3196

### DIFF
--- a/tensorflow/lite/micro/micro_interpreter.cc
+++ b/tensorflow/lite/micro/micro_interpreter.cc
@@ -17,6 +17,7 @@ limitations under the License.
 #include <cstdarg>
 #include <cstddef>
 #include <cstdint>
+#include <limits>
 
 #include "flatbuffers/flatbuffers.h"  // from @flatbuffers
 #include "tensorflow/lite/c/c_api_types.h"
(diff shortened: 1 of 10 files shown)


tensorflow/lite/micro/micro_interpreter.cc
```diff
@@ -105,6 +106,14 @@ void MicroInterpreter::Init(MicroProfilerInterface* profiler) {
 }
 
 TfLiteStatus MicroInterpreter::PrepareNodeAndRegistrationDataFromFlatbuffer() {
+  // needed for custom options when model is larger than 2Gb
```
veblush, line 109: Is it 2GiB?

tensorflow/lite/micro/micro_interpreter.cc
```diff
@@ -105,6 +106,14 @@ void MicroInterpreter::Init(MicroProfilerInterface* profiler) {
...
+  const uint8_t* flatbuffer_start =
+      flatbuffers::GetBufferStartFromRootPointer(model_);
+  if (flatbuffer_start == nullptr) {
+    MicroPrintf("%s: Unable to locate flatbuffer start", __func__);
```
veblush, line 113: Isn't it okay not to have `__func__` here?

veblush (commented):
Thanks for the PR! I've left a couple of minor comments regarding `__func__` and `2Gb`.

Beyond those, I have two broader suggestions for improvement:

### Optimize GetBufferStartFromRootPointer Calls
I'm concerned that this function's internal loop could increase initialization time. Since this is for a niche use case, could we make it to avoid if possible? Avoiding this call when it's not needed would be great.

### Simplify Buffer Access Logic
The current approach to accessing the buffer requires handling two separate cases, which leads to some repetitive and lengthy code. To improve readability and maintainability, would it be possible to encapsulate this logic in a helper function or a macro?

PR #3199, opened 2025-09-22
Switch to LiteRT repo

gonnet:
Modify the `ci/sync_from_upstream_tf.sh` script to pull TFLite sources from `https://github.com/google-ai-edge/LiteRT.git` instead of `https://github.com/tensorflow/tensorflow.git`.

This requires fixing some `#include` paths as the source and target directory names have changed. This change takes care to preserve the local directory structure.

BUG=CI


third_party/python_requirements.txt
```diff
@@ -1,12 +1,12 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.13
```
veblush: I think we cannot bump this yet.

gonnet: On my machine, I was not able to build without updating the dependencies because it was unable to download `tensorflow==2.18.0`.

I'll check if this was due to my local Python version.

gonnet: I've now reverted this change.

PR #3201, opened 2025-09-23
[CI] Install qemu

veblush:
- Improved ci/Dockerfile.micro further by
  - Installing QEMU from packages
  - Optimizing the docker definition by combining multiple layers into single layer to reduce the footprint.

BUG=Clean-up

### DIFF
--- a/ci/Dockerfile.micro
+++ b/ci/Dockerfile.micro
@@ -3,92 +3,80 @@
 # This container includes all dependencies for TFLM CI checks and is
 # recommended for local debugging to ensure a consistent environment.
 #
-# * Pull a prebuilt image:
-#   See all versions: https://github.com/users/TFLM-bot/packages/container/tflm-ci/versions
+# === Usage ===
+#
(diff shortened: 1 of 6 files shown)


ci/Dockerfile.micro
```diff
@@ -3,92 +3,80 @@
...
+    zip \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install LLVM/Clang and create symlinks
+RUN wget https://apt.llvm.org/llvm.sh \
+    && chmod +x llvm.sh \
+    && ./llvm.sh 21 \
+    && rm llvm.sh \
+    && apt-get update && apt-get install -y clang-21 clang++-21 clang-format-21 \
```
gonnet, line 49: Shouldn't `./llvm.sh 21` already install `clang-21` and `clang++-21`? Or do you need to explicitly install them so that `apt` gets the dependencies for `clang-format-21` right?

gonnet (approved):

PR #3206, opened 2025-09-25
Disabled two tests in recording_micro_allocator_test.cc

veblush:
This is a temporary workaround to resolve a test failure on the main branch. This change should be reverted once the underlying issues are fixed and the tests are re-enabled.

The following tests are failing when executed with `qemu-arm`:
- `TestRecordsPersistentTfLiteTensorData`
- `TestRecordsPersistentTfLiteTensorQuantizationData`

Repro: (From the TFLM root directory)
```
$ docker run -it --rm -v $(pwd):/opt/tflm -w /opt/tflm ghcr.io/tflm-bot/tflm-ci:0.6.5  /bin/bash
# tensorflow/lite/micro/tools/ci_build/test_cortex_m_qemu.sh
# qemu-arm -cpu cortex-m3 gen/cortex_m_qemu_cortex-m3_default_cmsis_nn_gcc/bin/recording_micro_allocator_test
```

BUG=QuickWorkaroundOnMain

### DIFF
--- a/tensorflow/lite/micro/recording_micro_allocator_test.cc
+++ b/tensorflow/lite/micro/recording_micro_allocator_test.cc
@@ -157,6 +157,10 @@ TF_LITE_MICRO_TEST(TestRecordsMultiTenantAllocations) {
                           tensors_count * TF_LITE_EVAL_TENSOR_STRUCT_SIZE * 2);
 }
 
+// TODO(veblush): Reenable this
+// Currently those two tests are failing with
+// "qemu: uncaught target signal 7 (Bus error) - core dumped"
+#if 0
 TF_LITE_MICRO_TEST(TestRecordsPersistentTfLiteTensorData) {
   const tflite::Model* model = tflite::GetModel(kTestConvModelData);
   uint8_t arena[kTestConvArenaSize];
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/recording_micro_allocator_test.cc
```diff
@@ -157,6 +157,10 @@ TF_LITE_MICRO_TEST(TestRecordsMultiTenantAllocations) {
                           tensors_count * TF_LITE_EVAL_TENSOR_STRUCT_SIZE * 2);
 }
 
+// TODO(veblush): Reenable this
+// Currently those two tests are failing with
+// "qemu: uncaught target signal 7 (Bus error) - core dumped"
+#if 0
 TF_LITE_MICRO_TEST(TestRecordsPersistentTfLiteTensorData) {
   const tflite::Model* model = tflite::GetModel(kTestConvModelData);
```
ddavis-2015, line 165: @veblush I am suspicious of the existing kTestConvModelData (a generated convolutional model).  The model we are using is from 2021, and when regenerated using the repo script, generates a very different model today.  In my PR for schema buffer and custom option offsets, I generated a new model and placed it into the PR.

veblush: Thanks for the pointer. I also concluded that the pre-generated test model file is misaligned, which is the culprit. I'm regenerating it in a separate PR to confirm this.

veblush: See https://github.com/tensorflow/tflite-micro/pull/3207

PR #3210, opened 2025-09-26
Updated QEMU doc

veblush:
Updated the QEMU manual to have more info about using docker images and how to debug.

BUG=doc

### DIFF
--- a/tensorflow/lite/micro/docs/qemu.md
+++ b/tensorflow/lite/micro/docs/qemu.md
@@ -39,14 +39,14 @@ $ tensorflow/lite/micro/tools/ci_build/test_cortex_m_qemu.sh
 ```
 
 After this script completes the initial build, you can run an individual
-test target like the following:
+test target as follows:
 
 ```
 $ make -f tensorflow/lite/micro/tools/make/Makefile \
   TARGET=cortex_m_qemu \
   TARGET_ARCH=cortex-m3 \
   OPTIMIZED_KERNEL_DIR=cmsis_nn
-  <tTEST_TARGET>
+  <TEST_TARGET>
 ```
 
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/docs/qemu.md
```diff
@@ -9,27 +9,83 @@
...
+
+## Live GDB Debugging
+
+### Terminal 1: Start QEMU, telling it to wait for a GDB connection on port 1234.
+
+```
+$ qemu-arm -cpu cortex-m3 -g 1234 <TEST_BINARY_PATH>
+```
+
+### Terminal 2: Launch GDB and connect to 랙the waiting QEMU session.
```
gonnet: Spurious character?

tensorflow/lite/micro/docs/qemu.md
```diff
@@ -9,27 +9,83 @@
...
-processor.
+
+You can run all Cortex-M unit tests using a single build script:
+
+```
+$ tensorflow/lite/micro/tools/ci_build/test_cortex_m_qemu.sh
+```
+
+After this script completes the initial build, you can run an individual
+test target like the following:
```
gonnet: "like the following" -> "as follows"

gonnet (changes requested):

tensorflow/lite/micro/docs/qemu.md
veblush: Oops. Good catch!

tensorflow/lite/micro/docs/qemu.md
veblush: Ack

PR #3213, opened 2025-10-03
Support for DECODE operator

### LINKED ISSUE #3212: Support for DECODE operator
@tensorflow/micro

Add support for alternate decompression memory to DECODE operator.
Additional unit tests.
Update generic benchmark application and Makefile.

ddavis-2015:
@tensorflow/micro

Add support for alternate decompression memory to DECODE operator. Additional unit tests.
Update generic benchmark application and Makefile.

bug=fixes #3212

### DIFF
--- a/tensorflow/lite/micro/micro_interpreter_context_test.cc
+++ b/tensorflow/lite/micro/micro_interpreter_context_test.cc
@@ -209,6 +209,13 @@ TF_LITE_MICRO_TEST(TestSetDecompressionMemory) {
       alt_memory_region = {{g_alt_memory, kAltMemorySize}};
   TfLiteStatus status;
 
+  // Test that all of the MicroInterpreterContext fences are correct, by
+  // forcing the MicroInterpreterContext state. The SetDecompressionMemory
+  // method should only be allowed during the kInit state, and can only be
+  // set once.  This is because alternate decompression memory is allocated
+  // during the application initiated kPrepare state, and all memory has already
+  // been statically allocted during the kInvoke state.
+
   // fail during Prepare state
   micro_context.SetInterpreterState(
       tflite::MicroInterpreterContext::InterpreterState::kPrepare);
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/micro_interpreter_context_test.cc
```diff
@@ -199,4 +199,109 @@ TF_LITE_MICRO_TEST(TestGetTempIntermediateTensor) {
   TF_LITE_MICRO_EXPECT_TRUE(invalid_output == nullptr);
 }
 
+TF_LITE_MICRO_TEST(TestSetDecompressionMemory) {
```
veblush, line 202: Could you please add comments explaining why these specific state checks are necessary?

tensorflow/lite/micro/micro_interpreter_context_test.cc
```diff
@@ -199,4 +199,109 @@ TF_LITE_MICRO_TEST(TestGetTempIntermediateTensor) {
...
+  uint8_t* p = static_cast<uint8_t*>(micro_context.AllocateDecompressionMemory(
+      kAllocateSize, tflite::MicroArenaBufferAlignment()));
+  TF_LITE_MICRO_EXPECT(p == &g_alt_memory[0]);
+
+  // allocate next 10 bytes
+  p = static_cast<uint8_t*>(micro_context.AllocateDecompressionMemory(
+      kAllocateSize, tflite::MicroArenaBufferAlignment()));
+  TF_LITE_MICRO_EXPECT(p == &g_alt_memory[16]);
+
+  // fail next allocation
```
veblush: Please add a comment here clarifying that this fails due to 16-byte alignment overhead (consuming 26 of the 30 bytes after just two allocations). The first time I saw this, this was puzzling because 10 x 3 = 30 ;)

tensorflow/lite/micro/kernels/decode.cc
```diff
@@ -43,6 +68,8 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
   TfLiteTensor* output = nullptr;
   TfLiteStatus status = kTfLiteOk;
 
+  micro_context->ResetDecompressionMemoryAllocations();
```
veblush, line 71: Does this mean that there will be only one Decode node in the entire graph? It looks like that if there were two Decode nodes, the second one's Prepare call would wipe out the memory assignments planned by the first one.

tensorflow/lite/micro/kernels/decode.cc
```diff
@@ -18,12 +18,37 @@ limitations under the License.
...
 #include "tensorflow/lite/micro/micro_log.h"
 
 namespace tflite {
 namespace {
 
+TfLiteStatus SetOutputTensorData(TfLiteContext* context, const TfLiteNode* node,
+                                 size_t tensor_output_index,
+                                 TfLiteTensor* output) {
+  if (output->data.data != nullptr) {
+    // If memory has already been assigned to the tensor, leave it be
```
veblush, line 32: I'm curious how this would happen?

veblush (commented):
Thank you for the PR; I have some minor feedback.

Regarding the decompression memory: How is the required size calculated, and do we have tooling for that? Also, when is it preferable to use multiple decompression regions rather than just one?

tensorflow/lite/micro/kernels/decode.cc
ddavis-2015: \<discussion moved to alternate venue>

tensorflow/lite/micro/kernels/decode.cc
ddavis-2015: \<discussion moved to alternate venue>

tensorflow/lite/micro/micro_interpreter_context_test.cc
ddavis-2015: Fixed.

tensorflow/lite/micro/micro_interpreter_context_test.cc
ddavis-2015: Fixed.

PR #3216, opened 2025-10-07
Support for DECODE operator

### LINKED ISSUE #3215: Support for DECODE operator
@tensorflow/micro

Add support for custom registration to DECODE operator.
Additional unit tests.

ddavis-2015:
@tensorflow/micro

Add support for application custom DECODE operator registrations.
Additional unit tests.

bug=fixes #3215


tensorflow/lite/micro/micro_context.h
```diff
@@ -169,12 +171,40 @@ class MicroContext {
...
+    uint8_t reserved1;  // reserved
+    uint8_t reserved2;  // reserved
+    uint8_t reserved3;  // reserved
+    tflite::DecodeState* (*func)(const TfLiteContext*, MicroProfilerInterface*);
+  };
+
+  // Set the custom DECODE operator registrations.
+  // Can only be called during the kInit state.
+  virtual TfLiteStatus SetCustomDecodeRegistrations(
+      const std::initializer_list<CustomDecodeRegistration>& registrations) {
```
veblush: `SetCustomDecodeRegistrations` could pose a lifetime safety issue. Storing a pointer to a `std::initializer_list` is dangerous because the underlying array is often a temporary object that gets destroyed immediately after the function returns. Please update the API to accept a pointer to a persistent array and a size count (e.g., const CustomDecodeRegistration* registrations, size_t size) so the caller is explicitly responsible for ensuring the data's lifetime matches that of the interpreter. (also its comment should mention about the lifetime expectation)

With this change, this kind of help function would be helpful .

```
template <size_t N>
TfLiteStatus SetCustomDecodeRegistrations(const CustomDecodeRegistration (&regs)[N]) {
  return SetCustomDecodeRegistrations(regs, N);
}
```

We might want to make a similar change to `decompress_regions_` as well but it could be a separate PR.

tensorflow/lite/micro/micro_context.h
```diff
@@ -169,12 +171,40 @@ class MicroContext {
     return nullptr;
   }
 
+  struct CustomDecodeRegistration {
+    uint8_t type;       // custom decode type
+    uint8_t reserved1;  // reserved
+    uint8_t reserved2;  // reserved
+    uint8_t reserved3;  // reserved
+    tflite::DecodeState* (*func)(const TfLiteContext*, MicroProfilerInterface*);
```
veblush: Instead of `func`, `factory` or `create_state` would be better to describes what the function will do.

veblush (commented):
Thank you for the PR!

tensorflow/lite/micro/micro_context.h
ddavis-2015: Fixed.

tensorflow/lite/micro/micro_context.h
ddavis-2015: Fixed.

veblush (dismissed):

Issue #3071, opened 2025-04-01
There is no check on the number of tensor dimensions which leads to a buffer overflow and a crash if sum of all dimensions is too long


hadi81:
There is no check on the number of of tensor dimension which leads to a buffer overflow and a crash if it is too long e.g, the sum of all dimensions in (dim1, dim2, dim3, ...) should not exceed the available buffer size. I have attached the backtrace, model file to replicate the crash and decompiled json file.

This is related to memory corruption and arbitrary code execution. There shoud be checks before consuming this model file. Other SDKs are able to handle this and successfully report the error instead of crashing. There should be atleast followng checks:

1) Size of dimension in the input, output and intermdiate tensor should fall under valid limits. 
2) The size of dimension should not exceed the available buffer size in the buffer table

[backtrace.txt](https://github.com/user-attachments/files/19556989/backtrace.txt)
![Image](https://github.com/user-attachments/assets/263a3c24-7502-4919-9688-c120d87a09d6)

[sinewave_model.zip](https://github.com/user-attachments/files/19557005/sinewave_model.zip)

Issue #3076, opened 2025-04-01
Non Existent Buffer in Intermediate Tensor Leads to a Crash


hadi81:
If there is some non existent buffer index in intermediate tensor table, this will result in a crash. I have attached the backtrace, model file to replicate the crash and decompiled json file.

This is related to memory corruption and arbitrary code execution. There shoud be checks before consuming this model file. Other SDKs are able to handle this and successfully report the error instead of crashing.

[non_existent_buffer_in_intermediate_tensor.zip](https://github.com/user-attachments/files/19557365/non_existent_buffer_in_intermediate_tensor.zip)

Issue #3074, opened 2025-04-01
Non Existent Buffer Index in Operator Input Leads to a Crash


hadi81:
If there is some non existent buffer index in input of operator table, this will result in a crash. I have attached the backtrace, model file to replicate the crash and decompiled json file.
This is related to memory corruption and arbitrary code execution. There shoud be checks before consuming this model file. Other SDKs are able to handle this and successfully report the error instead of crashing.

[non_existent_input_buffer.txt](https://github.com/user-attachments/files/19557322/non_existent_input_buffer.txt)
![Image](https://github.com/user-attachments/assets/882fe257-9622-43bd-aaa6-61efecd92da8)

[sinewave_model.zip](https://github.com/user-attachments/files/19557326/sinewave_model.zip)

Issue #3073, opened 2025-04-01
Non Existent Buffer Index in Operator Output Leads to a Crash


hadi81:
If there is some non existent buffer index in output of operator table, this will result in a crash. I have attached the backtrace, model file to replicate the crash and decompiled json file.

This is related to memory corruption and arbitrary code execution. There shoud be checks before consuming this model file. Other SDKs are able to handle this and successfully report the error instead of crashing.

![Image](https://github.com/user-attachments/assets/6f5a9b39-f72c-412e-a1df-5b9d442fd166)
[bt.txt](https://github.com/user-attachments/files/19557216/bt.txt)

[sinewave_model.zip](https://github.com/user-attachments/files/19557222/sinewave_model.zip)

PR #3235, opened 2025-10-18
Fix incorrect buffer index failure

### LINKED ISSUE #3075: Non Existent Buffer in Output Tensor Leads to a Crash
If there is some non existent buffer index in output tensor table, this will result in a crash. I have attached the backtrace, model file to replicate the crash and decompiled json file.

This is related to memory corruption and arbitrary code execution. There shoud be checks before consuming this model file. Other SDKs are able to handle this and successfully report the error instead of crashing.

[non_existent_buffer_in_output_tensor.zip](https://github.com/user-attachments/files/19557354/non_existent_buffer_in_output_tensor.zip)

ilya-kupchenko:
Description:
bug=fixes #3075 #3076 #3077

Problem
MicroAllocator does not validate the buffer indices. If a model provides an out-of-bounds buffer index for an input, output, or intermediate tensor, it results in segmentation fault.

Solution
Add validation inside MicroAllocator to check that the buffer index for every tensor. If an invalid index is found, InitializeTfLiteTensorFromFlatbuffer and InitializeTfLiteEvalTensorFromFlatbuffer will now return kTfLiteError.

Files Modified
    tensorflow/lite/micro/micro_interpreter.cc
    tensorflow/lite/micro/micro_interpreter_test.cc

Testing
Added new Bazel tests that use models with invalid buffer-to-tensor mappings.


tensorflow/lite/micro/micro_allocator.cc
```diff
@@ -230,6 +230,19 @@ TfLiteStatus InitializeTfLiteTensorFromFlatbuffer(
...
 
+  // Validate buffer index before any dereference of the buffers vector.
+  if (buffers == nullptr) {
+    MicroPrintf("Model buffers vector is null");
+    return kTfLiteError;
+  }
+  const uint32_t buffer_index = flatbuffer_tensor.buffer();
+  if (buffer_index >= buffers->size()) {
+    MicroPrintf(
+        "Tensor references invalid buffer index %u, model has only %d buffers",
```
gonnet: All the other calls to `MicroPrintf()` have strings ending in `\n`.

ddavis-2015: @gonnet Every call to MicroPrintf() will generate a newline as the last character output.  After changing the way debug output worked a few years ago, we kept the old behavior.

ilya-kupchenko:
I squashed the commit messages, fixed the style and added end-of-line symbols, just in case. Some MicroPrintf() calls in micro_allocator.cc have them, some not. Let me know if you want me to add/remove them for all MicroPrintf() calls it in all the project in a unified way. 
If you allow, I will fix #3071, #3072, #3073, #3074, #3128and #2721. Let me know if this is OK (or what other things I can be useful for).
PS. While trying to run the CI/CD tests I ran into some errors. I described them here #3240. If this is a bug, I can push the patch for it.
Thank you!

Issue #3253, opened 2025-11-16
Feature: Add RISC-V 32 Bare-Metal Full-System Emulation Support with QEMU and UART


sshahi-mips:
This feature adds comprehensive bare-metal support for RISC-V 32-bit targets, enabling TensorFlow Lite Micro (TFLM) applications to run under full-system QEMU emulation with UART-based console output.

Key Components:

- Minimal Bootloader (start.s) for system initialization and stack setup.
- Bare-metal linker script (linker.ld) defining memory layout and section mapping.
- Stub implementations for syscall-free environments.
- Lightweight UART driver for serial logging and debugging.
- Target-specific Makefile with integrated QEMU run support.
- README documentation to guide users in building and running demos.

This introduces a dedicated layer under tensorflow/lite/micro/riscv32_baremetal/, supporting examples such as person_detection, micro_speech, dtln and hello_world. Overall, this enhancement enables fully standalone execution of TFLM workloads on RISC-V 32-bit systems without OS dependencies, improving portability, testability and embedded readiness within the QEMU "virt" environment.

PR #3254, opened 2025-11-17
Feature: Complete RISC-V baremetal support with automated qemu emulation

sshahi-mips:
**Solves issue #3253**

Adds full bare-metal RISC-V 32-bit support for TensorFlow Lite Micro, including a minimal bootloader, linker script, UART driver logging, stubs (all under tensorflow/lite/micro/riscv32_baremetal/), and a new target (riscv32_baremetal_makefile.inc) with QEMU full-system emulation. This enables running TFLM examples without any OS dependencies and improves embedded portability and testability.

### DIFF
--- a/tensorflow/lite/micro/testing/test_with_qemu.sh
+++ b/tensorflow/lite/micro/testing/test_with_qemu.sh
@@ -30,7 +30,13 @@ MICRO_LOG_PATH=${TEST_TMPDIR}/${3}
 MICRO_LOG_FILENAME=${MICRO_LOG_PATH}/logs.txt
 
 mkdir -p ${MICRO_LOG_PATH}
-qemu-${1} -cpu ${2} ${3} 2>&1 | tee ${MICRO_LOG_FILENAME}
+if [[ "${5}" == "riscv32_baremetal" ]]; then
+    # unified call
+    qemu-${1} -nographic -M virt -bios none -kernel ${3} 2>&1 | tee ${MICRO_LOG_FILENAME}
+else
+    qemu-${1} -cpu ${2} ${3} 2>&1 | tee ${MICRO_LOG_FILENAME}
+fi
+
 if [[ ${4} != "non_test_binary" ]]
 then
   if grep -q "${4}" ${MICRO_LOG_FILENAME}


tensorflow/lite/micro/riscv32_baremetal/README.md
```diff
@@ -0,0 +1,38 @@
+### **RISC-V 32-Bit Bare-Metal Support**
```
veblush, line 1: The current setup appears to be highly specific to QEMU. If this is the case, perhaps we should rename `riscv32_baremetal` to `riscv32_qemu` for better clarity.

tensorflow/lite/micro/riscv32_baremetal/debug_log.cc
```diff
@@ -0,0 +1,72 @@
+/* Copyright 2023 The TensorFlow Authors. All Rights Reserved.
```
veblush, line 1: 2025?

tensorflow/lite/micro/riscv32_baremetal/debug_log.cc
```diff
@@ -0,0 +1,72 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "tensorflow/lite/micro/debug_log.h"
+#include <stdarg.h> 
+#include <stdio.h>  
+#include "tensorflow/lite/micro/riscv32_baremetal/lightweight_snprintf.cc"
+
+// UART definitions- Datasheet: https://pdos.csail.mit.edu/6.S081/2024/lec/16550.pdf
```
veblush, line 21: How about using Semihosting instead of having UART support here?

tensorflow/lite/micro/riscv32_baremetal/stubs.cc
```diff
@@ -0,0 +1,40 @@
+extern "C" {
```
veblush, line 1: This might not be necessary if Semihosting is used?

tensorflow/lite/micro/tools/make/Makefile
```diff
@@ -881,6 +881,36 @@ $(BINDIR)%_test : $(CORE_OBJDIR)%_test.o $(MICROLITE_LIB_PATH)
 	-o $@ $< \
 	$(MICROLITE_LIB_PATH) $(LDFLAGS) $(MICROLITE_LIBS)
 
+# Build rules and object setup for the RISC-V 32-bit bare-metal target,
```
veblush, line 884: The primary Makefile should only contain common definitions. All platform-specific configurations must be moved into their respective target makefiles.

tensorflow/lite/micro/tools/make/targets/riscv32_baremetal_makefile.inc
```diff
@@ -0,0 +1,31 @@
...
+MICROLITE_LIBS = -lc -lm -lgcc 
+
+TEST_SCRIPT := $(TENSORFLOW_ROOT)tensorflow/lite/micro/testing/test_with_qemu.sh
+
+# List of available examples
+EXAMPLES := person_detection hello_world micro_speech person_detection_test hello_world_test micro_speech_test dtln_test
+
+# Run a given example in QEMU using test_with_qemu.sh
+$(EXAMPLES):
+	@$(TEST_SCRIPT) system-riscv32 - gen/riscv32_baremetal_x86_64_default_gcc/bin/$@ non_test_binary riscv32_baremetal
```
veblush, line 29: This doesn't seem like the best place for the examples instructions. Why do we need this?

tensorflow/lite/micro/tools/make/targets/riscv32_baremetal_makefile.inc
```diff
@@ -0,0 +1,31 @@
+# Toolchain Configuration
+TARGET_ARCH := riscv32
+TARGET_TOOLCHAIN_PREFIX := riscv32-unknown-elf-
+
+# Toolchain Path 
+TARGET_TOOLCHAIN_ROOT := /home/mips/riscv32/bin/
```
veblush, line 6: I think this is the test code?

veblush (commented):
Thank you for the PR!

tensorflow/lite/micro/riscv32_baremetal/debug_log.cc
```diff
@@ -0,0 +1,72 @@
...
+
+// DebugLog()
+// TFLM's platform hook for debug logging.
+
+extern "C" void DebugLog(const char* format, va_list args) {
+#ifndef TF_LITE_STRIP_ERROR_STRINGS
+  char buffer[256];  // Temporary buffer for formatted string
+  int len = mini_vsnprintf(buffer, sizeof(buffer), format, args);
+  if (len > 0) {
+    UART0_FCR = UARTFCR_FFENA;
```
ddavis-2015, line 57: Should be part of uart_puts method.

tensorflow/lite/micro/riscv32_baremetal/debug_log.cc
ddavis-2015: @veblush I dont think we should have any specific UART support here, since this PR is for generic riscv32.  Any specific UART support should be part of the application, not TFLM.  This file should look identical to `tensorflow/lite/micro/riscv32_generic/debug_log.cc` with the `std::fputs` replaced with an external `uart_puts` which is supplied by the application.  Also there should be documentation explaining the application requirements for using riscv32-baremetal (stub methods, uart, etc.)

tensorflow/lite/micro/riscv32_baremetal/debug_log.cc
```diff
@@ -0,0 +1,72 @@
...
+  }
+}
+
+// DebugLog()
+// TFLM's platform hook for debug logging.
+
+extern "C" void DebugLog(const char* format, va_list args) {
+#ifndef TF_LITE_STRIP_ERROR_STRINGS
+  char buffer[256];  // Temporary buffer for formatted string
+  int len = mini_vsnprintf(buffer, sizeof(buffer), format, args);
```
ddavis-2015, line 55: Why are we using this mini_vsnprintf?  The `eyalroz_printf` code should be used here (as per `tensorflow/lite/micro/riscv32_generic/debug_log.cc`).  If the application developer wants a different printf library, they can define their own `DebugLog` and `DebugVsnprintf`.  Slapping a `weak` attribute on these two methods in `debug_log.cc` for bare-metal should be sufficient for that to happen.

tensorflow/lite/micro/riscv32_baremetal/lightweight_snprintf.cc
ddavis-2015, line 1: @veblush I'm not seeing why this should be part of TFLM.

tensorflow/lite/micro/riscv32_baremetal/stubs.cc
ddavis-2015, line 1: @veblush Don't see how any of the stubs should be part of TFLM (except `_exit`).  No example application or unit test should require any of these stubs, nor should they currently be calling any C or C++ standard library code that would use these stubs.  The exception I think would be `_exit` as a return from `main` may trigger `_exit` as part of the C/C++ runtime.

tensorflow/lite/micro/testing/test_with_qemu.sh
```diff
@@ -30,7 +30,13 @@ MICRO_LOG_PATH=${TEST_TMPDIR}/${3}
 MICRO_LOG_FILENAME=${MICRO_LOG_PATH}/logs.txt
 
 mkdir -p ${MICRO_LOG_PATH}
-qemu-${1} -cpu ${2} ${3} 2>&1 | tee ${MICRO_LOG_FILENAME}
+if [[ "${5}" == "riscv32_baremetal" ]]; then
+    # unified call
+    qemu-${1} -nographic -M virt -bios none -kernel ${3} 2>&1 | tee ${MICRO_LOG_FILENAME}
```
ddavis-2015, line 35: @veblush If something beyond `apt install qemu-system` is involved in running this command or some other Linux OS configuration is required, this should be documented.

tensorflow/lite/micro/tools/make/targets/riscv32_baremetal_makefile.inc
ddavis-2015: @veblush The toolchain download should occur here as per `tensorflow/lite/micro/tools/make/targets/riscv32_generic_makefile.inc`

Issue #3256, opened 2025-11-25
feat(compression): update tooling to use DECODE operators


rkuester:
## Summary

Update the compression tooling to produce models that use DECODE
operators instead of the metadata-based compression scheme.

The interpreter already supports DECODE operators for decompression.
The previous tooling stored compression metadata in the flatbuffer,
requiring the interpreter to detect compressed tensors and requiring
each kernel to implement decompression internally. The updated tooling
inserts explicit DECODE custom operators into the graph, handling
decompression before kernels execute and eliminating per-kernel
decompression logic.

Additionally, refactor the tooling to use a plugin architecture,
enabling multiple compression methods (LUT, Huffman, Pruning) to be
implemented independently.

## Changes

- Replace `model_facade` with `model_editor` for TFLite model manipulation
- Add DECODE operator insertion logic for compressed tensors
- Add `Compressor` protocol for compression plugins
- Implement LUT compression as a plugin (Huffman and Pruning are stubs)
- Add integration tests verifying compressed models produce correct
  inference results through the TFLM interpreter

## Testing

All changes include unit tests. Integration tests run with
`--//:with_compression` and verify:
- Compressed model outputs match uncompressed
- DECODE operators are inserted
- Compressed models are smaller than originals

PR #3257, opened 2025-11-25
feat(compression): update tooling to use DECODE operators

rkuester:
This is a draft PR for running CI tests on the full change. The commits
along this branch will be individually submitted for review.

See the linked issue for a description of the change.

BUG=implements #3256


tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,227 @@
...
+        reverse=True,
+    )
+
+    for consumer, info in consumer_pairs:
+      # Create ancillary data tensor (one per DECODE)
+      ancillary_tensor = _create_ancillary_tensor(
+          info.ancillary_data,
+          info.tensor,
+      )
+      subgraph.tensors.append(ancillary_tensor)
```
ddavis-2015, line 208: This appears to create too many ancillary tensor copies.  Each encoded tensor can reuse the associated ancillary tensor.  Should _create_ancillary_tensor have a cache keyed on the original tensor?

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,227 @@
...
+          inputs=[info.tensor, ancillary_tensor],
+          outputs=[output_tensor],
+      )
+
+      # Insert DECODE immediately before this consumer
+      insert_pos = subgraph.operators.index(consumer)
+      subgraph.operators.insert(insert_pos, decode_op)
+
+      # Rewire only this consumer to use the decoded output
+      _rewire_consumers([consumer], info.tensor, output_tensor)
```
ddavis-2015, line 227: The code to comply with "Input tensor pairs" and "Tensor rewriting" in the design doc seem to be missing.  Should there be another loop after the loop with _rewire_consumers that iterates compression_results and rewrites the original tensor (removing quantization, setting type to UINT8, changing shape to single rank and equal to the number of bytes in the encoded tensor)?

tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -0,0 +1,572 @@
...
+          # Quantization parameters present
+          scales = list(fb_quant.scale)
+          zeros = list(
+              fb_quant.zeroPoint
+          ) if fb_quant.zeroPoint is not None else [0] * len(scales)
+
+          # Handle axis: only set if per-channel (more than one scale)
+          axis = None
+          if len(scales) > 1 and fb_quant.quantizedDimension is not None:
+            axis = fb_quant.quantizedDimension
```
ddavis-2015, line 342: The axis should always be copied if it exists.  It is perfectly valid for len(scales) to be 1, the quantized_dimension to be 3, and shape[3] is equal to 1.

tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -0,0 +1,572 @@
...
+      # Read quantization parameters if present
+      quant = None
+      if fb_tensor.quantization:
+        fb_quant = fb_tensor.quantization
+        if fb_quant.scale is not None and len(fb_quant.scale) > 0:
+          # Quantization parameters present
+          scales = list(fb_quant.scale)
+          zeros = list(
+              fb_quant.zeroPoint
+          ) if fb_quant.zeroPoint is not None else [0] * len(scales)
```
ddavis-2015, line 337: The zero_point vector should never be expanded, only copied (this is a converter optimization).  It is valid to generate quantization data where the length of the scale vector is > 1 and the zero_point vector has length 1 (this is a converter optimization). This case is already handled in TFLM.

tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -0,0 +1,572 @@
...
+
+
+def _dtype_to_numpy(dtype: tflite.TensorType) -> np.dtype:
+  """Convert TFLite dtype to numpy dtype."""
+  type_map = {
+      tflite.TensorType.INT8: np.int8,
+      tflite.TensorType.INT16: np.int16,
+      tflite.TensorType.INT32: np.int32,
+      tflite.TensorType.UINT8: np.uint8,
+      tflite.TensorType.FLOAT32: np.float32,
```
ddavis-2015, line 402: missing INT64

python/tflite_micro/_runtime.cc
ddavis-2015, line 1: It occurs to me there is no reason to include alternate decompression memory APIs for the Python MicroInterpreter.  Not sure why I didn't bring this up before...

python/tflite_micro/interpreter_wrapper.cc
ddavis-2015, line 1: It occurs to me there is no reason to include alternate decompression memory APIs for the Python MicroInterpreter.  Not sure why I didn't bring this up before...

python/tflite_micro/interpreter_wrapper.h
ddavis-2015, line 1: It occurs to me there is no reason to include alternate decompression memory APIs for the Python MicroInterpreter.  Not sure why I didn't bring this up before...

python/tflite_micro/runtime.py
ddavis-2015, line 1: It occurs to me there is no reason to include alternate decompression memory APIs for the Python MicroInterpreter.  Not sure why I didn't bring this up before...

tensorflow/lite/micro/compression/compress.py
```diff
@@ -293,46 +121,50 @@ def _apply_flatbuffer_alignment(model_bytes: bytearray) -> bytearray:
...
+      if len(tensor_spec.compression) != 1:
+        raise compressor.CompressionError(
+            "Each tensor must have exactly one compression method")
+
+      method = tensor_spec.compression[0]
+      plugin = _get_compressor(method)
+      result = plugin.compress(tensor, method)
+
+      # Replace tensor data with encoded data
+      tensor.buffer.data = result.encoded_data
```
ddavis-2015, line 152: Could also occur later, after the DECODE insertion, when the rest of the tensor needs to be updated (removing quantization, changing shape and type, etc.)

tensorflow/lite/micro/compression/pruning.py
```diff
@@ -0,0 +1,59 @@
...
+Supported tensor types (when implemented): All TFLM tensor types
+"""
+
+from tflite_micro.tensorflow.lite.micro.compression import compressor
+from tflite_micro.tensorflow.lite.micro.compression import decode
+from tflite_micro.tensorflow.lite.micro.compression import model_editor
+from tflite_micro.tensorflow.lite.micro.compression import spec
+
+
+class PruningCompressor:
```
ddavis-2015, line 28: This does not inherit from Compressor?  Am I missing something here?

tensorflow/lite/micro/compression/lut.py
```diff
@@ -0,0 +1,317 @@
...
+      print(
+          f"warning: index_bitwidth larger than necessary: only "
+          f"{actual_bitwidth} bits needed, but {spec_bitwidth} specified",
+          file=sys.stderr)
+
+    # Pack indices into bytes
+    encoded_data = pack_indices(compressed.indices, spec_bitwidth)
+
+    # Pack value tables
+    table_len = 2**spec_bitwidth
```
ddavis-2015, line 295: There is an opportunity here to save space, if the table_len is the max of the length of the value table slices.  I know this would not apply to the clustering method, but there may be some model somewhere that is created where they only use 9 of 16 entries in the value table for bitwidth 4 (as an example).  Maybe a `TODO(rkuester):` comment here for the future?

tensorflow/lite/micro/compression/compress.py
```diff
@@ -293,46 +121,50 @@ def _apply_flatbuffer_alignment(model_bytes: bytearray) -> bytearray:
...
+          tensor_spec.tensor]
+
+      # Currently only one compression method per tensor
+      if len(tensor_spec.compression) != 1:
+        raise compressor.CompressionError(
+            "Each tensor must have exactly one compression method")
+
+      method = tensor_spec.compression[0]
+      plugin = _get_compressor(method)
+      result = plugin.compress(tensor, method)
```
ddavis-2015, line 149: Warn user if this resulted in expansion (sizeof encoded_data + sizeof ancillary_data > original tensor length)?

tensorflow/lite/micro/compression/lut.py
```diff
@@ -0,0 +1,317 @@
...
+      # Use one value table for the entire tensor
+      return None
+
+    if q.axis is not None and q.axis < len(tensor.shape):
+      if quantization_channels == tensor.shape[q.axis]:
+        return q.axis
+
+  raise compressor.CompressionError(
+      "Invalid or no quantization parameters from which to "
+      "infer the axis along which tensor should be compressed.")
```
ddavis-2015, line 172: This should be `return None`.  The TFLM code already assumes number of channels to be 1 if there are no quantization parameters.

tensorflow/lite/micro/compression/lut.py
```diff
@@ -0,0 +1,317 @@
...
+  """
+  buffer = bytearray()
+  for t in tables:
+    padding_needed = table_len - len(t)
+    padded = np.pad(t, (0, padding_needed), mode='constant', constant_values=0)
+    buffer.extend(padded.tobytes())
+  return bytes(buffer)
+
+
+class LutCompressor:
```
ddavis-2015, line 243: This does not inherit from Compressor? Am I missing something here?

tensorflow/lite/micro/compression/compress_test.py
```diff
@@ -358,58 +152,199 @@ def test_multiple_tables_with_padding(self):
...
+            ],
+        )
+    ])
+    fb = model.build()
+
+    specs = [
+        spec.Tensor(
+            subgraph=0,
+            tensor=0,
+            compression=[spec.LookUpTableCompression(index_bitwidth=4)])
```
ddavis-2015, line 295: index_bitwidth=2?

tensorflow/lite/micro/compression/compress_test.py
```diff
@@ -358,58 +152,199 @@ def test_multiple_tables_with_padding(self):
...
```
ddavis-2015, line 234: index_bitwidth=2?

tensorflow/lite/micro/compression/compression_integration_test.py
```diff
@@ -0,0 +1,397 @@
...
+
+  @unittest.skip("Pruning compression not yet implemented")
+  def test_pruning_compressed_model_is_smaller(self):
+    """Pruning-compressed model is smaller than original."""
+    pass
+
+
+if __name__ == "__main__":
+  os.environ["TF_CPP_MIN_LOG_LEVEL"] = "2"
+  os.environ["TF_ENABLE_ONEDNN_OPTS"] = "0"
```
ddavis-2015, line 396: Could you add a comment describing these environment vars?

tensorflow/lite/micro/compression/compression_integration_test.py
```diff
@@ -0,0 +1,397 @@
...
+  input_t = model_editor.Tensor(
+      shape=(1, cols),
+      dtype=tflite.TensorType.INT8,
+      name="input",
+  )
+  output_t = model_editor.Tensor(
+      shape=(1, rows),
+      dtype=tflite.TensorType.INT8,
+      name="output",
+  )
```
ddavis-2015, line 69: Assign tensor numbers as comments

tensorflow/lite/micro/compression/compression_integration_test.py
```diff
@@ -0,0 +1,397 @@
...
+      dtype=tflite.TensorType.INT8,
+      name="intermediate",
+      quantization=quant,
+  )
+  output2 = model_editor.Tensor(
+      shape=(1, 4),
+      dtype=tflite.TensorType.INT8,
+      name="output2",
+      quantization=quant,
+  )
```
ddavis-2015, line 253: Can you add comments assigning tensor numbers to each Tensor?

tensorflow/lite/micro/compression/compression_integration_test.py
```diff
@@ -0,0 +1,397 @@
...
+    sg = model.subgraphs[0]
+
+    # Find DECODE operators
+    decode_ops = [
+        op for op in sg.operators if op.opcode == tflite.BuiltinOperator.CUSTOM
+        and op.custom_code == decode_insert.DECODE_CUSTOM_OP_NAME
+    ]
+
+    self.assertGreater(len(decode_ops), 0,
+                       "DECODE operators should be present")
```
ddavis-2015, line 149: Should be exactly 1, as per the specs created above

tensorflow/lite/micro/compression/compressor.py
```diff
@@ -0,0 +1,80 @@
...
+  """Protocol that compression plugins must implement.
+
+  Each compression method (LUT, Huffman, Pruning) provides a class implementing
+  this protocol. The compress() function uses duck typing to call the plugin.
+  """
+
+  @property
+  def decode_type(self) -> decode.DecodeType:
+    """The DecodeType constant for this compression method."""
+    ...
```
ddavis-2015, line 59: syntactically legal, but shouldn't it just be `pass`?

tensorflow/lite/micro/compression/compressor.py
```diff
@@ -0,0 +1,80 @@
...
+      method: The compression method spec (e.g., LookUpTableCompression).
+
+    Returns:
+      CompressionResult with encoded tensor data and ancillary data bytes.
+
+    Raises:
+      CompressionError: If compression fails (e.g., too many unique values
+                        for specified bitwidth, missing quantization, etc.).
+    """
+    ...
```
ddavis-2015, line 80: syntactically legal, but shouldn't it just be `pass`?

tensorflow/lite/micro/compression/decode.py
```diff
@@ -0,0 +1,240 @@
...
+    # bytes 2-3 remain zero (reserved)
+    result[4:16] = user_data
+    return bytes(result)
+
+
+class AncillaryDataSerializer(Protocol):
+  """Protocol for objects that can serialize ancillary data."""
+
+  def to_bytes(self) -> bytes:
+    ...
```
ddavis-2015, line 197: syntactically legal, but shouldn't it just be `pass`?

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,227 @@
...
+      output_tensor = _create_output_tensor(info.tensor)
+      subgraph.tensors.append(output_tensor)
+
+      # Create DECODE operator
+      decode_op = model_editor.Operator(
+          opcode=tflite.BuiltinOperator.CUSTOM,
+          custom_code=DECODE_CUSTOM_OP_NAME,
+          inputs=[info.tensor, ancillary_tensor],
+          outputs=[output_tensor],
+      )
```
ddavis-2015, line 220: This being located here does not allow for a single DECODE operator to have multiple encoded inputs and ancillary tensors.  The example would be CONCATENATION which takes multiple inputs, where several might be encoded tensors.

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,227 @@
...
+  by_subgraph: dict[int, list[_CompressedTensorInfo]] = defaultdict(list)
+
+  for (sg_idx, tensor_idx), result in compression_results.items():
+    subgraph = model.subgraphs[sg_idx]
+    tensor = subgraph.tensors[tensor_idx]
+    consumers = _find_tensor_consumers(subgraph, tensor)
+
+    if not consumers:
+      # Tensor not used as input anywhere---no DECODE needed
+      continue
```
ddavis-2015, line 175: This should trigger a warning to the user.  A valid model can be constructed with multiple subgraphs, and which uses the IF, WHILE, or STABLE_HLO_COMPOSITE operators.  A subgraph other than 0 could contain an encoded tensor, which also appears as one of the outputs of the same subgraph.  Since the encoded tensor did not appear as the input for any operator in that subgraph, this code would be executed.  This should be a fatal error, as regards the tensor specs originally passed to the compression script.  This would require the user to change the tensor specs to include the corresponding output tensor in subgraph 0 as the one to compress.

Or (and I think this is what should happen), if there are no consumers here, then check the current subgraph outputs to see if any match the tensor.  If there is a match, add it to a list of encoded tensors. The list of encoded tensors is then used to generate a single DECODE operator which is placed at the end of the subgraph's operators, and the current subgraph outputs are rewritten.
 
In any case, the user should be notified if they have a tensor in the specs, that is never used in the model.

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,393 @@
...
+  output1 = model_editor.Tensor(
+      shape=(1, 4),
+      dtype=tflite.TensorType.INT8,
+      name="output1",
+  )
+  output2 = model_editor.Tensor(
+      shape=(1, 4),
+      dtype=tflite.TensorType.INT8,
+      name="output2",
+  )
```
ddavis-2015, line 94: Comments to label what number each tensor is.

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,393 @@
...
+  input_t = model_editor.Tensor(
+      shape=(1, 4),
+      dtype=tflite.TensorType.INT8,
+      name="input",
+  )
+  output_t = model_editor.Tensor(
+      shape=(1, 4),
+      dtype=tflite.TensorType.INT8,
+      name="output",
+  )
```
ddavis-2015, line 49: Comments to label what number each tensor is (or make it part of the tensor name, like T#5 perhaps?)

tensorflow/lite/micro/compression/huffman.py
```diff
@@ -0,0 +1,60 @@
...
+Supported tensor types (when implemented): INT8, INT16
+"""
+
+from tflite_micro.tensorflow.lite.micro.compression import compressor
+from tflite_micro.tensorflow.lite.micro.compression import decode
+from tflite_micro.tensorflow.lite.micro.compression import model_editor
+from tflite_micro.tensorflow.lite.micro.compression import spec
+
+
+class HuffmanCompressor:
```
ddavis-2015, line 28: Not a subclass of Compressor?  Am I missing something here?

tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -0,0 +1,572 @@
...
+    - buffer=: Explicitly provide a Buffer object (can be shared between tensors)
+    - data=: Convenience parameter that auto-creates a Buffer
+
+    Cannot specify both buffer and data at initialization.
+    """
+  shape: tuple
+  dtype: tflite.TensorType
+  buffer: Optional[Buffer] = None
+  quantization: Optional[Quantization] = None
+  name: Optional[str] = None
```
ddavis-2015, line 104: `is_variable` must be maintained

tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -0,0 +1,572 @@
...
+        """
+    return _dtype_to_numpy(self.dtype)
+
+
+@dataclass
+class OperatorCode:
+  """Operator code specification."""
+  builtin_code: tflite.BuiltinOperator
+  custom_code: Optional[str] = None
+  version: int = 1
```
ddavis-2015, line 207: must maintain `deprecated_builtin_code`, and it must interoperate as per the schema.fbs comments with `builtin_code`.

tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -0,0 +1,572 @@
...
+@dataclass
+class Operator:
+  """Declarative operator specification."""
+  opcode: Union[tflite.BuiltinOperator, int]
+  inputs: List[Tensor]
+  outputs: List[Tensor]
+  custom_code: Optional[str] = None
+
+  # Set when reading from existing model
+  opcode_index: Optional[int] = None
```
ddavis-2015, line 219: Must maintain the following fields:
```
  builtin_options:BuiltinOptions;
  custom_options:[ubyte];
  custom_options_format:CustomOptionsFormat;

  // A list of indices to the subgraph's "tensors" that are internal to an Op.
  // Internal tensors are those that do not flow in or out of the operation,
  // but instead are part of internal computation. As such, the operation's
  // implementation may manage its memory more efficiently. They are needed
  // however (i.e. not just an implementation detail) since they are part of the
  // computation, which may require relevant metadata such as quantization
  // parameters.
  intermediates:[int];

  // Flatbuffers union struct has a 128 elements limit in JAVA, so a second
  // union is added, in the case of where BuitlinOptions2 runs out, a third
  // one can be added
  builtin_options_2 : BuiltinOptions2;

  // Index into operators_debug_metadata list.
  debug_metadata_index: int = -1;
```

`builtin_options` and `builtin_options_2` must interoperate correctly.

tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -0,0 +1,572 @@
...
+
+
+@dataclass
+class Subgraph:
+  """Declarative subgraph specification with imperative methods."""
+  tensors: List[Tensor] = field(default_factory=list)
+  operators: List[Operator] = field(default_factory=list)
+  inputs: List[Tensor] = field(default_factory=list)
+  outputs: List[Tensor] = field(default_factory=list)
+  name: Optional[str] = None
```
ddavis-2015, line 231: `debug_metadata_index` should be maintained.

tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -0,0 +1,572 @@
...
+
+@dataclass
+class Model:
+  """Top-level model specification."""
+  subgraphs: List[Subgraph] = field(default_factory=list)
+  buffers: _BufferList = field(
+      default_factory=_BufferList)  # Auto-sets buffer.index on append
+  operator_codes: List[OperatorCode] = field(default_factory=list)
+  metadata: dict = field(default_factory=dict)
+  description: Optional[str] = None
```
ddavis-2015, line 267: must maintain the following fields:
```
version
signature_defs
```

tensorflow/lite/micro/compression/decode_insert.py
rkuester: I've created PR #3400 to address the review comments on this PR. The new PR incorporates the feedback and restructures the code. Below I'll reply to each comment explaining how it's been addressed.

---

Addressed. Ancillary tensors are now cached so that all DECODEs of the same compressed tensor share a single ancillary tensor. See [`ancillary_cache`](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-07340f9b1e2dd8d5fd5f026e5c8074b7e5168f1a415c0b6e7046001ad9325012R225-R245).

Each DECODE still gets its own output tensor, but the ancillary tensor is created only once per compressed tensor.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Addressed. The new PR adds [`_rewrite_encoded_tensor()`](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-07340f9b1e2dd8d5fd5f026e5c8074b7e5168f1a415c0b6e7046001ad9325012R124-R141) which rewrites the compressed tensor after DECODE insertion.

This removes quantization, sets the type to UINT8, and reshapes to a 1D byte array matching the encoded data length. Rewriting is deferred until after all DECODE output tensors are created. Since the same compressed tensor may have multiple consumers, we don't want to modify its metadata until all output tensors have been created from it.

tensorflow/lite/micro/compression/model_editor.py
rkuester: Addressed. The [quantization parsing logic](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-42847d6ce72f11e805ba31973cf09ae235e36fd7c68ded63d7afc5343801edc9R562-R568) now preserves the axis whenever it's non-zero, regardless of scale count. Only the case of axis==0 with a single scale is treated as per-tensor quantization (axis=None). This preserves per-channel quantization with a single channel at a non-zero axis.

tensorflow/lite/micro/compression/model_editor.py
rkuester: Addressed. Zero_points are now [copied as-is without expansion](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-42847d6ce72f11e805ba31973cf09ae235e36fd7c68ded63d7afc5343801edc9R559-R561). A single zero_point with multiple scales is preserved.

tensorflow/lite/micro/compression/model_editor.py
rkuester: Addressed. INT64 is now included in the [dtype mapping](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-42847d6ce72f11e805ba31973cf09ae235e36fd7c68ded63d7afc5343801edc9R629), with a [dedicated test](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-6aab2520fc52bb39d7976aed22d7094e16cf319d710c4c874a3c67ee975e33afR998).

python/tflite_micro/_runtime.cc
rkuester: Kept for now. The alt memory API is used in integration tests to exercise that code path, which caught a bug earlier in development where sharing DECODE outputs broke in the presence of alt memory.

python/tflite_micro/interpreter_wrapper.cc
rkuester: Kept for now. The alt memory API is used in integration tests to exercise that code path, which caught a bug earlier in development where sharing DECODE outputs broke in the presence of alt memory.

python/tflite_micro/interpreter_wrapper.h
rkuester: Kept for now. The alt memory API is used in integration tests to exercise that code path, which caught a bug earlier in development where sharing DECODE outputs broke in the presence of alt memory.

python/tflite_micro/runtime.py
rkuester: Kept for now. The alt memory API is used in integration tests to exercise that code path, which caught a bug earlier in development where sharing DECODE outputs broke in the presence of alt memory.

tensorflow/lite/micro/compression/compress.py
rkuester: Addressed. See response to comment 2 above: tensor rewriting now happens in [`_rewrite_encoded_tensor()`](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-07340f9b1e2dd8d5fd5f026e5c8074b7e5168f1a415c0b6e7046001ad9325012R124-R141) after DECODE insertion.

tensorflow/lite/micro/compression/pruning.py
rkuester: Addressed. Compressor classes now [explicitly inherit from the Protocol](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-966f3c40707584ee16088e431e73550bf56d397f732a0949673fb960b78d7e7eR28). Python Protocols use duck typing, so inheritance isn't strictly required, but explicit inheritance enables static type checking at definition time and makes the interface self-documenting.

tensorflow/lite/micro/compression/lut.py
rkuester: See response above.

tensorflow/lite/micro/compression/huffman.py
rkuester: See response above.

tensorflow/lite/micro/compression/lut.py
rkuester: Addressed. The [value table packing](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-c5cdc6468968dc366f457e1ba0425fa8b70128927a566eee63e595f109ffc651R296) now uses `max(len(t) for t in tables)` as the stride, so tables smaller than `2**bitwidth` save space.

tensorflow/lite/micro/compression/compress.py
rkuester: Addressed. The code now [warns if compression results in expansion](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-1ca95116d18953c4b50538524c131a4443fe18af60a5d351288f567710dea6d1R155-R161).

tensorflow/lite/micro/compression/lut.py
rkuester: Addressed. The [identify_compression_axis](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-c5cdc6468968dc366f457e1ba0425fa8b70128927a566eee63e595f109ffc651R156-R158) function now returns `None` for unquantized tensors, indicating per-tensor compression.

tensorflow/lite/micro/compression/compress_test.py
rkuester: The tests now include both scenarios. Some tests use minimal bitwidth (like [`index_bitwidth=2` at lines 142-155](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-defc9e243c4ac63bb60953f79e2a1f6e6bce73d158f9e3f085da8f9dc0568ae3R142-R155)), while others intentionally use larger-than-necessary bitwidth to verify the warning behavior.

tensorflow/lite/micro/compression/compress_test.py
rkuester: See response above.

tensorflow/lite/micro/compression/compression_integration_test.py
rkuester: Addressed. The [environment variables are now documented](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-e72dddf8cd66c93e81f64bff38fe7c9f4a5e0ada35a01ffb7a3204c42009ab3dR502-R505).

tensorflow/lite/micro/compression/compression_integration_test.py
rkuester: Rather than adding magic number comments, we added [`tensor_by_name()`](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-42847d6ce72f11e805ba31973cf09ae235e36fd7c68ded63d7afc5343801edc9R432) so tests can look up tensors by name instead of index.

tensorflow/lite/micro/compression/compression_integration_test.py
rkuester: Rather than adding magic number comments, we added [`tensor_by_name()`](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-42847d6ce72f11e805ba31973cf09ae235e36fd7c68ded63d7afc5343801edc9R432) so tests can look up tensors by name instead of index.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Rather than adding magic number comments, we added [`tensor_by_name()`](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-42847d6ce72f11e805ba31973cf09ae235e36fd7c68ded63d7afc5343801edc9R432) so tests can look up tensors by name instead of index.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Rather than adding magic number comments, we added [`tensor_by_name()`](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-42847d6ce72f11e805ba31973cf09ae235e36fd7c68ded63d7afc5343801edc9R432) so tests can look up tensors by name instead of index.

tensorflow/lite/micro/compression/compression_integration_test.py
rkuester: Addressed. The test now [asserts exactly 1 DECODE operator](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-e72dddf8cd66c93e81f64bff38fe7c9f4a5e0ada35a01ffb7a3204c42009ab3dR169).

tensorflow/lite/micro/compression/compressor.py
rkuester: Kept as `...`. The ellipsis is the conventional way to indicate abstract methods in Protocol classes per PEP 544.

tensorflow/lite/micro/compression/compressor.py
rkuester: Kept as `...`. The ellipsis is the conventional way to indicate abstract methods in Protocol classes per PEP 544.

tensorflow/lite/micro/compression/decode.py
rkuester: Kept as `...`. The ellipsis is the conventional way to indicate abstract methods in Protocol classes per PEP 544.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: The compressor design currently uses one DECODE operator per compressed tensor. Looking at the C++ kernel, I see it already supports multiple input/output pairs. Were you expecting us to batch them into a single DECODE for cases like CONCATENATION? Would that run into the same alt decompression memory problems as reusing the output of a single DECODE? Does using multiple DECODEs also run into those issues?

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Addressed. The new PR checks if the tensor is a subgraph output (raises `NotImplementedError` explaining compressed subgraph outputs aren't yet supported) or warns if it has no consumers and isn't an output. See [decode_insert.py L180-197](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-07340f9b1e2dd8d5fd5f026e5c8074b7e5168f1a415c0b6e7046001ad9325012R180-R197).

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: See earlier reply about using `tensor_by_name()` to avoid magic tensor indices.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: See earlier reply about using `tensor_by_name()` to avoid magic tensor indices.

tensorflow/lite/micro/compression/model_editor.py
rkuester: We reworked model_editor to wrap the backing `*T` types (`TensorT`, `OperatorT`, `ModelT`, etc.) rather than using pure dataclasses. When compiling, we reuse the wrapped object and only update fields we explicitly manage. This preserves all other fields, including those added by future schema changes.

For Tensor, see [model_editor.py L769-770](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-42847d6ce72f11e805ba31973cf09ae235e36fd7c68ded63d7afc5343801edc9R769-R770). Verified by [test_tensor_is_variable_preserved](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-6aab2520fc52bb39d7976aed22d7094e16cf319d710c4c874a3c67ee975e33afR1096).

The same approach applies to similar comments below.

tensorflow/lite/micro/compression/model_editor.py
rkuester: See above about wrapping `*T` types. Verified by [test_operator_code_deprecated_builtin_code_preserved](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-6aab2520fc52bb39d7976aed22d7094e16cf319d710c4c874a3c67ee975e33afR1194).

tensorflow/lite/micro/compression/model_editor.py
rkuester: See above about wrapping `*T` types. Verified by [test_operator_builtin_options_preserved](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-6aab2520fc52bb39d7976aed22d7094e16cf319d710c4c874a3c67ee975e33afR1127), [test_operator_intermediates_preserved](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-6aab2520fc52bb39d7976aed22d7094e16cf319d710c4c874a3c67ee975e33afR1165), and [test_operator_builtin_options_2_preserved](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-6aab2520fc52bb39d7976aed22d7094e16cf319d710c4c874a3c67ee975e33afR1316).

tensorflow/lite/micro/compression/model_editor.py
rkuester: See above about wrapping `*T` types. Verified by [test_operator_debug_metadata_index_preserved](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-6aab2520fc52bb39d7976aed22d7094e16cf319d710c4c874a3c67ee975e33afR1180).

tensorflow/lite/micro/compression/model_editor.py
rkuester: See above about wrapping `*T` types. Verified by [test_model_version_preserved](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-6aab2520fc52bb39d7976aed22d7094e16cf319d710c4c874a3c67ee975e33afR1223) and [test_model_signature_defs_preserved](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-6aab2520fc52bb39d7976aed22d7094e16cf319d710c4c874a3c67ee975e33afR1237).

tensorflow/lite/micro/compression/compressor.py
ddavis-2015: TIL... 😄

tensorflow/lite/micro/compression/decode_insert.py
ddavis-2015: If a single operator (for example CONCATENATION) has multiple compressed tensors, then those tensors and their DCMs should be batched into a single DECODE operator.  The DECODE kernel already handles this (as you observed) and handles the alternate decompression memory correctly in this case.

Or to put it more succinctly: multiple encoded tensors for a single operator, **MUST** be passed as multiple inputs to a **single** DECODE operator.

rkuester:
Closing this draft. It was the original full-change preview for the
DECODE-operator tooling (#3256): replacing the metadata-based
compression scheme with explicit DECODE custom operators, plus the
plugin architecture for LUT/Huffman/Pruning compressors.

Reworked since then. The current full change lives in #3400, with
individual commits going up for review one at a time, starting with
#3439. #3400 supersedes this branch entirely. Closing it just to keep
the open-PR list focused on what's actually moving toward merge.

PR #3259, opened 2025-12-02

rkuester:
models with a clean API. This provides an easy, readable solution for
creating models in unit tests, unified with our need for editing models
in the compression code itself.

The module eliminates manual index tracking and buffer management
through automatic bookkeeping, supporting both declarative and
imperative construction styles.

Supports quantization parameters (per-tensor and per-channel), metadata
key-value pairs, and read-modify-write workflows.

BUG=part of #3256

### DIFF
--- a/tensorflow/lite/micro/compression/BUILD
+++ b/tensorflow/lite/micro/compression/BUILD
@@ -237,6 +237,28 @@ py_test(
     ],
 )
 
+py_library(
+    name = "model_editor",
+    srcs = ["model_editor.py"],
+    deps = [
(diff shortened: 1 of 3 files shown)


rkuester:
Closing this draft. It was an early standalone PR for the model_editor
module, the unified API for creating, reading, and modifying TFLite
models that the compression tooling builds on.

That commit has been reworked and is now up for review as #3439, and
also appears in the full-change preview #3400. Follow #3439 for
model_editor; #3256 tracks the overall plan.

PR #3265, opened 2025-12-09
Add RFFT kernel to MicroMutableOpResolver.

shlmregev:
This change includes the RFFT kernel header, making the RFFT operation specialized for different types (int16 etc.) available for use in TFLite Micro models via the MicroMutableOpResolver..

BUG=315941833


shlmregev:
> @shlmregev @veblush I am confused by this PR change. Could not the application developer have included `signal/micro/kernels/rfft.h` in their own code, and then called `AddRfft` with the registration of their choice?

The visibility for register_signal_ops is limited to ["//tensorflow/lite/micro"](https://github.com/tensorflow/tflite-micro/blob/9cbf6e040335234e319cb969fa494e1f06f294ac/signal/micro/kernels/BUILD#L36-L38
) and that's the only rule that exposes this header.
It look like including the header in `micro_mutable_op_resolver.h` is the preferred way of [doing this](https://github.com/tensorflow/tflite-micro/blob/9cbf6e040335234e319cb969fa494e1f06f294ac/tensorflow/lite/micro/micro_mutable_op_resolver.h#L26-L40).

tensorflow/lite/micro/micro_mutable_op_resolver.h
```diff
@@ -18,6 +18,7 @@ limitations under the License.
 #include <cstdio>
 #include <cstring>
 
+#include "signal/micro/kernels/rfft.h"
```
veblush, line 22: I think the standard approach here is to add the register functions to https://github.com/tensorflow/tflite-micro/blob/9cbf6e040335234e319cb969fa494e1f06f294ac/tensorflow/lite/micro/kernels/micro_ops.h#L145-L158 instead of including new headers.

shlmregev: The reason why I need to include this file, is so that I can do this:
`AddRfft(::tflite::tflm_signal::Register_RFFT_INT16());`

Should I add `::tflite::tflm_signal::Register_RFFT_INT16()` to `micro_ops.h`?

micro_ops.h doesn't define any other Register_ function for specialized types. And looking at [tensorflow/lite/micro/micro_mutable_op_resolver.h](https://github.com/tensorflow/tflite-micro/pull/3265/files/79ecdd819a577a9f9478821247e8fc05b53fd0ec#diff-f69c2b555d1156a054319ef3a6381c1f70667e14c3f86fee5e43cffb74e339a3), there are multiple kernel headers included there, presumably for the same purpose.

ddavis-2015: @shlmregev @veblush Ah, I understand now.  This is how CONV_2D, FULLY_CONNECTED, MUL, etc. expose their additional registrations as well. So `micro_mutable_op_resolver.h` is the right place to put the include.  I'm thinking that `signal/micro/kernels/irfft.h` should also be included here, as it also has multiple registrations.

shlmregev: @ddavis-2015 @veblush can you please review this change.

PR #3277, opened 2025-12-29
Increase number of axis that could be reduced.

mmatuszak85:
I'm working with a Conformer model that requires more 2 axis during reduction. Without that change Address and Memory sanitizers are reporting errors.


ddavis-2015:
@mmatuszak85 Thank you for your interest in TFLM and for contributing to its improvement!  You PR has been superseded by PR #3281, and will now be closed.

PR #3296, opened 2026-01-12
[Dep] Upgraded Flatbuffer to 25.9.23

veblush:
This update upgrades FlatBuffers to v25.9.23. Beyond the standard version bump, this includes a few major modifications:

- Refined `flatbuffers.patch` to avoid dynamic allocation.
- To comply with the new requirement to `Reset()` builders after serialization, each model now utilizes its own dedicated builder instance. This prevents shared-instance conflicts and unintended data removal.
- Memory Alignment: Integrated `alignas` specifiers across test datasets to meet strict platform-specific alignment standards.

BUG=n/a

### DIFF
--- a/tensorflow/lite/micro/memory_helpers_test.cc
+++ b/tensorflow/lite/micro/memory_helpers_test.cc
@@ -207,9 +207,9 @@ TF_LITE_MICRO_TEST(TestBytesRequiredForTensor) {
 
 TF_LITE_MICRO_TEST(TestAllocateOutputDimensionsFromInput) {
   constexpr int kDimsLen = 4;
-  alignas(TfLiteIntArray) int input1_dims[] = {1, 1};
-  alignas(TfLiteIntArray) int input2_dims[] = {kDimsLen, 5, 5, 5, 5};
-  alignas(TfLiteIntArray) int output_dims[] = {0, 0, 0, 0, 0};
+  int input1_dims[] = {1, 1};
+  int input2_dims[] = {kDimsLen, 5, 5, 5, 5};
+  int output_dims[] = {0, 0, 0, 0, 0};
   TfLiteTensor input_tensor1 = tflite::testing::CreateTensor<int32_t>(
       nullptr, tflite::testing::IntArrayFromInts(input1_dims));
   TfLiteTensor input_tensor2 = tflite::testing::CreateTensor<int32_t>(


tensorflow/lite/micro/kernels/if_test.cc
```diff
@@ -135,7 +135,7 @@ TF_LITE_MICRO_TEST(IfShouldInvokeSubgraphConditionTrue) {
 }
 
 TF_LITE_MICRO_TEST(IfShouldInvokeSubgraphConditionFalse) {
-  constexpr int kArenaSize = 5000;
+  constexpr int kArenaSize = 16384;
```
ddavis-2015: @veblush Thats a big increase in arena size.  Any ideas why this expansion happened?

tensorflow/lite/micro/kernels/if_test.cc
```diff
@@ -163,7 +163,7 @@ TF_LITE_MICRO_TEST(IfShouldInvokeSubgraphConditionFalse) {
 }
 
 TF_LITE_MICRO_TEST(IfShouldNotOverwriteTensorAcrossSubgraphs) {
-  constexpr int kArenaSize = 5000;
+  constexpr int kArenaSize = 16384;
```
ddavis-2015: @veblush Same as line 138 question

tensorflow/lite/micro/tools/make/flatbuffers.patch
```diff
@@ -79,10 +82,10 @@ index 8e8cac14..52dae316 100644
            return d;
 +#endif
          }
-         case FBT_VECTOR: return static_cast<double>(AsVector().size());
-         case FBT_BOOL:
```
ddavis-2015, line 83: FBT_BOOL is no more?

tensorflow/lite/micro/micro_allocation_info_test.cc
```diff
@@ -76,7 +76,7 @@ TF_LITE_MICRO_TEST(TestSingleSubgraphWithIntermediates) {
 }
 
 TF_LITE_MICRO_TEST(TestMultiSubgraphWithIf) {
-  constexpr int kArenaSize = 1024;
+  constexpr int kArenaSize = 2048;
```
ddavis-2015: @veblush Did the updated flatbuffers library cause the model schema to expand in  size?

tensorflow/lite/micro/test_helpers.cc
```diff
@@ -1939,10 +1936,9 @@ const Model* GetNoOpModelWithTensorShape(
 const Tensor* Create1dFlatbufferTensor(int size, bool is_variable) {
   using flatbuffers::Offset;
   flatbuffers::FlatBufferBuilder* builder = BuilderInstance();
-  constexpr size_t tensor_shape_size = 1;
-  const int32_t tensor_shape[tensor_shape_size] = {size};
+  std::vector<int32_t> shape = {size};
```
ddavis-2015: @veblush will the verboten std::vector port to non x86 platforms?

tensorflow/lite/micro/kernels/if_test.cc
veblush: Reverted.

tensorflow/lite/micro/micro_allocation_info_test.cc
veblush: Reverted.

tensorflow/lite/micro/test_helpers.cc
veblush: Reverted.

tensorflow/lite/micro/kernels/if_test.cc
veblush: Reverted.

tensorflow/lite/micro/tools/make/flatbuffers.patch
veblush: This isn't part of the changes?

tensorflow/lite/micro/compression/metadata_saved.h
ddavis-2015, line 1: I thought this file was generated?  Should ask @rkuester about it.

tensorflow/lite/micro/memory_helpers_test.cc
```diff
@@ -23,7 +24,8 @@ namespace {
 // This just needs to be big enough to handle the array of 5 ints allocated
 // in TestAllocateOutputDimensionsFromInput below.
 const int kGlobalPersistentBufferLength = 100;
-char global_persistent_buffer[kGlobalPersistentBufferLength];
+alignas(TfLiteIntArray) char global_persistent_buffer
```
ddavis-2015: @veblush this should be 
```
const int kGlobalPersistentBufferLength = 5;
int global_persistent_buffer[kGlobalPersistentBufferLength];
```

no alignment attributes

tensorflow/lite/micro/testing_helpers_test.cc
```diff
@@ -111,7 +113,7 @@ TF_LITE_MICRO_TEST(CreateSymmetricPerChannelQuantizedTensor) {
   TfLiteIntArray* dims = tflite::testing::IntArrayFromInts(dims_arr);
 
   int zero_points[channels + 1];
-  float scales[channels + 1];
+  alignas(tflite::MicroArenaBufferAlignment()) float scales[channels + 1];
```
ddavis-2015: no alignment attribute should be used

veblush: Done.

tensorflow/lite/micro/memory_helpers_test.cc
veblush: I think it still needs it because FakeAllocatePersistentBuffer is supposed to return aligned memory like others;
https://github.com/tensorflow/tflite-micro/blob/b46e68e79f267ee3af6ddb075df723540eddaae6/tensorflow/lite/micro/micro_allocator.cc#L690-L693

tensorflow/lite/micro/compression/metadata_saved.h
rkuester: It is. [BUILD](https://github.com/veblush/tflite-micro/blob/ad6ce83b3eac5a92c3e969313c3f06891277fade/tensorflow/lite/micro/compression/BUILD) contains instructions for updating. Hopefully these changes are generated too. There's a unit test that should fail if the generated and checked-in files aren't the same.

PR #3298, opened 2026-01-13
Added ResetVariableTensor

veblush:
Added a new  `ResetVariableTensor` method to `MicroInterpreterGraph`. This allows for the manual resetting of variable tensors (stateful tensors) to their initial values. This is particularly useful for complex models where only a partial set of tensors needs to be cleared.

BUG=n/a

### DIFF
--- a/tensorflow/lite/micro/micro_interpreter_graph_test.cc
+++ b/tensorflow/lite/micro/micro_interpreter_graph_test.cc
@@ -15,6 +15,7 @@ limitations under the License.
 
 #include <cstdint>
 
+#include "tensorflow/lite/micro/kernels/kernel_util.h"
 #include "tensorflow/lite/micro/memory_helpers.h"
 #include "tensorflow/lite/micro/micro_interpreter.h"
 #include "tensorflow/lite/micro/test_helpers.h"
@@ -44,13 +45,15 @@ TF_LITE_MICRO_TEST(TestResetVariableTensor) {
       interpreter.GetTensor(variable_tensor_idx);
   TF_LITE_MICRO_ASSERT(nullptr != variable_tensor);
 
-  if (variable_tensor->data.raw != nullptr) {
+  if (variable_tensor->data.data != nullptr) {
     // Fill the variable tensor with non-zero values.
     size_t buffer_size;
     TF_LITE_MICRO_ASSERT_EQ(kTfLiteOk, tflite::TfLiteEvalTensorByteLength(
                                            variable_tensor, &buffer_size));
+    uint8_t* variable_tensor_buffer =
+        tflite::micro::GetTensorData<uint8_t>(variable_tensor);
     for (size_t i = 0; i < buffer_size; ++i) {
-      variable_tensor->data.uint8[i] = 0xAA;
+      variable_tensor_buffer[i] = 0xAA;
     }
 
     // Reset the variable tensor.
@@ -59,7 +62,7 @@ TF_LITE_MICRO_TEST(TestResetVariableTensor) {
 
     // Verify that the variable tensor is zeroed out.
     for (size_t i = 0; i < buffer_size; ++i) {
-      TF_LITE_MICRO_EXPECT_EQ(0, variable_tensor->data.uint8[i]);
+      TF_LITE_MICRO_EXPECT_EQ(0, variable_tensor_buffer[i]);
     }
   }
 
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/micro_interpreter_graph.cc
```diff
@@ -329,6 +317,26 @@ TfLiteStatus MicroInterpreterGraph::ResetVariableTensors() {
...
+    int value = 0;
+    if (tensor->type() == tflite::TensorType_INT8) {
+      value = tensor->quantization()->zero_point()->Get(0);
+    }
+    memset(subgraph_allocations_[subgraph_index].tensors[tensor_index].data.raw,
+           value, buffer_size);
+  }
+  return kTfLiteOk;
+}
+
```
ddavis-2015, line 367: Since this only works for TensorFlow 1.0 variable type tensors, the default return value should be failure.  Also TensorType_INT8 should not be used for quantization checks.  Instead the existence of the `quantization` field in the schema should be used.  Multi-channel quantization checks also need to be here, or alternatively fail this method.

ddavis-2015:
@veblush This will not work for RESOURCE variable tensors (TensorFlow 2.0)

tensorflow/lite/micro/micro_interpreter_graph.cc
veblush: > Since this only works for TensorFlow 1.0 variable type tensors, the default return value should be failure. 

Done.

> Also TensorType_INT8 should not be used for quantization checks. Instead the existence of the quantization field in the schema should be used. Multi-channel quantization checks also need to be here, or alternatively fail this method.

This reset logic is identical to the existing ResetVariableTensors. We can improve it in a separate PR if this is needed.

tensorflow/lite/micro/micro_interpreter_graph_test.cc
ddavis-2015, line 1: Only the use of `data.data` is allowed from the `TfLitePtrUnion` struct.  All other members have been deprecated by the LiteRT team (formerly the TfLite folks).

ddavis-2015, line 1: Please be advised that `TF_LITE_MICRO_EXPECT_` macros do not end the test.  This could cause the test to crash on null pointer checks, and the logs for the test will be lost.

Perhaps we should add a new macro:
```
#define TF_LITE_MICRO_CHECK_FAIL_AND_STOP if (micro_test::did_test_fail) continue
```

tensorflow/lite/micro/micro_interpreter_graph.cc
```diff
@@ -329,6 +321,48 @@ TfLiteStatus MicroInterpreterGraph::ResetVariableTensors() {
...
+      tensor, &subgraph_allocations_[subgraph_index].tensors[tensor_index]);
+}
+
+TfLiteStatus MicroInterpreterGraph::ResetTensorData(
+    const tflite::Tensor* tensor, TfLiteEvalTensor* eval_tensor) {
+  size_t buffer_size;
+  TF_LITE_ENSURE_STATUS(TfLiteEvalTensorByteLength(eval_tensor, &buffer_size));
+
+  int value = 0;
+  if (tensor->type() == tflite::TensorType_INT8) {
```
ddavis-2015: You're right, a multi-channel tensor variable probably doesn't exist.  But this line would be better as:
```
if (tensor->type() == tflite::TensorType_INT8 && tensor->quantization() && tensor->quantization()->zero_point() && tensor->quantization->zero_point()->size() > 0) {
```

tensorflow/lite/micro/micro_interpreter_graph_test.cc
ddavis-2015, line 1: Also, this file needs to be added to `MICROLITE_TEST_SRCS` in the Makefile (around line 339)

tensorflow/lite/micro/micro_interpreter_graph.cc
veblush: Done!

tensorflow/lite/micro/micro_interpreter_graph_test.cc
veblush: Done!

veblush: Good point. I just introduced a whole set of EXPECTs.

veblush: Would you elaborate? What is `data.data`?

ddavis-2015: `tensor->data.uint8` and `tensor->data.raw`, etc.  These are all deprecated.  Only `tensor->data.data` with a cast operation is supposed to be used.

veblush: Ah. got it. Done!

PR #3325, opened 2026-01-23
Fix StackAllocator memory corruption due to alignment

veblush:
Fixes a memory corruption in `tensorflow/lite/micro/test_helpers.cc`. StackAllocator's alignment offset caused `FlatBufferBuilder` to receive a pointer range extending beyond the backing store. Since `FlatBufferBuilder` writes from the end of the buffer, this caused immediate out-of-bounds writes into `StackAllocator`'s member variables.

The fix increases `data_backing_` size to accommodate alignment, initializes `data_size_` to the offset, and checks bounds against the physical backing size. This ensures the memory provided to `FlatBufferBuilder` is always valid.

In addition to that, I doubled the size of the StackAllocator internal buffer to resolve allocation failures in compression tests.

```
[ RUN      ] MicroInterpreterTest.TestInterpreterCompressionAltMemory
allocate(size=12368) full: data_size_=8200
```

🎂 & 🎁

BUG=n/a

### DIFF
--- a/tensorflow/lite/micro/test_helpers.cc
+++ b/tensorflow/lite/micro/test_helpers.cc
@@ -48,40 +48,32 @@ namespace {
 
 class StackAllocator : public flatbuffers::Allocator {
  public:
-  StackAllocator(size_t alignment) : data_size_(0) {
-    data_ = AlignPointerUp(data_backing_, alignment);
-    data_size_ = data_ - data_backing_;
-  }
+  StackAllocator() : offset_(0) {}
 
   uint8_t* allocate(size_t size) override {
-    if ((data_size_ + size) > sizeof(data_backing_)) {
-      MicroPrintf("allocate(size=%u) full: data_size_=%u", size, data_size_);
+    if (offset_ + size > kBufferSize) {
+      MicroPrintf("allocate(size=%u) full: offset_=%u", size, offset_);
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/test_helpers.cc
```diff
@@ -69,10 +74,12 @@ class StackAllocator : public flatbuffers::Allocator {
...
-  static constexpr size_t kStackAllocatorSize = 8192;
+  static constexpr size_t kStackAllocatorSize = 16384;
 
  private:
-  uint8_t data_backing_[kStackAllocatorSize];
+  static constexpr size_t kStackAllocatorBackingSize =
+      kStackAllocatorSize + MicroArenaBufferAlignment();
+  uint8_t data_backing_[kStackAllocatorBackingSize];
   uint8_t* data_;
   int data_size_;
```
ddavis-2015: @veblush change to size_t to match the MicroPrintf above?

veblush: What do you mean?

tensorflow/lite/micro/test_helpers.cc
ddavis-2015: This appears to be out-of-sync with StackAllocator::data_size_.

PR #3349, opened 2026-01-25
Correctly initialize `ae_int8x8` zero point vectors by explicitly pac…

shlmregev:
…king scalar values.

The implicit broadcast doesn't compile with toolchain version RI-2020.5-linux and a hifi5 core.

BUG=478153404

### DIFF
--- a/tensorflow/lite/micro/kernels/xtensa/xtensa_decode_state_prune.cc
+++ b/tensorflow/lite/micro/kernels/xtensa/xtensa_decode_state_prune.cc
@@ -189,8 +189,7 @@ void XtensaDecodeStatePrune::DecompressToBufferPerChannelInt8_Xtensa(
   AE_DBI_IP((const unsigned short*)p_stream, 16);
 
   for (size_t channel = 0; channel < num_channels_; channel++) {
-    int8_t scalar_zero = zero_points_[channel];
-    ae_int8x8 zero = AE_MOVDA8((int)scalar_zero);
+    ae_int8x8 zero = AE_MOVDA8((int)zero_points_[channel]);
 
     uint32_t mask_low, mask_high;
 


tensorflow/lite/micro/kernels/xtensa/xtensa_decode_state_prune.cc
```diff
@@ -75,7 +75,8 @@ void XtensaDecodeStatePrune::DecompressToBufferInt8_Xtensa(void* buffer) {
   ae_int8x8 shfl0, shfl1, shfl2, shfl3;
   const int count = count_indices_;
   int8_t* __restrict pCoeff = static_cast<int8_t*>(buffer);
-  ae_int8x8 zero = single_zero_point_;
+  int32_t packed32_zero = single_zero_point_ | (single_zero_point_ << 8) | (single_zero_point_ << 16) | (single_zero_point_ << 24);
```
ddavis-2015: @shlmregev I think you need to account for integer promotion

Or perhaps just AE_MOVDA8 with a cast

tensorflow/lite/micro/kernels/xtensa/xtensa_decode_state_prune.cc
```diff
@@ -189,7 +190,9 @@ void XtensaDecodeStatePrune::DecompressToBufferPerChannelInt8_Xtensa(
   AE_DBI_IP((const unsigned short*)p_stream, 16);
 
   for (size_t channel = 0; channel < num_channels_; channel++) {
-    ae_int8x8 zero = zero_points_[channel];
+    int8_t scalar_zero = zero_points_[channel];
+    int32_t packed32_zero = scalar_zero | (scalar_zero << 8) | (scalar_zero << 16) | (scalar_zero << 24);
```
ddavis-2015: @shlmregev I think you need to account for integer promotion

Or perhaps just AE_MOVDA8 with a cast

tensorflow/lite/micro/kernels/xtensa/xtensa_decode_state_prune.cc
shlmregev: Thanks @ddavis-2015, the sign extension indeed wasn't handled correctly.
And thanks for the tip to use AE_MOVDA8.

Issue #3128, opened 2025-07-01
feature request: hermetic Python toolchain in Bazel


rkuester:
Currently, TFLM's Bazel build uses the system Python installation, which can lead to reproducibility issues and dependency conflicts. A hermetic Python toolchain would ensure all developers and CI use the exact same Python version and environment.

### Current Problems

When system Python dependencies mix with Bazel's pip repositories, we see issues like:

- **Version conflicts**: System-installed packages (e.g., `numpy`, `protobuf`) can conflict with versions specified in `python_requirements.txt`
- **Tool inconsistency**: Developers may have different versions of `yapf`, `cpplint`, or other tools installed globally
- **Hidden dependencies**: Builds may accidentally depend on system packages not declared in requirements
- **"Works on my machine"**: Builds succeed locally but fail in CI or on other developers' machines due to different system Python setups

### Proposed Solution

Use rules_python's hermetic toolchain support. Something like:

```python
# In WORKSPACE
load("@rules_python//python:repositories.bzl", "python_register_toolchains")

python_register_toolchains(
    name = "python",
    python_version = "3.10",  # Pin to specific version
)

# Pip dependencies fully isolated from system
load("@rules_python//python:pip.bzl", "pip_parse")
pip_parse(
    name = "pip_deps",
    requirements_lock = "//third_party:python_requirements.txt",
    python_interpreter_target = "@python//:python",  # Use hermetic Python
)
```

This would yield:
- Bazel downloads and manages its own Python interpreter
- All pip dependencies are isolated from system packages
- Same behavior across all platforms and environments

This has been wished for on several occasions: #2027, #2743, #3126.

PR #3364, opened 2026-01-29
Bazel Module

veblush:
Completed the migration to Bzlmod and removed the legacy `WORKSPACE` file.

Key Changes:
   * Bzlmod Adoption: Replaced `WORKSPACE` with `MODULE.bazel`. Removed various legacy support files.
   * Hermetic Python: Configured `rules_python` toolchains (3.10-3.13). (3.14 is not possible at this moment since tensorflow is not available)
   * New python namespace shim `tflite_micro.py` is added to maintain import compatibility (from tflite_micro.tensorflow...) without restructuring the physical source tree.
   * Flatbuffers Cleanup: Simplified `flatbuffers/build_defs.bzl` by using upstream flatbuffers rules.
   * Updated Bazel to 8.x

Fixed https://github.com/tensorflow/tflite-micro/issues/3128

BUG=n/a

### DIFF
--- a/python/tflite_micro/__init__.py
+++ b/python/tflite_micro/__init__.py
@@ -26,4 +26,3 @@ from tflite_micro.python.tflite_micro._version import __version__
 
 # Provide a convenient alias for the compression module
 from tflite_micro.tensorflow.lite.micro import compression
-
--- a/python/tflite_micro/postinstall_check.py
+++ b/python/tflite_micro/postinstall_check.py
@@ -23,9 +23,9 @@ from tflite_micro import compression
 
 import numpy as np
 try:
-    from importlib import resources
+  from importlib import resources
 except ImportError:
-    import importlib_resources as resources
+  import importlib_resources as resources
 import sys
 import tempfile
 import os


tflite_micro.py
rkuester, line 1: I suspect this shim could be avoided by using an `imports` attribute of a well-placed `py_library`. E.g.:

```starlark
# BUILD (@ the repository root)
py_library(
    name = "python_root",
    imports = ["."],
    visibility = ["//visibility:public"],
)
```

This should add the repo root to `PYTHONPATH`.

And then any dep on `//:tflite_micro_shim` can become a dep on `//:python_root`.

python/py_pkg_cc_deps.bzl
rkuester, line 1: `rules_python` now has a `whl_mods` extension that can inject modifications directly into a pip package's Bazel repository, and can take over what we're trying to do here in a supported, non-hacky way. Something like:

```starlark
# python/tensorflow_cc_targets.bzl

def extra_tensorflow_targets():
    cc_library(
        name = "cc_headers",
        hdrs = native.glob(
            ["site-packages/tensorflow/include/**"],
        ),
        includes = ["site-packages/tensorflow/include"],
        visibility = ["//visibility:public"],
    )

    cc_library(
        name = "cc_library",
        srcs = ["site-packages/tensorflow/libtensorflow_framework.so.2"],
        deps = [":cc_headers"],
        visibility = ["//visibility:public"],
    )
```
```starlark
# MODULE.bazel

whl_mods = use_extension("@rules_python//python/extensions:pip.bzl", "whl_mods")

whl_mods.create(
    hub_name = "tflm_pip_deps",
    whl_name = "tensorflow",
    additive_build_content = """\
load("@tflite_micro//python:tensorflow_cc_targets.bzl", "extra_tensorflow_targets")
extra_tensorflow_targets()
""",
)
```

Then in each `pip.parse` call, wire up the modifications and aliases:

```starlark
pip.parse(
    hub_name = "tflm_pip_deps",
    ...
    whl_modifications = {
        "@tflm_pip_deps_whl_mods//:tensorflow.json": "tensorflow",
    },
    extra_hub_aliases = {
        "tensorflow": ["cc_headers", "cc_library"],
    },
)
```

And then at the point of use:

```starlark
deps = ["@tflm_pip_deps//tensorflow:cc_headers"]
```

tflite_micro.py
veblush: I tried this idea first but it didn't work out. Python appears to have become more strict on this imports.

python/py_pkg_cc_deps.bzl
veblush: Neat! I tried it and see the "Review update" commit.

rkuester: Nice! That looks elegant.

PR #3391, opened 2026-01-31
Missing kTfLiteInt16 in ComparisonsPrepare

veblush:
BUG=n/a

### DIFF
--- a/tensorflow/lite/micro/kernels/comparisons.cc
+++ b/tensorflow/lite/micro/kernels/comparisons.cc
@@ -559,7 +559,7 @@ TfLiteStatus ComparisonsPrepare(TfLiteContext* context, TfLiteNode* node) {
       micro_context->AllocateTempInputTensor(node, kInputTensor2);
   TF_LITE_ENSURE(context, input2 != nullptr);
 
-  if (input1->type == kTfLiteInt8) {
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/kernels/comparisons.cc
```diff
@@ -559,7 +559,7 @@ TfLiteStatus ComparisonsPrepare(TfLiteContext* context, TfLiteNode* node) {
       micro_context->AllocateTempInputTensor(node, kInputTensor2);
   TF_LITE_ENSURE(context, input2 != nullptr);
 
-  if (input1->type == kTfLiteInt8) {
+  if (input1->type == kTfLiteInt8 || input1->type == kTfLiteInt16) {
```
ddavis-2015, line 562: @veblush should also have a check if the tensor is in fact quantized (unless the TfLite converter always produces quantized INT16???).  I always assumed un-quantized INT16 was just as valid as unquantized INT32.  Also add an `is_quantized` bool to the `OpData`?

tensorflow/lite/micro/kernels/comparisons.cc
ddavis-2015, line 1: @veblush The actual comparison Eval methods should also have INT16 support added.  Not all LiteRT comparisons have INT16 support, and it is confusing how some LiteRT comparisons assume (or don't) that the INT16 inputs are not quantized.

PR #3400, opened 2026-02-03
feat(compression): update tooling to use DECODE operators

rkuester:
This is a draft PR for running CI, review, and seeing the commits in
context. The commits along this branch will be individually submitted
for merge.

This obsoletes the original feat-decode branch (#3257), which has been
reworked to address review feedback.

See the linked issue for a description of the change.

BUG=implements #3256


python/tflite_micro/python_ops_resolver.cc
```diff
@@ -40,7 +40,9 @@ PythonOpsResolver::PythonOpsResolver() {
   AddConv2D();
   AddCos();
   AddCumSum();
+#ifdef USE_TFLM_COMPRESSION
   AddDecode();
+#endif
```
ddavis-2015: AddDecode shouldn't be dependent on `USE_TFLM_COMPRESSION` (none of the DECODE code is conditionally compiled)

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,268 @@
...
+      decode_op = model_editor.Operator(
+          opcode=tflite.BuiltinOperator.CUSTOM,
+          custom_code=DECODE_CUSTOM_OP_NAME,
+          inputs=[info.tensor, ancillary_tensor],
+          outputs=[output_tensor],
+      )
+
+      # Insert DECODE immediately before this consumer
+      insert_pos = subgraph.operators.index(consumer)
+      subgraph.operators.insert(insert_pos, decode_op)
```
ddavis-2015: > This being located here does not allow for a single DECODE operator to have multiple encoded inputs and ancillary tensors. The example would be CONCATENATION which takes multiple inputs, where several might be encoded tensors.

> The compressor design currently uses one DECODE operator per compressed tensor. Looking at the C++ kernel, I see it already supports multiple input/output pairs. Were you expecting us to batch them into a single DECODE for cases like CONCATENATION? Would that run into the same alt decompression memory problems as reusing the output of a single DECODE? Does using multiple DECODEs also run into those issues?

If a single operator (for example CONCATENATION) has multiple compressed tensors, then those tensors and their DCMs should be batched into a single DECODE operator. The DECODE kernel already handles this (as you observed) and handles the alternate decompression memory correctly in this case.

Or to put it more succinctly: multiple encoded tensors for a single operator, MUST be passed as multiple inputs to a single DECODE operator.

tensorflow/lite/micro/compression/compression_integration_test.py
ddavis-2015, line 1: Perhaps add a test where the compression spec is empty?  The original model and the no-spec "compressed" model should give the same results.

tensorflow/lite/micro/compression/decode_insert_test.py
ddavis-2015, line 1: Will need a test where the simple model has an operator with two inputs that are compressed (FULLY_CONNECTED weight + bias, or two inputs of CONCATENATION).  Each encoded tensor could have different bit-width, thus generating different DCM for each.  Or perhaps extend an existing test?

python/tflite_micro/python_ops_resolver.cc
rkuester: Addressed in commit d2ac3cec. The `#ifdef USE_TFLM_COMPRESSION` guard around `AddDecode()` is removed, so DECODE is registered unconditionally. The compression and proprietary integration tests also drop their `with_compression_enabled` gating since DECODE-based models no longer require the flag.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Addressed. Multiple compressed tensor inputs to the same operator are now [batched into a single DECODE](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-07340f9b1e2dd8d5fd5f026e5c8074b7e5168f1a415c0b6e7046001ad9325012R213-R271). The grouping is per-consumer, so a tensor shared across different consumers still gets a separate DECODE before each one to avoid clobbering the alternate decompression memory.

tensorflow/lite/micro/compression/compression_integration_test.py
rkuester: Rather than pass through an empty spec silently, the compressor now [rejects an empty spec](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-1ca95116d18953c4b50538524c131a4443fe18af60a5d351288f567710dea6d1R135-R139) as an error, since it's almost certainly a mistake. There's a [corresponding test](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-defc9e243c4ac63bb60953f79e2a1f6e6bce73d158f9e3f085da8f9dc0568ae3R316-R319).

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed. [`test_multiple_compressed_inputs_batched`](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-3e3e2111b0a641047cf6ef4487a02ac5a3f6be4841e7a9a2e6024322acc6b0d3R388) tests a CONCATENATION with two compressed tensor inputs at different bitwidths, verifying a single DECODE with 4 inputs and 2 outputs where each ancillary tensor carries its own distinct data. [`test_mixed_compressed_and_uncompressed_inputs`](https://github.com/tensorflow/tflite-micro/pull/3400/files#diff-3e3e2111b0a641047cf6ef4487a02ac5a3f6be4841e7a9a2e6024322acc6b0d3R462) covers the case where only one of two CONCATENATION inputs is compressed.

python/tflite_micro/test_compression_unsupported.py
```diff
@@ -12,84 +12,82 @@
...
 
-    # Then load regular model (should succeed)
     interpreter = runtime.Interpreter.from_bytes(model_data)
     self.assertIsNotNone(interpreter)
 
 
 if __name__ == '__main__':
-  # Set TF environment variables to suppress warnings
   os.environ['TF_CPP_MIN_LOG_LEVEL'] = '2'
   os.environ['TF_ENABLE_ONEDNN_OPTS'] = '0'
```
ddavis-2015, line 94: Should add a comment on the meaning of these environment vars

rkuester: Addressed in d40a84e.

rkuester:
Replaced by #3574, which rebuilds this tooling fresh on current main (tf-free tests via model_editor) instead of rebasing this long-stale branch.

PR #3405, opened 2026-02-05
Fixed compression logic for Batch matmul CREF code.

unmeshna017:
Fixed a bug in 3Dx2D tensor use-case dimension compression logic in Batch Matmul CREF code.

### DIFF
--- a/tensorflow/lite/micro/kernels/batch_matmul.cc
+++ b/tensorflow/lite/micro/kernels/batch_matmul.cc
@@ -366,25 +366,29 @@ TfLiteStatus BatchMatMulEval(TfLiteContext* context, TfLiteNode* node) {
   // Compress BatchMatMul when third from last RHS dimension is one.
   int32_t rhs_dims_count = orig_rhs_shape.DimensionsCount();
   int32_t lhs_dims_count = orig_lhs_shape.DimensionsCount();
+  RuntimeShape rhs_shape =
+      adj_y ? orig_rhs_shape : SwapRowColumnDims(orig_rhs_shape);
+  RuntimeShape lhs_shape =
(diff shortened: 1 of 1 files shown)


ddavis-2015:
@unmeshna017 @veblush There does not seem  to be any actual issue with the original code, and this PR appears to just be a rearrangement of the original code.  This PR should be rejected.

1) The original Eval code follows the LiteRT reference code almost line for line, other than changes for TFLM tensor usage and memory allocation.  This code has been in place for years.
2) The special case dimensionality reduction code is correct as-is and meets the intent of the comment:
```
  // Compress ops where rhs shape is [..., 1, X, Y] and lhs shape is
  // [..., Q, R, S] which is equivalent to rhs: [..., X, Y] and
  // lhs: [..., Q * R, S].
```
3) The original code maintains the intent of the comment regarding inference computation:
```
// Perform a batch matrix multiply on
// LHS <..., A, B>  X  RHS<..., B, C>
// where the leading dimensions of LHS and RHS obey broadcasting rules
// (this Op will apply broadcasting rules).
// We assume that LHS and RHS are both row oriented (adjacent values in memory
// are in the same row) and will output in the same memory layout. However,
// our fast GEMM libraries assume RCC layout (LHS row oriented,
// RHS column oriented, output column oriented). Therefore, we perform
// RHS <..., C, B> X LHS <..., B, A>
// where output is a C X A column-oriented, which is equivalent to
// A X C row-oriented.
```
4) The existing unit tests are insufficient for full code coverage of the dimensionality reduction.  Specifically when the LHS shape has a batch size > 1.  This can be remedied through changes to the existing units tests, as per issue #3412.

PR #3439, opened 2026-02-10

rkuester:
models with a clean API. The module eliminates manual index tracking
and buffer management through automatic bookkeeping, supporting both
declarative and imperative construction styles.

Wrapper classes (Tensor, Operator, Subgraph, Model) hold the underlying
flatbuffer T objects as backing storage rather than copying fields into
dataclasses. This ensures all schema fields are preserved during
read-modify-write cycles, even fields not explicitly handled by
model_editor. Future schema additions will be preserved automatically.

Add comprehensive test coverage including field preservation tests that
verify unhandled schema fields survive read-modify-write.

BUG=implements #3256

### DIFF
--- a/tensorflow/lite/micro/compression/BUILD
+++ b/tensorflow/lite/micro/compression/BUILD
@@ -247,6 +247,28 @@ py_test(
     ],
 )
 
+py_library(
+    name = "model_editor",
+    srcs = ["model_editor.py"],
+    deps = [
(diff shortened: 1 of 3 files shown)


rkuester:
Closing along with the obsolete #3400 umbrella. The decode tooling was rebuilt fresh on current main (tf-free) under #3574; a new model_editor submit PR will be sliced from that stack.

PR #3492, opened 2026-03-05
Updated CREF to handle bias=null for FC,Conv 8x16 CREF code

unmeshna017:
This change enhances the CREF kernels for Fully Connected and Convolutional layers to use higher precision accumulation when no bias term is present. Previously, the implementation would default to int32 accumulation when bias=NULL or INT32. This change is applicable only for the Int16x8 kernel type.

Adding @vp-cad and @joshih-cad as watchers.

BUG=3498

### DIFF
--- a/tensorflow/lite/micro/kernels/fully_connected.cc
+++ b/tensorflow/lite/micro/kernels/fully_connected.cc
@@ -54,7 +54,7 @@ TfLiteStatus FullyConnectedPrepare(TfLiteContext* context, TfLiteNode* node) {
   TfLiteTensor* output = micro_context->AllocateTempOutputTensor(
       node, kFullyConnectedOutputTensor);
   TF_LITE_ENSURE(context, output != nullptr);
-  TF_LITE_ENSURE_TYPES_EQ(context, input->type, output->type);
+  //TF_LITE_ENSURE_TYPES_EQ(context, input->type, output->type);
 
   if ((input->type == kTfLiteFloat32 && filter->type != kTfLiteFloat32) ||
       (input->type == kTfLiteInt8 &&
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/kernels/fully_connected.cc
```diff
@@ -54,7 +54,7 @@ TfLiteStatus FullyConnectedPrepare(TfLiteContext* context, TfLiteNode* node) {
   TfLiteTensor* output = micro_context->AllocateTempOutputTensor(
       node, kFullyConnectedOutputTensor);
   TF_LITE_ENSURE(context, output != nullptr);
-  TF_LITE_ENSURE_TYPES_EQ(context, input->type, output->type);
+  //TF_LITE_ENSURE_TYPES_EQ(context, input->type, output->type);
```
rameshkunasi: Any particular reason for disabling this condition check?

unmeshna017: Hi, we need to disable the condition check to support FC layer with hybrid precision.

rameshkunasi: Can you please elaborate this? I don't think we need to disable this condition check for this change. Even though we are changing the precision of bias value it won't change the output type.

unmeshna017: This needs to be disabled to support int8 input and int16 output

joshih-cad: @rameshkunasi - Agree with you. This change is not relevant to this specific PR. It's coming from Cadence's TFLM fork. 
To reiterate @unmeshna017's point, we do want to extend support to hybrid precision FC layers (int8, int16) for which we would want to disable this condition check. We'll create a separate PR for merging this change. Thanks!

rameshkunasi: CC: @veblush 
@joshih-cad thanks for the clarification!

unmeshna017: Hi @rameshkunasi , I've uncommented the condition check as we're planning to create a separate PR to merge this change.

unmeshna017:
Hi, 

For the bias == nullptr case, the changes proposed in this PR takes the 64-bit accumulator path, which helps improve accuracy. In contrast, the current implementation in the repository uses a 32-bit accumulator for the same. Could you please revise the changes to use the 64-bit accumulator path for this case?

ddavis-2015:
Resolved with PRs #3522 and #3537 and #3597

PR #3535, opened 2026-04-17
Validate buffer index before accessing buffers vector

TristanInSec:
Check that the tensor buffer index is within the bounds of the model buffers vector before accessing it, matching the validation already present in the compression code path.

BUG=b/general


TristanInSec:
This fix addresses a confirmed vulnerability. Still needed.

veblush:
Thank you for the PR but it's against the [Error Handling & Defensive Programming Guide](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/error_handling_guide.md).

PR #3559, opened 2026-05-16
 build: pin actions/github-script to commit SHA in pull_request_target workflow

XananasX7:
Pin `actions/github-script` to full commit SHA instead of mutable `v8` tag
  in the `pr_test.yml` workflow.

  This workflow uses `pull_request_target` trigger with access to
  `secrets.TFLM_BOT_REPO_TOKEN`. Pinning to SHA ensures immutability
  and prevents supply chain attacks via tag manipulation.

  Ref: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
       - name: check-scope
         id: check-scope
-        uses: actions/github-script@v8
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
         with:
           github-token: ${{ secrets.TFLM_BOT_REPO_TOKEN }}
           script: |


XananasX7:
## Summary

This PR pins `actions/github-script` to a full commit SHA in the `pull_request_target` workflow.

### Why this matters
This workflow uses `pull_request_target` — a trigger that runs in the context of the **base repository** with access to repository secrets, including `TFLM_BOT_REPO_TOKEN`. The current `actions/github-script@v7` tag is mutable and could be redirected by a supply chain attack.

### Fix
Pinned to the exact commit SHA of `actions/github-script` at v7, with the version annotated in a comment. This is a one-line, no-logic change that eliminates the supply chain risk.

CLA signed. Ready for review.

ddavis-2015:
resolved with PRs #3566 and #3567

PR #3564, opened 2026-05-21
[CMSIS-NN] Fix stateful execution and batch-major striding for CMSIS-NN LSTM

veblush:
### Problem

The current CMSIS-NN LSTM wrapper uses arm_lstm_unidirectional_s8 and arm_lstm_unidirectional_s16. These CMSIS-NN functions are designed for stateless sequence evaluation: they explicitly wipe the cell state at t=0 and ignore any initial hidden state, returning only the sequence outputs.

This breaks TFLM's streaming/embedded ML workloads which rely on stateful LSTMs where the CellStateTensor and HiddenStateTensor persist as variable tensors across Invoke() calls.

Furthermore, CMSIS-NN's internal implementation for batch-major tensors (time_major=false with batch_size > 1) incorrectly jumps memory by time_steps, causing an out-of-bounds read on the contiguous hidden_state buffer.

### Solution

1. Fallback to explicit looping: Implemented a manual time/batch loop within CMSIS_NN_EvalInteger8x8_16Lstm and CMSIS_NN_EvalInteger16x8_16Lstm that bypasses the stateless sequence evaluator and instead iteratively calls the single-step CMSIS-NN kernels (arm_nn_lstm_step_s8 and arm_nn_lstm_step_s16).
1. State Persistence: The fallback loop properly preserves the CellStateTensor and HiddenStateTensor across timesteps and invocations.
1. Stride Bug Bypass: For time_major=false, the loop evaluates one batch at a time (batch_size=1 passed to the kernel), which guarantees cache-friendly contiguous memory reads and avoids CMSIS-NN's batch striding bug entirely.
1. Future-proofing: Introduced #ifdef CMSIS_NN_STATEFUL_LSTM. Once ARM merges a fix upstream to support the optional hidden_state context pointer, this flag will seamlessly switch back to using the native CMSIS-NN sequence evaluator. (https://github.com/ARM-software/CMSIS-NN/pull/219)

BUG=N/A

### DIFF
--- a/tensorflow/lite/micro/kernels/cmsis_nn/unidirectional_sequence_lstm.cc
+++ b/tensorflow/lite/micro/kernels/cmsis_nn/unidirectional_sequence_lstm.cc
@@ -21,6 +21,7 @@ limitations under the License.
 #include <limits>
 
 #include "Include/arm_nnfunctions.h"
+#include "Include/arm_nnsupportfunctions.h"
 #include "tensorflow/lite/kernels/internal/quantization_util.h"
 #include "tensorflow/lite/kernels/kernel_util.h"
 #include "tensorflow/lite/micro/kernels/fully_connected.h"
@@ -270,7 +271,7 @@ TfLiteStatus CMSIS_NN_PortOpData(TfLiteContext* context, OpDataLSTM* params_ref,
 }
 
 TfLiteStatus CMSIS_NN_EvalInteger8x8_16Lstm(
-    const OpData& op_data, const LSTMKernelContents& kernel_content,
+    const OpData& op_data, LSTMKernelContents& kernel_content,
     const LSTMBuffers<int16_t>& buffers) {
   TFLITE_DCHECK(
       kernel_content.GetInternalTensor(tflite::kLstmInputTensor)->dims->size >=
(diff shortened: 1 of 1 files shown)


tensorflow/lite/micro/kernels/cmsis_nn/unidirectional_sequence_lstm.cc
```diff
@@ -282,21 +283,74 @@ TfLiteStatus CMSIS_NN_EvalInteger8x8_16Lstm(
...
+
+      arm_cmsis_nn_status status = arm_nn_lstm_step_s8(
+          data_in, step_hidden_in, hidden_out, &params, &cmsis_buffers, 1);
+      if (status != ARM_CMSIS_NN_SUCCESS) return kTfLiteError;
+      step_hidden_in = hidden_out;
+    }
+    if (params.time_steps > 0) {
+      std::copy_n(step_hidden_in, params.batch_size * params.hidden_size,
+                  hidden_state);
+    }
```
ddavis-2015, line 321: not sure why this is here.  When using the greedy memory planner, the hidden_state may be overwritten by subsequent operator's output(s).  See next comment for more info.

tensorflow/lite/micro/kernels/cmsis_nn/unidirectional_sequence_lstm.cc
```diff
@@ -282,21 +283,74 @@ TfLiteStatus CMSIS_NN_EvalInteger8x8_16Lstm(
...
+        arm_cmsis_nn_status status =
+            arm_nn_lstm_step_s8(data_in, step_hidden_in, hidden_out,
+                                &step_params, &cmsis_buffers, 1);
+        if (status != ARM_CMSIS_NN_SUCCESS) return kTfLiteError;
+        step_hidden_in = hidden_out;
+      }
+      if (params.time_steps > 0) {
+        std::copy_n(step_hidden_in, params.hidden_size,
+                    hidden_state + b * params.hidden_size);
+      }
```
ddavis-2015, line 344: Same as the above comment with this additional info: I have not been able to produce a Colab where the. converter will produce a stateful, fused LSTM operation with quantization.  The converter (and the Colab session) crash every time.  The only time I can make a stateful LSTM in Colab, always produces an unfused LSTM.

tensorflow/lite/micro/kernels/cmsis_nn/unidirectional_sequence_lstm.cc
```diff
@@ -308,15 +362,63 @@ TfLiteStatus CMSIS_NN_EvalInteger16x8_16Lstm(
...
+      int16_t* hidden_out =
+          output + (t * params.batch_size * params.hidden_size);
+
+      arm_cmsis_nn_status status = arm_nn_lstm_step_s16(
+          data_in, hidden_state, hidden_out, &params, &cmsis_buffers, 1);
+      if (status != ARM_CMSIS_NN_SUCCESS) return kTfLiteError;
+
+      // Update hidden state for next step
+      std::copy_n(hidden_out, params.batch_size * params.hidden_size,
+                  hidden_state);
```
ddavis-2015, line 397: Don't understand why this is inside the step loop.  Why not just update the hidden state input pointer as was done in the s8 code?

tensorflow/lite/micro/kernels/cmsis_nn/unidirectional_sequence_lstm.cc
```diff
@@ -308,15 +362,63 @@ TfLiteStatus CMSIS_NN_EvalInteger16x8_16Lstm(
...
+        int16_t* current_hidden = hidden_state + b * params.hidden_size;
+        cmsis_buffers.cell_state = cell_state + b * params.hidden_size;
+
+        arm_cmsis_nn_status status =
+            arm_nn_lstm_step_s16(data_in, current_hidden, hidden_out,
+                                 &step_params, &cmsis_buffers, 1);
+        if (status != ARM_CMSIS_NN_SUCCESS) return kTfLiteError;
+
+        // Update hidden state for next step
+        std::copy_n(hidden_out, params.hidden_size, current_hidden);
```
ddavis-2015, line 417: Don't understand why this is inside the step loop. Why not just update the hidden state input pointer as was done in the s8 code?

suleshahid:
Could we add test case where its failing before/working after the fix?

veblush:
Closed this issue because [https://github.com/tensorflow/tflite-micro/pull/3607](https://github.com/tensorflow/tflite-micro/pull/3607) has been merged. This upgrades CMSIS-NN, which includes the fix on their end.

tensorflow/lite/micro/kernels/cmsis_nn/unidirectional_sequence_lstm.cc
veblush: `hidden_state` is a variable tensor keeping the state so this tensor would be preserved at all times

tensorflow/lite/micro/kernels/cmsis_nn/unidirectional_sequence_lstm.cc
veblush: I've create a test tflite file not from Tensorflow but from flatbuffer ;)

tensorflow/lite/micro/kernels/cmsis_nn/unidirectional_sequence_lstm.cc
veblush: Yep. You're right.

PR #3573, opened 2026-05-26
build(bazel): inject tflite_micro shim via tflm_py_* wrappers

rkuester:
Python targets in this repository import one another under the
"tflite_micro" package namespace, which //:tflite_micro_shim synthesizes
at import time. The shim is required under Bzlmod, where the main
repository's runfiles root is the fixed name "_main" rather than the
module name, so the "tflite_micro" prefix no longer resolves on its own.
Every such target therefore had to list //:tflite_micro_shim in its
deps, which was repetitive and easy to forget.

Add tflm_py_library, tflm_py_test, and tflm_py_binary wrappers in a new
//python:py_rules.bzl that inject the shim dependency automatically,
following the naming convention of the existing tflm_cc_* wrappers, and
document the shim's rationale there. Convert every target that
previously listed the shim to the corresponding wrapper and drop the
explicit dependency. The dependency graph is unchanged; only the means
by which the shim is attached differs.

BUG=part of #3128

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -21,6 +21,11 @@ config_setting(
     },
 )
 
+# Synthesizes the "tflite_micro" top-level package namespace at import time, so
+# that `from tflite_micro...` imports resolve (see tflite_micro.py). Necessary
+# because Bzlmod fixes the main repo's runfiles root to "_main" rather than the
+# module name. Targets should generally not depend on this directly: use the
+# tflm_py_* wrappers in //python:py_rules.bzl, which add it automatically.
 py_library(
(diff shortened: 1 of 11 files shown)


veblush (dismissed):

rkuester:
It looks like the buildifier in the latest tflm-ci image (as of September 2025) has stronger opinions on the ordering of load statements than did older buildifier versions.  I'll fix the ordering, and update my local tools to match CI.

veblush (approved):

PR #3575, opened 2026-05-26

rkuester:
models with a clean API. The module eliminates manual index tracking
and buffer management through automatic bookkeeping, supporting both
declarative and imperative construction styles.

Wrapper classes (Tensor, Operator, Subgraph, Model) hold the underlying
flatbuffer T objects as backing storage rather than copying fields into
dataclasses. This ensures all schema fields are preserved during
read-modify-write cycles, even fields not explicitly handled by
model_editor. Future schema additions will be preserved automatically.

Add comprehensive test coverage including field preservation tests that
verify unhandled schema fields survive read-modify-write.

BUG=part of #3256

### DIFF
--- a/tensorflow/lite/micro/compression/model_editor.py
+++ b/tensorflow/lite/micro/compression/model_editor.py
@@ -1,0 +1,815 @@
+# Copyright 2026 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Unified TFLite model manipulation module.
+
+Provides a clean API for creating, reading, and modifying TFLite models.
+"""
+
+from dataclasses import dataclass, field
+from typing import Optional, Union, List
+import numpy as np
+import flatbuffers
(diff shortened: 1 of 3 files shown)


tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -0,0 +1,826 @@
...
+  """Convert TFLite dtype to numpy dtype."""
+  type_map = {
+      tflite.TensorType.INT8: np.int8,
+      tflite.TensorType.INT16: np.int16,
+      tflite.TensorType.INT32: np.int32,
+      tflite.TensorType.INT64: np.int64,
+      tflite.TensorType.UINT8: np.uint8,
+      tflite.TensorType.FLOAT32: np.float32,
+  }
+  return type_map.get(dtype, np.uint8)
```
veblush: Maybe we want to cover all TensorType here or throw an error? Otherwise, it will fall back to uint8 which could be a surprise.

veblush (commented):
Thanks for the PR!

tensorflow/lite/micro/compression/model_editor.py
rkuester: Good catch, agreed.

Looking this over, I realize the same thing is already duplicated in main and will be duplicated one more time by the end of this commit series. So I've hoisted it into its own utility module that can be used in all three places, addressed the completeness and error issues, and pushed that as #3578. Once that lands, I'll rebase this PR and the rest of the series onto it, and re-request your review here.

veblush (approved):

rkuester:
@veblush this got auto-dequeued by the merge-queue bot last night, but it looks like a transient runner failure rather than a problem with the code. In the `merge_group` run ([26607809382](https://github.com/tensorflow/tflite-micro/actions/runs/26607809382)) four jobs died at container startup (`Executing the custom container implementation failed`), so they never actually ran. PR checks on the same head are green. Mind re-enqueuing?

PR #3578, opened 2026-05-27
refactor(compression): hoist numpy dtype map into tensor_type

rkuester:
Add a tensor_type module that holds the single mapping from a TFLite
TensorType to a numpy dtype, and convert view.py to use it. The mapping
was inlined in view.py; centralizing it gives the compression tooling
one place to maintain as more callers need to read tensor buffers as
numpy arrays.

tensor_type.to_numpy() raises ValueError for types with no clean numpy
equivalent (STRING, RESOURCE, VARIANT, BFLOAT16, and the sub-byte
integer types) instead of silently returning a wrong dtype. Only types
with an unambiguous little-endian numpy representation are mapped.

BUG=part of #3256

### DIFF
--- a/tensorflow/lite/micro/compression/tensor_type.py
+++ b/tensorflow/lite/micro/compression/tensor_type.py
@@ -1,0 +1,61 @@
+# Copyright 2026 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Single source of truth for mapping a TFLite TensorType to a numpy dtype.
+
+Compression tooling reads tensor buffer bytes as numpy arrays, so it needs to
+know the element type. Only the TensorTypes with a clean numpy equivalent are
+mapped; anything else raises rather than silently guessing a type.
+"""
+
+import numpy as np
+
+from tflite_micro.tensorflow.lite.python import schema_py_generated as tflite
+
+# TFLite buffers are little-endian, so the dtypes are pinned to little-endian
+# byte order to keep np.frombuffer correct on any host.
+_TO_NUMPY = {
+    tflite.TensorType.FLOAT16: np.dtype("<f2"),
+    tflite.TensorType.FLOAT32: np.dtype("<f4"),
+    tflite.TensorType.FLOAT64: np.dtype("<f8"),
+    tflite.TensorType.INT8: np.dtype("<i1"),
+    tflite.TensorType.INT16: np.dtype("<i2"),
+    tflite.TensorType.INT32: np.dtype("<i4"),
+    tflite.TensorType.INT64: np.dtype("<i8"),
+    tflite.TensorType.UINT8: np.dtype("<u1"),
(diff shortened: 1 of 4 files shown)


tensorflow/lite/micro/compression/tensor_type_test.py
```diff
@@ -0,0 +1,45 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import numpy as np
+
+from tflite_micro.tensorflow.lite.python import schema_py_generated as tflite
+from tflite_micro.tensorflow.lite.micro.compression import tensor_type
```
veblush, line 20: Which one is recommended? having `tflite_micro.` or not? view.py doesn't use it though.

veblush (approved):

tensorflow/lite/micro/compression/tensor_type_test.py
rkuester: Good question. Using `tflite_micro` is strongly preferred because that is the only form that will work if the module gets included in our public Python package.  A module that says, outside of Bazel's specially crafted environment, `import tensorflow.*` is bound to get an import error or code from the wrong package.

I did not change `view.py` because I was mindlessly matching the imports that were already there (and it doesn't ship in the package). But you're right to note it's technically wrong: it sets a bad example, and it would break if we ever decided to include it in the package. I'll follow up with an independent patch for that.

rkuester:
Force-pushing a rebase onto latest main to clear a stuck merge. No code
changes. Applying ci:ready and ci:full together fired two concurrent CI
runs; the concurrency group cancelled one, which left a stale failing
`tests-passed` required check that blocks the merge. Rebasing onto a
clean commit gives a single fresh run.

tensorflow/lite/micro/compression/tensor_type.py
```diff
@@ -0,0 +1,61 @@
...
+    tflite.TensorType.FLOAT64: np.dtype("<f8"),
+    tflite.TensorType.INT8: np.dtype("<i1"),
+    tflite.TensorType.INT16: np.dtype("<i2"),
+    tflite.TensorType.INT32: np.dtype("<i4"),
+    tflite.TensorType.INT64: np.dtype("<i8"),
+    tflite.TensorType.UINT8: np.dtype("<u1"),
+    tflite.TensorType.UINT16: np.dtype("<u2"),
+    tflite.TensorType.UINT32: np.dtype("<u4"),
+    tflite.TensorType.UINT64: np.dtype("<u8"),
+}
```
ddavis-2015, line 39: TensorType.RESOURCE should be np.dtype("uint32").  Models constructed using Keras can often have RESOURCE tensors in them.

ddavis-2015 (changes requested):

tensorflow/lite/micro/compression/tensor_type.py
rkuester: I'm not sure RESOURCE can/should get mapped to a numpy type. In the schema, it seems like more of an abstract type for the sake of the graph, and up to the runtime how to represent it. Would it ever even have a non-empty value in a .tflite model?

ddavis-2015: @rkuester It has no buffer in a .tflite schema model, with unique ID supplied through the VAR_HANDLE operator parameters.  So yes, it is a runtime representation.  However, any tool seeing a RESOURCE tensor should not throw an exception just because there is no direct mapping to an numpy dtype.

rkuester: Agreed, but a tool shouldn't be trying to map a RESOURCE to a numpy dtype. If it's trying to make such a conversion by calling this function, the result should be "caller, you're doing something invalid." If the tool crashes out on that, the tool isn't handling the case well and should be fixed.

ddavis-2015 (approved):

PR #3580, opened 2026-05-29
refactor(compression): migrate compress.py from model_facade to model_editor

rkuester:
Replace model_facade with model_editor in compress.py and tests.
model_editor provides a cleaner API with better buffer and metadata
handling.

Update BUILD dependencies accordingly.

BUG=part of #3256


rkuester:
@veblush another flaky one. `Cortex-M Corstone 300 (Arm Clang)` hit a 502 from `gitlab.arm.com` and never built. Mind re-running the failed job?

veblush (approved):

Issue #3582, opened 2026-05-31
security: Multiple Integer Overflows in Tensor Dimension Calculations and Quantization Allocation


deezsecc:
### **Description:**
I have identified two integer overflow issues within the model loading and allocation code that can lead to memory corruption (heap out-of-bounds read/write).

---

#### **Issue 1: Signed Integer Overflow in `BytesRequiredForTensor`**
* **Location:** `tensorflow/lite/micro/memory_helpers.cc`
* **Function:** `tflite::BytesRequiredForTensor`
* **Details:**
  When calculating the element count of a tensor shape, the dimension values are multiplied sequentially using signed `int` arithmetic without boundary checks:
  ```cpp
  if (flatbuffer_tensor.shape() != nullptr) {
    for (size_t n = 0; n < flatbuffer_tensor.shape()->size(); ++n) {
      element_count *= flatbuffer_tensor.shape()->Get(n);
    }
  }
  ```
  If a model specifies large/overflowing dimensions (e.g. `[65536, 65536]`), the product overflows, wrapping around to `0` or negative values. This leads to `0`-byte allocation for the tensor's activation buffer in the arena while the tensor metadata (`allocated_tensor.dims`) holds the original huge sizes, resulting in heap out-of-bounds read/write access during kernel execution.

---

#### **Issue 2: Integer Overflow in Quantization Channels Allocation**
* **Location:** `tensorflow/lite/micro/micro_allocator.cc`
* **Function:** `tflite::internal::InitializeTfLiteTensorFromFlatbuffer`
* **Details:**
  When allocating space for per-channel quantization parameters (zero points), the code calls `TfLiteIntArrayGetSizeInBytes(channels)`. On 32-bit platforms, if the number of channels (`src_quantization->scale()->size()`) is controlled (e.g. `0x3fffffff`), the calculation `4 + 4 * channels` overflows to `0` modulo $2^{32}$. 
  The allocator returns a valid pointer to a `0`-byte buffer, but the loop continues writing up to `0x3fffffff` zero point elements into the buffer, triggering a massive heap out-of-bounds write.

---

#### **Suggested Fix**
I have created patches that add standard checked integer multiplication and boundaries checks to reject malformed/overflowing shape and channel parameters with `kTfLiteError`.

A Pull Request containing these fixes and accompanying regression tests will be submitted shortly.

deezsecc:
Requesting review.

PR #3583, opened 2026-05-31
security: prevent integer overflows in shape multiplication and quantization allocation

deezsecc:
##  Summary

This PR addresses the bug reported in #3582 

### `memory_helpers.cc` — Guard against shape dimension overflow
- Added `<limits>` include.
- Before multiplying `element_count` by each dimension, validates that the dimension is non-negative and that the multiplication would not exceed `std::numeric_limits<int>::max()`. Returns `kTfLiteError` immediately if either check fails.

### `micro_allocator.cc` — Guard against quantization channel count overflow
- Added `<limits>` and `<new>` includes.
- Before computing the allocation size for per-channel quantization params (`TfLiteIntArray`), validates that `channels` is positive and that `sizeof(TfLiteIntArray) + channels * sizeof(int)` would not overflow `size_t`. Returns `kTfLiteError` with a diagnostic message if the check fails.

## Testing

Two new tests added to `micro_allocator_test.cc`:

- **`TestTensorShapeIntegerOverflow`** — constructs a flatbuffer tensor with shape `[65536, 65536]` (product overflows `int`) and asserts that `InitializeTfLiteTensorFromFlatbuffer` returns `kTfLiteError`.
- **`TestQuantizationChannelsIntegerOverflow`** — on 32-bit targets (`sizeof(size_t) == 4`), crafts a malformed flatbuffer with a fake channel count of `0x3fffffff` (triggers `TfLiteIntArrayGetSizeInBytes` overflow) and asserts that `InitializeTfLiteTensorFromFlatbuffer` returns `kTfLiteError`.

BUG=#3582

### DIFF
--- a/tensorflow/lite/micro/memory_helpers.cc
+++ b/tensorflow/lite/micro/memory_helpers.cc
@@ -17,6 +17,7 @@ limitations under the License.
 
 #include <cstddef>
 #include <cstdint>
+#include <limits>
 
 #include "flatbuffers/flatbuffers.h"  // from @flatbuffers
 #include "tensorflow/lite/c/common.h"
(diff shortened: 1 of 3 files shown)


veblush:
Closed in favor of #3606

Issue #3552, opened 2026-05-08
Integer overflow in tensor byte-size calculation in memory_helpers.cc


Forbiddem:
## Summary

`tensorflow/lite/micro/memory_helpers.cc` computes the byte length of tensors using a 32-bit signed `int` running product before assigning the result to a `size_t`. Because tensor shape dimensions and element type sizes come from the (untrusted) flatbuffer model, a model with a shape such as `[65536, 65536]` over a 4-byte element type produces an element-count product of 2^32, which wraps to `0` in the existing code and causes `*bytes = 0`.

The same pattern appears in three places:

- `BytesRequiredForTensor` (called during initial allocation from the flatbuffer model)
- `TfLiteEvalTensorByteLength`
- `AllocateOutputDimensionsFromInput`

A downstream allocator that trusts `*bytes` then operates on a tensor whose advertised size is `0` while its dimensions imply ~17 GiB of storage, which is a memory corruption surface when models are loaded from an untrusted source.

## Reproducer

The arithmetic from `BytesRequiredForTensor` extracted into a self-contained C++ program:

```cpp
size_t VulnerableBytesRequiredForTensor(const int32_t* shape, size_t rank,
                                        size_t type_size) {
  int element_count = 1;
  for (size_t n = 0; n < rank; ++n) {
    element_count *= shape[n];
  }
  return element_count * type_size;
}

int main() {
  const int32_t shape[] = {65536, 65536};
  const size_t type_size = 4;  // float32
  // Mathematically: 17,179,869,184 bytes.
  // Observed:       0 (signed 32-bit wraparound).
  std::cout << VulnerableBytesRequiredForTensor(shape, 2, type_size) << "\n";
}
```

## Proposed fix

See PR — running product moved to `size_t`, every multiplication guarded against `size_t` overflow, negative dimensions rejected, and three regression tests added in `memory_helpers_test.cc`.

ZX41R:
One detail to keep separate in the patch: `AllocateOutputDimensionsFromInput()` currently uses the same `size` variable for byte-count accumulation and then passes it to `TfLiteIntArrayGetSizeInBytes(size)`. That helper wants the number of dimension entries, not the tensor byte length.

[...]

PR #3588, opened 2026-06-03
fix: prevent integer overflow in tensor byte-size calculations

Ashutosh0x:
## Summary

Fix signed integer overflow in `BytesRequiredForTensor`, `TfLiteEvalTensorByteLength`, and `AllocateOutputDimensionsFromInput` that leads to heap out-of-bounds read/write when loading untrusted models.

## Vulnerability (#3552)

All three functions compute a running product of tensor dimensions using signed `int` arithmetic. A model with shape `[65536, 65536]` over `float32` requires 2^34 bytes (17 GiB), but the signed `int` product wraps to `0`. This causes a 0-byte allocation while the tensor metadata holds the original dimensions, leading to heap OOB read/write during kernel execution.

## Fix

- Switch running product from `int` to `size_t`
- Add checked multiplication (`SIZE_MAX / x`) before each multiply
- Reject dimensions `<= 0`
- Return `kTfLiteError` on overflow instead of silently wrapping

All three vulnerable functions in `memory_helpers.cc` are fixed with the same pattern.

BUG=#3552

### DIFF
--- a/tensorflow/lite/micro/memory_helpers.cc
+++ b/tensorflow/lite/micro/memory_helpers.cc
@@ -106,12 +106,23 @@ TfLiteStatus TfLiteTypeSizeOf(TfLiteType type, size_t* size) {
 
 TfLiteStatus BytesRequiredForTensor(const tflite::Tensor& flatbuffer_tensor,
                                     size_t* bytes, size_t* type_size) {
-  int element_count = 1;
+  // Use size_t to avoid signed integer overflow. Validate each multiplication
+  // to reject malformed models with dimensions that would wrap around.
+  size_t element_count = 1;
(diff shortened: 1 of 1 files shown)


Ashutosh0x:
Hi @advaitjain — this fixes the integer overflow in tensor byte-size calculations reported in #3552.

All three functions (BytesRequiredForTensor, TfLiteEvalTensorByteLength, AllocateOutputDimensionsFromInput) use signed int for the running product of tensor dimensions, which wraps to 0 on shapes like [65536, 65536]. The allocator then gives a 0-byte buffer while the tensor metadata says it's huge — classic heap OOB write.

The fix uses size_t with checked multiplication (SIZE_MAX / x) before each multiply, and rejects dimensions <= 0 with kTfLiteError. Happy to adjust!

PR #3589, opened 2026-06-04
feat(compression): add DECODE operator types and metadata

rkuester:
Add a decode module with DecodeType constants and DecodeCommonMetadata,
per the TFLM DECODE Operator Design document. This is the first piece of
the DECODE compression tooling that builds on the model_editor migration.

BUG=part of #3256

### DIFF
--- a/tensorflow/lite/micro/compression/decode.py
+++ b/tensorflow/lite/micro/compression/decode.py
@@ -1,0 +1,238 @@
+# Copyright 2026 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""DECODE compression module."""
+
+# Implements the DECODE operator compression scheme described in the
+# "TFLM DECODE Operator Design" document, revised May 20, 2025.
+#
+# The DECODE operator transforms an encoded tensor, alongside a paired
+# ancillary data tensor, into a tensor ready for use as input to any
+# operator. For example, an encoded tensor might contain compressed
+# data, while the paired ancillary data tensor holds the information
+# necessary for decompression. The DECODE operator's output is a fully
+# decompressed tensor.
+#
+# DECODE operators are inserted into the TfLite model subgraph
+# immediately before each operation that uses a decodable tensor as
+# input.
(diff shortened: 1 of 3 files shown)


rkuester:
@veblush

Every open PR seems to show "1 expected check waiting for a workflow to run." It looks broken, but it appears to be CodeQL. "Analyze (python)" looks like a required status check, yet CodeQL default setup seems to run only on merge_group and push, not pull_request. So on an open PR nothing reports that context and it seems to sit "expected" until enqueue, when the merge_group run posts it. Merges look unaffected.

That noise is misleading: it's fooled a couple of us into thinking a check was stuck.

A few options might be:
1. Add pull_request to CodeQL's triggers.
2. Drop "Analyze (python)" from required checks and let the merge queue enforce it.
3. Require it only in the queue.

tensorflow/lite/micro/compression/decode.py
```diff
@@ -0,0 +1,240 @@
...
+  """
+  decode_type: DecodeType
+  version: int = 1
+  user_data: bytes = b'\x00' * 12
+
+  def to_bytes(self) -> bytes:
+    """Serialize DCM to 16-byte sequence."""
+    decode_code = int(self.decode_type)
+    if not 0 <= self.version <= 255:
+      raise ValueError(f"version must be 0-255, got {self.version}")
```
veblush: I think this validation logic could be in `__post_init__`.

tensorflow/lite/micro/compression/decode_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Bytes 2-3: reserved (should be zero)
+    self.assertEqual(result[2], 0)
+    self.assertEqual(result[3], 0)
+
+    # Bytes 4-15: user_data (default all zeros)
+    self.assertEqual(result[4:16], b'\x00' * 12)
+
+  def testCustomVersion(self):
+    dcm = decode.DecodeCommonMetadata(decode_type=1, version=2)
```
veblush: decode_type=1 ? maybe decode_type=decode.DecodeType(1) ?

veblush (commented):
Thanks for the PR. It just has a couple of minor comments.

tensorflow/lite/micro/compression/decode.py
rkuester: Good thought. I decided to remove it instead. to_bytes packs the version into a bytearray (result[1] = self.version), and a bytearray slot only accepts an int in 0..255, so an out of range version already raises a ValueError right there. That made the explicit check redundant. We could add a __post_init__ to get an earlier error, but it would only catch bad values at construction, not later assignments to the field. It is Python after all, so it's hard to be exhaustively type-safe here.

tensorflow/lite/micro/compression/decode_test.py
rkuester: Good catch. In fact, I switched all the bare integer decode_type values in the tests to named constants (DecodeType.LUT / HUFFMAN), not just this one.

veblush (approved):

PR #3590, opened 2026-06-06
feat(compression): add Compressor protocol

rkuester:
Define the plugin interface for compression methods. Each compressor
implements the Compressor protocol with a compress() method that returns
encoded data and ancillary data. This is the interface the per-method
plugins (LUT, Huffman, Pruning) will implement in following slices.

BUG=part of #3256


rkuester:
@veblush, I can't seem to get a clean CI run on this due to what look like CI infrastructure issues. Would you please take a look?

veblush (approved):

PR #3593, opened 2026-06-12
Reject out-of-range GATHER index values at runtime to prevent OOB read

evilgensec:
BUG=GATHER reference kernel reads out of bounds on out-of-range index values (memory-safety fix)

## Problem

`GatherEval` (`tensorflow/lite/micro/kernels/gather.cc`) indexes the input tensor with the values from the `positions` tensor. Those values are only checked with `TFLITE_DCHECK_GE`/`TFLITE_DCHECK_LT`, which expand to a no-op under `-DNDEBUG` — i.e. in the `release` / `release_with_logs` builds used on device. `GatherPrepare` validates `axis` and `batch_dims`, but never the index values themselves.

When a GATHER op's `positions` are a runtime input (e.g. an embedding lookup whose index comes from the application's input), a negative or too-large index makes the kernel read outside the input tensor — a heap-buffer-overflow read whose bytes are copied into the output. The full TFLite GATHER kernel guards index values at runtime (`TF_LITE_ENSURE(context, indices_has_only_positive_elements)` plus the reference bounds handling) and returns `kTfLiteError`; the Micro kernel does not.

## Reproduction

`release_with_logs` build (so `-DNDEBUG` strips the DCHECKs) with AddressSanitizer; a benign GATHER over a 3x4 table with one runtime index value of `3` (axis size is 3):

```
ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 16
    #1 tflite::(anonymous namespace)::GatherEval(...) gather.cc:204
    #2 tflite::micro::KernelRunner::Invoke() kernel_runner.cc:111
0x...  is located 0 bytes after 48-byte region   <- the embedding table
```

A negative index (`-1`) reads before the table.

## Fix

Validate the index values in `GatherEval` and return `kTfLiteError` for any index outside `[0, axis_size)` before dispatching to the reference kernel. Adds regression tests `GatherOp_IndexGreaterThanAxisSizeFailsClosed` and `GatherOp_NegativeIndexFailsClosed`. With the fix the out-of-range index returns `kTfLiteError` with no out-of-bounds access; `kernel_gather_test` passes 20/20.

### DIFF
--- a/tensorflow/lite/micro/kernels/gather.cc
+++ b/tensorflow/lite/micro/kernels/gather.cc
@@ -199,6 +199,28 @@ TfLiteStatus GatherEval(TfLiteContext* context, TfLiteNode* node) {
       tflite::micro::GetEvalOutput(context, node, kOutputTensor);
 
   if (coords->type == kTfLiteInt32) {
+    // The reference Gather() below indexes `input` with the values in `coords`
+    // and only guards them with TFLITE_DCHECK, which is compiled out in release
+    // (-DNDEBUG) builds. Validate the index values at runtime so that an
+    // out-of-range index supplied as a model's runtime input fails closed with
+    // kTfLiteError instead of reading out of bounds. This mirrors the runtime
+    // index validation performed by the full TFLite GATHER kernel.
+    const TfLiteIntArray* input_dims = input->dims;
+    int axis = params->axis;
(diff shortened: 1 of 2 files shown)


veblush:

veblush:
Closed in favor of #3606

PR #3595, opened 2026-06-13
Validate tensor rank in CIRCULAR_BUFFER Prepare to prevent out-of-bounds read

BiswajeetRay7:
### Problem
CircularBufferPrepare in circular_buffer_common.cc reads input/output dims->data[0] through dims->data[3] without validating the tensors are rank-4. TfLiteIntArray stores dims as a flexible-array member (int data[]) sized to exactly `size` ints, so on a rank<4 tensor, dims->data[3] reads past the end — an out-of-bounds read. The same assumption is in the cycles_max <= 0 branch and the eval path in circular_buffer.cc.

Sibling kernels already guard this way (depth_to_space.cc, batch_to_space_nd.cc, broadcast_to.cc, concatenation.cc, cumsum.cc) via NumDimensions(). This brings CIRCULAR_BUFFER in line.

### Fix
Add NumDimensions(input)==4 and NumDimensions(output)==4 checks before any dims->data[..] access. No new include needed (kernel_util.h already included).

### Refs
Reported via GitHub Security Advisory GHSA-3x72-x298-9pjx and Google OSS VRP issue 523561915.

BUG=523561915

Signed-off-by: Biswajeet Ray <raybiswajeet2@gmail.com>

### DIFF
--- a/tensorflow/lite/micro/kernels/circular_buffer_common.cc
+++ b/tensorflow/lite/micro/kernels/circular_buffer_common.cc
@@ -49,6 +49,10 @@ TfLiteStatus CircularBufferPrepare(TfLiteContext* context, TfLiteNode* node) {
 
   TF_LITE_ENSURE(context, input != nullptr);
   TF_LITE_ENSURE(context, output != nullptr);
+  // The kernel indexes dims->data[0..3] below; ensure both tensors are rank-4
+  // before any access to avoid an out-of-bounds read on malformed models.
+  TF_LITE_ENSURE_EQ(context, NumDimensions(input), 4);
+  TF_LITE_ENSURE_EQ(context, NumDimensions(output), 4);
   TF_LITE_ENSURE_EQ(context, input->dims->data[0], output->dims->data[0]);
   TF_LITE_ENSURE_EQ(context, 1, input->dims->data[1]);
   TF_LITE_ENSURE_EQ(context, input->dims->data[2], output->dims->data[2]);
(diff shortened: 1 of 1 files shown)


BiswajeetRay7:
Gentle ping — this is a small (4-line) defensive fix bringing CIRCULAR_BUFFER in line with the rank validation already used by sibling kernels (depth_to_space, concatenation, etc.). All checks are green. Would a code owner be able to take a look when convenient? Happy to add a regression test if that would help. Thanks!

veblush:

BiswajeetRay7:
Thanks for the pointer to the guide. Rank mismatch here is an "Invalid Model Topology," so per the guide I've kept the validation in Prepare, and combined the two rank checks into a single TF_LITE_ENSURE to avoid the extra ROM cost of a second logged condition. Let me know if you'd prefer any further changes.

BiswajeetRay7:
Hi @veblush  — gentle follow-up. The Error Handling Guide changes you suggested were applied ~40 days ago (combined into a single `TF_LITE_ENSURE`), all checks are green, and the PR is currently blocked only on code-owner review from `tensorflow/micro`.

This is also the fix tracked by GHSA-3x72-x298-9pjx and Google OSS VRP issue 523561915, both of which are waiting on merge to close. If you or another `tensorflow/micro` code owner could take a look when convenient, that would help move all three tracks forward. Happy to add a regression test if that would be preferred.

Thanks!

PR #3597, opened 2026-06-15
Fix missing quantizedBiasType setting for 16x8 requantization

veblush:
When a model is compiled without explicit bias tensors, `requantize_flatbuffer_utils` previously skipped the bias upgrade logic entirely. As a result, the `quantizedBiasType` hint in the operator's options was not updated. This caused TFLM to fall back to a 32-bit accumulator for these layers, which overflows.

This change explicitly injects `op.builtinOptions.quantizedBiasType` = int64 directly into the operator options for supported operators, fixing precision loss for bias-less layers running 16x8 kernels.

BUG=n/a


veblush:
cc: @rameshkunasi

suleshahid (approved):

PR #3413, opened 2026-02-09
[Fix] Batch MatMul Op

veblush:
Batmul
* Disabled optimization when adj_x=true.
* Extracted logic into `ReshapeToFlattenBatchDimsIfPossible` for better maintainability.
* Added two new tests.

BUG=https://github.com/tensorflow/tflite-micro/issues/3412

Issue #3412, opened 2026-02-08
Increase correctness testing for BATCH_MATMUL dimensionality reduction


ddavis-2015:
@tensorflow/micro 

The existing BATCH_MATMUL dimensionality reduction code does not fully test the correctness of that code.  Specifically there are no tests for a LHS tensor with shape [..., N, A, B] where N > 1, and a RHS tensor with shape [..., 1, B, C].  Current tests only check for N = 1.  This means that any change to the code involving the row/column transpose of these tensors may pass successfully, even though results for N > 1 would be incorrect.

Existing rank 3 tensor unit tests for BATCH_MATMUL should have the LHS tensor shape batch size be N > 1 for each of the four possible adjunct parameter settings.

PR #3599, opened 2026-06-16
Re-enable 3D x 2D BatchMatMul dimension collapse for adj_x=true

veblush:
Re-enables the outer dimension collapse optimization (`[B, C, K] -> [B*C, K]`) for 3D x 2D BatchMatMul when `adj_x = true`, removing the workaround from PR #3413.
  
By executing dimension collapse **after** shape assignment, `lhs_shape` is in the correct canonical form to safely collapse `[Batch, Col]` into `[Batch * Col]` across both ReferenceOps and CMSIS-NN. Added unit tests to verify and protect this optimization.

BUG=https://github.com/tensorflow/tflite-micro/issues/3412

### DIFF
--- a/tensorflow/lite/micro/kernels/batch_matmul.cc
+++ b/tensorflow/lite/micro/kernels/batch_matmul.cc
@@ -340,37 +340,36 @@ TfLiteStatus EvalInt16(TfLiteContext* context, const OpDataBatchMatmul& data,
   return kTfLiteOk;
 }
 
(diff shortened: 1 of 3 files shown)


ddavis-2015 (approved):
looks good to me.

rameshkunasi:
@suleshahid can you please review this PR

suleshahid (approved):

PR #3594, opened 2026-06-12
Fix: Add parameter magnitude validation in signal kernel Prepare functions

foodlook:
## Summary

Add `TF_LITE_ENSURE` magnitude checks in 6 signal kernel `Prepare` functions to prevent out-of-bounds memory access and divide-by-zero from crafted model parameters.

## Problem

Each signal kernel's `Prepare` validates tensor dimension **count** and **type** but never the dimension **values** or the relationship between op flexbuffer parameters and buffer sizes. A crafted `.tflite` model can set e.g. `frame_step > frame_size`, `input_length > fft_length`, or `n_frames = 0`, triggering heap-buffer-overflow or divide-by-zero in `Eval`.

All 7 sinks are dynamically proven with ASan/UBSan crashes on the real `signal/src/` functions.

## Changes

| File | Added Checks |
|------|-------------|
| `overlap_add.cc` | `frame_size>0`, `n_frames>0`, `frame_step>0`, `frame_step<=frame_size` |
| `rfft.cc` | `fft_length>0`, `input_length>0`, `input_length<=fft_length` |
| `energy.cc` | `start_index>=0`, `end_index>start_index`, `end_index<=input_elems` |
| `window.cc` | `weight_size>0`, `input_size>0`, `input_size%weight_size==0`, input==output size |
| `filter_bank_spectral_subtraction.cc` | `num_channels>0`, matches input/output/noise dims |
| `pcan.cc` | `input_elems>0`, matches noise_estimate/output dims |

## References

- Google Issue Tracker: [#523091874](https://issuetracker.google.com/issues/523091874)
- ASan crash evidence: 7/7 sinks reproduced with libFuzzer + AddressSanitizer

## Testing

The fix adds early-return `kTfLiteError` for invalid parameters that previously caused undefined behavior. Existing tests should pass unchanged (they use valid parameters).

PR #3602, opened 2026-06-18
Validate num_channels against tensor sizes in FILTER_BANK kernel

adilburaksen:
BUG=out-of-bounds read/write in FILTER_BANK kernel from unvalidated num_channels

### What

The `FILTER_BANK` signal kernel takes its channel count from the op's init flexbuffer
(`num_channels`) and six runtime tensors. `FilterBankInit` reads `num_channels` and allocates a
work area of `num_channels + 1`; `FilterbankAccumulateChannels` then loops `num_channels + 1` times,
reading `channel_frequency_starts[i]`, `channel_weight_starts[i]` and `channel_widths[i]` and writing
`num_channels` output elements.

`FilterBankPrepare` validated only the rank and type of each tensor — never that `num_channels` is
consistent with the per-channel metadata tensor lengths or the output length. A model whose flexbuffer
declares a `num_channels` larger than those tensors therefore caused an out-of-bounds read.

This was confirmed with AddressSanitizer two ways:
- a kernel unit test (32-channel flexbuffer with 17-element metadata tensors), and
- loading a real `audio_preprocessor_int8.tflite` (with a single byte changed so `num_channels = 80`,
  metadata tensors left at 41) through `GetModel()` + `MicroInterpreter` + `Invoke()`, which reads off
  the end of the model buffer inside `FilterbankAccumulateChannels`.

This is the same class of bug, and the same `num_channels`-vs-tensor-size check, that was added for the
sibling `FilterBankSpectralSubtraction` kernel in #3594; `filter_bank.cc` was not covered there.

### Change

- `filter_bank.cc`: `FilterBankPrepare` now checks `num_channels > 0`, that
  `channel_frequency_starts`, `channel_weight_starts` and `channel_widths` each have
  `num_channels + 1` elements, and that the output has `num_channels` elements.
- `filter_bank_test.cc`: regression test that the mismatched model is rejected (no OOB).

No change for valid models.

### DIFF
--- a/signal/micro/kernels/filter_bank.cc
+++ b/signal/micro/kernels/filter_bank.cc
@@ -61,6 +61,16 @@ void* FilterBankInit(TfLiteContext* context, const char* buffer,
                                 length);
   params->config.num_channels = fbw.ElementAsInt32(kNumChannelsIndex);
 
+  // Reject a num_channels that is non-positive or large enough to overflow the
+  // work-area size computation below. Without this, a crafted model can wrap
+  // the multiplication (notably where size_t is 32 bits) into a tiny
+  // allocation, which FilterbankAccumulateChannels then writes past.
+  if (params->config.num_channels <= 0 ||
+      static_cast<size_t>(params->config.num_channels) >
+          SIZE_MAX / sizeof(uint64_t) - 1) {
+    return nullptr;
+  }
+
   params->work_area = static_cast<uint64_t*>(context->AllocatePersistentBuffer(
       context, (params->config.num_channels + 1) * sizeof(uint64_t)));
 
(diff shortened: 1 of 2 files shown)


veblush:

adilburaksen:
Thanks @veblush — I read the guide and pushed a follow-up that reframes this per §2 Phase 1: the `num_channels` consistency/overflow check now lives in `FilterBankPrepare` as a semantic model-topology validation (`TF_LITE_ENSURE`), rather than a bare `return nullptr` in `Init`, so an inconsistent topology is rejected during preparation with a clear message that `TF_LITE_STRIP_ERROR_STRINGS` can strip in production. PTAL.

PR #3603, opened 2026-06-18
Validate output size in FILTER_BANK_LOG kernel

adilburaksen:
BUG=out-of-bounds write in FILTER_BANK_LOG kernel from unchecked output size

### What

`FilterBankLogEval` takes `num_channels` from the **input** tensor
(`num_channels = input->dims->data[0]`) and `FilterbankLog` writes `num_channels`
elements to the output:

```cpp
for (int i = 0; i < num_channels; ++i) { output[i] = ...; }
```

`FilterBankLogPrepare` validated only the rank and type of the input and output
tensors — never that the output is at least as large as the input. A model whose
`FILTER_BANK_LOG` output tensor is smaller than its input therefore causes an
out-of-bounds **write** (e.g. input of 32 elements, output of 1), confirmed with
AddressSanitizer:

```
ERROR: AddressSanitizer: stack-buffer-overflow  WRITE of size 2
  #0 tflite::tflm_signal::FilterbankLog(...)  signal/src/filter_bank_log.cc
  #1 tflite::FilterBankLogEval(...)           signal/micro/kernels/filter_bank_log.cc:97
  #2 tflite::micro::KernelRunner::Invoke()
```

### Change

- `filter_bank_log.cc`: `FilterBankLogPrepare` now requires
  `ElementCount(output) >= ElementCount(input)`.
- `filter_bank_log_test.cc`: regression test that the mismatched model is rejected.

No change for valid models.

### DIFF
--- a/signal/micro/kernels/filter_bank_log.cc
+++ b/signal/micro/kernels/filter_bank_log.cc
@@ -78,6 +78,12 @@ TfLiteStatus FilterBankLogPrepare(TfLiteContext* context, TfLiteNode* node) {
   TF_LITE_ENSURE_TYPES_EQ(context, input->type, kTfLiteUInt32);
   TF_LITE_ENSURE_TYPES_EQ(context, output->type, kTfLiteInt16);
 
+  // FilterbankLog writes ElementCount(input) (num_channels, taken from the
+  // input tensor) elements to the output, so the output must be at least as
+  // large or it is written out of bounds.
+  TF_LITE_ENSURE(context,
+                 ElementCount(*output->dims) >= ElementCount(*input->dims));
+
   micro_context->DeallocateTempTfLiteTensor(input);
   micro_context->DeallocateTempTfLiteTensor(output);
   return kTfLiteOk;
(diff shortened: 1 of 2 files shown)


veblush:

adilburaksen:
Thanks @veblush — read the guide and reframed accordingly: the output-vs-input size check is a semantic model-topology validation done in `FilterBankLogPrepare` with `TF_LITE_ENSURE` (§2 Phase 1), so an inconsistent output shape is rejected during preparation rather than in `Eval`. PTAL.

adilburaksen:
Commenting to keep this out of the stale bot's path — it announced closure on 2026-08-17 and the PR is somehow still open.

Status: @veblush asked me to follow the Error Handling & Defensive Programming Guide on 2026-06-30; I reworked it the next day so the output-vs-input size check is a Phase 1 `TF_LITE_ENSURE` in `FilterBankLogPrepare` rather than a runtime check in `Eval`. There has been no review since.

@rkuester — you're the one actively landing changes here; could you take a look, or reassign it? One note on the checks: `zizmor-scan` and `zizmor-config` pass and only `zizmor-output` fails, which is the SARIF-upload step that can't run from a fork, and `approval-gate` is skipped so the platform test jobs never started. Nothing there is a real failure, but it needs a maintainer to approve the workflows.

rkuester:
Hey @adilburaksen, I think I'm the wrong person for this, so I'm kicking it back to @veblush.

PR #3604, opened 2026-06-19
Validate STACKER init parameters against overflow and tensor sizes

adilburaksen:
BUG=out-of-bounds access in STACKER kernel from unvalidated init parameters

### What

`StackerInit` reads `num_channels`, `stacker_left_context`, `stacker_right_context`
and `stacker_step` from the init flexbuffer and computes:

```cpp
buffer_size = num_channels * (stacker_left_context + stacker_right_context + 1);
step_size   = num_channels * stacker_step;
```

`StackerPrepare` validated only tensor ranks/types — never these parameters. A
crafted model (e.g. `stacker_right_context = -1`, making `buffer_size = 0`, or
values whose product overflows) produces a zero/undersized circular buffer.
`StackerEval` then calls `CircularBufferWrite` / `CircularBufferExtend`, copying
`num_channels` values per frame and reading `buffer_size` values into the output,
out of bounds.

In release builds the `ASSERT(...)` in `CircularBufferWrite` (`#define ASSERT
assert`) is compiled out under `-DNDEBUG`, so this is an out-of-bounds
read/write. Confirmed with AddressSanitizer (built with `-DNDEBUG`):

```
ERROR: AddressSanitizer: global-buffer-overflow in __asan_memcpy
  #1 CircularBufferWrite   signal/src/circular_buffer.cc:97
  #2 CircularBufferExtend  signal/src/circular_buffer.cc:178
  #3 StackerEval           signal/micro/kernels/stacker.cc:139
```

### Change

- `stacker.cc`: `StackerPrepare` validates `num_channels > 0`, contexts `>= 0`,
  `stacker_step > 0`, `num_channels <= input size`, that the `buffer_size` product
  does not overflow, and that the output is large enough.
- `stacker_test.cc`: regression test that the malformed parameters are rejected.

No change for valid models.

### DIFF
--- a/signal/micro/kernels/stacker.cc
+++ b/signal/micro/kernels/stacker.cc
@@ -112,9 +112,36 @@ TfLiteStatus StackerPrepare(TfLiteContext* context, TfLiteNode* node) {
   TF_LITE_ENSURE_TYPES_EQ(context, output->type, kTfLiteInt16);
   TF_LITE_ENSURE_TYPES_EQ(context, output_valid->type, kTfLiteBool);
 
+  const int input_size = ElementCount(*input->dims);
+  const int output_size = ElementCount(*output->dims);
+
   micro_context->DeallocateTempTfLiteTensor(input);
   micro_context->DeallocateTempTfLiteTensor(output);
   micro_context->DeallocateTempTfLiteTensor(output_valid);
+
+  // Validate the init-flexbuffer parameters before they are used to size the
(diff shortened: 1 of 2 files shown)


veblush:

adilburaksen:
Thanks @veblush — read the guide and reframed accordingly: the STACKER init-parameter and overflow validation now sits in `Prepare` as semantic model-topology checks (`TF_LITE_ENSURE`, §2 Phase 1), keeping `Eval` free of redundant checks. PTAL.

PR #3614, opened 2026-07-01
feat(python): add alt decompression memory parameter to interpreter

rkuester:
Add an alt_decompression_memory_size parameter to the Python interpreter
API. When non-zero, the interpreter allocates a separate memory region
for DECODE operator outputs and calls SetDecompressionMemory before
AllocateTensors, keeping decompressed tensors out of the main arena.

Python and interpreter-wrapper only; additive with a default of 0 (no
behavior change when unset).

Part of the DECODE compression-tooling stack.

BUG=#3256

### DIFF
--- a/python/tflite_micro/interpreter_wrapper.h
+++ b/python/tflite_micro/interpreter_wrapper.h
@@ -19,6 +19,7 @@ limitations under the License.
 
 #include "python/tflite_micro/python_ops_resolver.h"
 #include "tensorflow/lite/micro/micro_allocator.h"
+#include "tensorflow/lite/micro/micro_context.h"
 #include "tensorflow/lite/micro/micro_interpreter.h"
 #include "tensorflow/lite/micro/recording_micro_allocator.h"
 
@@ -40,7 +41,8 @@ class InterpreterWrapper {
   InterpreterWrapper(
       PyObject* model_data, const std::vector<std::string>& registerers_by_name,
       size_t arena_size, int num_resource_variables,
-      InterpreterConfig config = InterpreterConfig::kAllocationRecording);
+      InterpreterConfig config = InterpreterConfig::kAllocationRecording,
+      size_t alt_decompression_memory_size = 0);
   ~InterpreterWrapper();
 
   void PrintAllocations();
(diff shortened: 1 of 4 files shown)


python/tflite_micro/python_ops_resolver.cc
```diff
@@ -40,7 +40,9 @@ PythonOpsResolver::PythonOpsResolver() {
   AddConv2D();
   AddCos();
   AddCumSum();
+#ifdef USE_TFLM_COMPRESSION
```
veblush: Do we use this flag for Decode? I thought this was only for legacy compression that new Decode will replace?

python/tflite_micro/interpreter_wrapper.h
```diff
@@ -40,7 +41,8 @@ class InterpreterWrapper {
   InterpreterWrapper(
       PyObject* model_data, const std::vector<std::string>& registerers_by_name,
       size_t arena_size, int num_resource_variables,
-      InterpreterConfig config = InterpreterConfig::kAllocationRecording);
+      InterpreterConfig config = InterpreterConfig::kAllocationRecording,
+      size_t alt_decompression_memory_size = 0);
```
veblush, line 45: And maybe we should have a test for using this new alt_decompression_memory_size?

python/tflite_micro/python_ops_resolver.cc
rkuester: Agreed, this does not belong. Removed.

python/tflite_micro/interpreter_wrapper.h
rkuester: Since this is merely a Python wrapper passing this information to the underlying C++ interpreter, testing it without a comprehensive integration test is challenging. Even then, verifying that the alternate memory is being utilized might be a bit tricky. My suggestion is we review the wrapper by inspection to ensure it correctly passes this argument to the C++ code, and then rely on the existing C++ unit test for the underlying feature.

veblush: Sounds good.

veblush (approved):

PR #3624, opened 2026-07-06
feat(compression): add DECODE operator insertion

rkuester:
Insert DECODE operators before consumers of compressed tensors. Each
consumer gets its own DECODE operator to support alternate decompression
memory, which resets allocations between DECODE invocations.

After insertion, compressed tensors are rewritten to hold encoded data
as UINT8 with shape matching byte count.

Compression tooling only; adds decode_insert.py and its test, no
runtime changes.

Part of the DECODE compression-tooling stack.

BUG=#3256

### DIFF
--- a/tensorflow/lite/micro/compression/decode_insert.py
+++ b/tensorflow/lite/micro/compression/decode_insert.py
@@ -263,22 +263,27 @@ def insert_decode_operators(
       )
       return op, outputs
 
+    # Positions of the original operators, computed once so the sort and
+    # insertions below avoid a linear scan per lookup.
+    op_position = {op: i for i, op in enumerate(subgraph.operators)}
+
     # Group compressed tensors by consumer, then handle consumers in
-    # reverse position order so insertions don't invalidate positions
+    # reverse position order so insertions don't invalidate positions:
+    # each insertion falls after every consumer still to be processed,
+    # leaving the recorded positions valid.
     by_consumer: dict[model_editor.Operator, list[_CompressedTensorInfo]] = {}
     for info in tensor_infos:
       for consumer in info.consumers:
         by_consumer.setdefault(consumer, []).append(info)
 
     for consumer in sorted(by_consumer,
-                           key=subgraph.operators.index,
+                           key=lambda op: op_position[op],
                            reverse=True):
       infos = by_consumer[consumer]
       decode_op, decoded_tensors = build_decode(infos)
 
       # Insert DECODE immediately before this consumer
-      insert_pos = subgraph.operators.index(consumer)
-      subgraph.operators.insert(insert_pos, decode_op)
+      subgraph.operators.insert(op_position[consumer], decode_op)
 
       # Rewire only this consumer to use the decoded outputs
       for info, decoded in zip(infos, decoded_tensors):
--- a/tensorflow/lite/micro/compression/decode_insert.py
+++ b/tensorflow/lite/micro/compression/decode_insert.py
@@ -1,0 +1,267 @@
+# Copyright 2026 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""DECODE operator insertion into TFLite model graphs.
+
+This module inserts DECODE operators into a compressed model. DECODE operators
+transform encoded tensors (with their paired ancillary data tensors) into
+tensors ready for use by downstream operators.
+
+The DECODE operator is registered as a custom operator named "TFLM_DECODE".
+Each DECODE output requires two inputs: the encoded tensor and the ancillary
+data tensor (containing the DCM header and decode-type-specific data).
+"""
+
+import warnings
+from collections import defaultdict
+from dataclasses import dataclass
+
+from tflite_micro.tensorflow.lite.micro.compression import compressor
+from tflite_micro.tensorflow.lite.micro.compression import model_editor
+from tflite_micro.tensorflow.lite.python import schema_py_generated as tflite
+
+# Custom operator name for DECODE
+DECODE_CUSTOM_OP_NAME = "TFLM_DECODE"
+
+
+@dataclass
+class _CompressedTensorInfo:
+  """Information about a compressed tensor for DECODE insertion."""
+  subgraph_idx: int
+  tensor_idx: int
+  tensor: model_editor.Tensor
+  encoded_data: bytes
+  ancillary_data: bytes
+  consumers: list[model_editor.Operator]
+
+
+def _find_tensor_consumers(
+    subgraph: model_editor.Subgraph,
+    tensor: model_editor.Tensor,
+) -> list[model_editor.Operator]:
+  """Find all operators in subgraph that use tensor as an input."""
+  consumers = []
+  for op in subgraph.operators:
+    if tensor in op.inputs:
+      consumers.append(op)
+  return consumers
+
+
+def _create_ancillary_tensor(
+    ancillary_data: bytes,
+    original_tensor: model_editor.Tensor,
+) -> model_editor.Tensor:
+  """Create an ancillary data tensor for a compressed tensor.
+
+  Args:
+    ancillary_data: The complete ancillary data (DCM + type-specific data).
+    original_tensor: The original tensor being decoded, for naming.
+
+  Returns:
+    A new Tensor containing the ancillary data.
+  """
+  name = None
+  if original_tensor.name:
+    name = f"{original_tensor.name}_ancillary"
+
+  return model_editor.Tensor(
+      shape=(len(ancillary_data), ),
+      dtype=tflite.TensorType.UINT8,
+      data=ancillary_data,
+      name=name,
+  )
+
+
+def _create_output_tensor(
+    original_tensor: model_editor.Tensor, ) -> model_editor.Tensor:
+  """Create the output tensor for a DECODE operator.
+
+  The output tensor has the same shape, dtype, and quantization as the
+  original tensor would have when decoded. It has no data---the DECODE
+  operator produces its values at runtime.
+
+  Args:
+    original_tensor: The original tensor being decoded.
+
+  Returns:
+    A new Tensor for the DECODE output.
+  """
+  name = None
+  if original_tensor.name:
+    name = f"{original_tensor.name}_decoded"
+
+  return model_editor.Tensor(
+      shape=original_tensor.shape,
+      dtype=original_tensor.dtype,
+      quantization=original_tensor.quantization,
+      name=name,
+  )
+
+
+def _rewire_consumers(
+    consumers: list[model_editor.Operator],
+    old_tensor: model_editor.Tensor,
+    new_tensor: model_editor.Tensor,
+) -> None:
+  """Replace old_tensor with new_tensor in all consumer inputs."""
+  for consumer in consumers:
+    consumer.inputs = [
+        new_tensor if t is old_tensor else t for t in consumer.inputs
+    ]
+
+
+def _rewrite_encoded_tensor(
+    tensor: model_editor.Tensor,
+    encoded_data: bytes,
+) -> None:
+  """Rewrite a compressed tensor to hold encoded data.
+
+  The original tensor contained uncompressed values with quantization. After
+  compression, it holds packed indices (or other encoded form) as raw bytes.
+  This function updates the tensor in place to reflect its new role.
+
+  Args:
+    tensor: The tensor to rewrite.
+    encoded_data: The compressed/encoded data bytes.
+  """
+  tensor.shape = (len(encoded_data), )
+  tensor.dtype = tflite.TensorType.UINT8
+  tensor.quantization = None
+  tensor.buffer.data = encoded_data
+
+
+def insert_decode_operators(
+    model: model_editor.Model,
+    compression_results: dict[tuple[int, int], compressor.CompressionResult],
+) -> None:
+  """Insert DECODE operators for all compressed tensors.
+
+  This function modifies the model in-place, inserting DECODE operators
+  before any operator that uses a compressed tensor as input.
+
+  A separate DECODE is inserted before each consumer, rather than sharing one
+  DECODE output among all consumers. This is required because the interpreter's
+  alternate decompression memory resets its allocation offset for each DECODE's
+  Prepare, causing all DECODE outputs to be allocated at the same address. If
+  two consumers share one DECODE and another DECODE runs between them, the
+  intervening DECODE overwrites the shared output, corrupting data for the
+  second consumer.
+
+  For each consumer of a compressed tensor:
+  1. Create an ancillary data tensor containing DCM + type-specific data
+  2. Create an output tensor with the same shape/dtype as the decoded tensor
+  3. Insert a DECODE operator immediately before the consumer
+  4. Rewire the consumer to use the DECODE output
+
+  Args:
+    model: The model to modify in-place.
+    compression_results: Map from (subgraph_idx, tensor_idx) to the
+                         CompressionResult containing ancillary_data.
+  """
+  # Group compressed tensors by subgraph
+  by_subgraph: dict[int, list[_CompressedTensorInfo]] = defaultdict(list)
+
+  for (sg_idx, tensor_idx), result in compression_results.items():
+    subgraph = model.subgraphs[sg_idx]
+    tensor = subgraph.tensors[tensor_idx]
+    consumers = _find_tensor_consumers(subgraph, tensor)
+
+    if not consumers:
+      # Check if tensor is a subgraph output
+      is_output = tensor in subgraph.outputs
+      if is_output:
+        # TODO: Handle compressed tensors that are subgraph outputs.
+        # This occurs in multi-subgraph models using IF/WHILE where a
+        # compressed tensor flows out of a subgraph.
+        raise NotImplementedError(
+            f"Compressed tensor {tensor.name!r} (subgraph {sg_idx}, "
+            f"tensor {tensor_idx}) is a subgraph output with no consumers. "
+            "Compressed subgraph outputs are not yet supported.")
+      else:
+        warnings.warn(
+            f"Compressed tensor {tensor.name!r} (subgraph {sg_idx}, "
+            f"tensor {tensor_idx}) has no consumers and is not a subgraph "
+            "output. No DECODE operator will be inserted.",
+            stacklevel=2)
+      continue
+
+    info = _CompressedTensorInfo(
+        subgraph_idx=sg_idx,
+        tensor_idx=tensor_idx,
+        tensor=tensor,
+        encoded_data=result.encoded_data,
+        ancillary_data=result.ancillary_data,
+        consumers=consumers,
+    )
+    by_subgraph[sg_idx].append(info)
+
+  # Process each subgraph
+  for sg_idx, tensor_infos in by_subgraph.items():
+    subgraph = model.subgraphs[sg_idx]
+
+    # Collect all (consumer, tensor_info) pairs and sort by consumer position
+    # in reverse order so insertions don't invalidate positions
+    consumer_pairs = []
+    for info in tensor_infos:
+      for consumer in info.consumers:
+        consumer_pairs.append((consumer, info))
+
+    consumer_pairs.sort(
+        key=lambda pair: subgraph.operators.index(pair[0]),
+        reverse=True,
+    )
+
+    # Cache ancillary tensors by original tensor to avoid duplicates. Each
+    # DECODE needs its own output tensor, but ancillary data is identical for
+    # all DECODEs of the same compressed tensor.
+    ancillary_cache: dict[model_editor.Tensor, model_editor.Tensor] = {}
+
+    # Track tensors to rewrite after all output tensors are created, since
+    # _create_output_tensor reads the original tensor's shape/dtype/quantization.
+    tensors_to_rewrite: dict[model_editor.Tensor, bytes] = {}
+
+    for consumer, info in consumer_pairs:
+      # Reuse or create ancillary data tensor
+      if info.tensor not in ancillary_cache:
+        ancillary_tensor = _create_ancillary_tensor(
+            info.ancillary_data,
+            info.tensor,
+        )
+        subgraph.tensors.append(ancillary_tensor)
+        ancillary_cache[info.tensor] = ancillary_tensor
+        tensors_to_rewrite[info.tensor] = info.encoded_data
+      else:
+        ancillary_tensor = ancillary_cache[info.tensor]
+
+      # Create output tensor (one per DECODE)
+      output_tensor = _create_output_tensor(info.tensor)
+      subgraph.tensors.append(output_tensor)
+
+      # Create DECODE operator
+      decode_op = model_editor.Operator(
+          opcode=tflite.BuiltinOperator.CUSTOM,
+          custom_code=DECODE_CUSTOM_OP_NAME,
+          inputs=[info.tensor, ancillary_tensor],
+          outputs=[output_tensor],
+      )
+
+      # Insert DECODE immediately before this consumer
+      insert_pos = subgraph.operators.index(consumer)
+      subgraph.operators.insert(insert_pos, decode_op)
+
+      # Rewire only this consumer to use the decoded output
+      _rewire_consumers([consumer], info.tensor, output_tensor)
+
+    # Rewrite encoded tensors after all output tensors are created
+    for tensor, encoded_data in tensors_to_rewrite.items():
+      _rewrite_encoded_tensor(tensor, encoded_data)
--- a/tensorflow/lite/micro/compression/decode_insert.py
+++ b/tensorflow/lite/micro/compression/decode_insert.py
@@ -46,18 +46,6 @@ class _CompressedTensorInfo:
   is_output: bool
 
 
-def _find_tensor_consumers(
-    subgraph: model_editor.Subgraph,
-    tensor: model_editor.Tensor,
-) -> list[model_editor.Operator]:
-  """Find all operators in subgraph that use tensor as an input."""
-  consumers = []
-  for op in subgraph.operators:
-    if tensor in op.inputs:
-      consumers.append(op)
-  return consumers
-
-
 def _create_ancillary_tensor(
     ancillary_data: bytes,
     original_tensor: model_editor.Tensor,
@@ -240,7 +228,7 @@ def insert_decode_operators(
   for (sg_idx, tensor_idx), result in compression_results.items():
     subgraph = model.subgraphs[sg_idx]
     tensor = subgraph.tensors[tensor_idx]
-    consumers = _find_tensor_consumers(subgraph, tensor)
+    consumers = subgraph.consumers_of(tensor)
     is_output = tensor in subgraph.outputs
 
     if not consumers and not is_output:
--- a/tensorflow/lite/micro/compression/decode_insert.py
+++ b/tensorflow/lite/micro/compression/decode_insert.py
@@ -195,34 +195,26 @@ def insert_decode_operators(
   before any operator that uses a compressed tensor as input, and appending
   a DECODE for compressed tensors listed as subgraph outputs.
 
-  A separate DECODE is inserted before each consumer, rather than sharing one
-  DECODE output among all consumers. This is required because the interpreter's
-  alternate decompression memory resets its allocation offset for each DECODE's
-  Prepare, causing all DECODE outputs to be allocated at the same address. If
-  two consumers share one DECODE and another DECODE runs between them, the
-  intervening DECODE overwrites the shared output, corrupting data for the
-  second consumer.
-
-  Conversely, all compressed tensors read by one consumer are decoded by a
-  single DECODE. Values a consumer needs simultaneously must coexist, and
-  only the outputs of a single DECODE do: the allocation reset happens
-  between DECODE operators, not between the outputs of one.
+  A separate DECODE is inserted before each consumer, and one DECODE
+  decodes all the compressed tensors its consumer reads. DECODE outputs
+  are tensors with a lifetime limited to the very next operator in the
+  subgraph, so sharing one DECODE among multiple consumers would
+  violate the lifetime rule. The DECODE operator trades increased
+  latency for decreased memory usage.
 
   For each consumer of compressed tensors:
   1. Create an ancillary data tensor (DCM + type-specific data) for each
      compressed tensor the consumer reads
-  2. Create an output tensor with the same shape/dtype as each decoded
-     tensor
+  2. Create an output tensor as a copy of each original tensor
   3. Insert one DECODE operator immediately before the consumer
   4. Rewire the consumer to use the DECODE outputs
 
-  A subgraph's output list is treated as one more consumer, one which reads
-  its tensors only after the last operator runs: a calling operator (IF,
-  WHILE) copies subgraph outputs when the subgraph returns, and the client
-  reads model outputs after invocation. Compressed tensors in the output
-  list are therefore decoded by a single DECODE appended after the last
-  operator, and their output list entries are rewired to the decoded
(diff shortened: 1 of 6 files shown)


tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,268 @@
...
+
+  for (sg_idx, tensor_idx), result in compression_results.items():
+    subgraph = model.subgraphs[sg_idx]
+    tensor = subgraph.tensors[tensor_idx]
+    consumers = _find_tensor_consumers(subgraph, tensor)
+
+    if not consumers:
+      # Check if tensor is a subgraph output
+      is_output = tensor in subgraph.outputs
+      if is_output:
```
veblush: I guess `if it_output` should be outside of `if not consumers`? It might be consumed and an oiutput of subgraph?

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,268 @@
...
+    subgraph = model.subgraphs[sg_idx]
+
+    # Collect all (consumer, tensor_info) pairs and sort by consumer position
+    # in reverse order so insertions don't invalidate positions
+    consumer_pairs = []
+    for info in tensor_infos:
+      for consumer in info.consumers:
+        consumer_pairs.append((consumer, info))
+
+    consumer_pairs.sort(
```
veblush: This could be slow because of subgraph.operators.index in the sort?

tensorflow/lite/micro/compression/decode_insert.py
rkuester: You're right: the check missed tensors that are both consumed by an operator and listed as a subgraph output, so a subgraph would have output the encoded bytes instead of the decoded values. Addressed in 39998895 by treating the output list as one more consumer, fed by a DECODE appended after the last operator.

f3a8e564 pulls forward a fix from later in the series: separate DECODEs per tensor don't work with alternate decompression memory when the decoded values are needed at the same time, and with output handling added, it's even more worth getting this right initially than rewriting it later. It's a separate commit for easy review, relying on the squash when the PR merges.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Addressed in 50f2fcf4: a position map built once per subgraph replaces the scans in the sort and insertions.

tensorflow/lite/micro/compression/decode_insert.py
ddavis-2015: @rkuester @veblush Re: Passing the output of DECODE to or from a subgraph

The design document will be updated to address what to do in this situation.

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+class TestHelperFunctions(unittest.TestCase):
+  """Tests for internal helper functions."""
+
+  def test_find_tensor_consumers(self):
+    """_find_tensor_consumers finds all ops using a tensor."""
+    model = _build_shared_weights_model()
+    sg = model.subgraphs[0]
+    weights = sg.tensor_by_name("shared_weights")
+
+    consumers = decode_insert._find_tensor_consumers(sg, weights)
```
ddavis-2015: Use of a private method (_find_tensor_consumers).  Maybe have a set of public methods just for use by unit tests?

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+      model_editor.Subgraph(
+          tensors=[weights],
+          operators=[
+              model_editor.Operator(
+                  opcode=tflite.BuiltinOperator.FULLY_CONNECTED,
+                  inputs=[input_t, weights],
+                  outputs=[output_t],
+              )
+          ],
+      )
```
ddavis-2015, line 72: Doesn't the Subgraph inputs/outputs need to be specified?  Otherwise the subgraph compiler will just assign empty vectors to the tflite.SubGraphT.

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+          operators=[
+              model_editor.Operator(
+                  opcode=tflite.BuiltinOperator.FULLY_CONNECTED,
+                  inputs=[input_t, weights],
+                  outputs=[output_t],
+              )
+          ],
+          outputs=[output_t, table],
+      )
+  ])
```
ddavis-2015, line 206: Doesn't the Subgraph inputs need to be specified? Otherwise the subgraph compiler will just assign an empty vector in the tflite.SubGraphT.

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+    """DECODE operator inserted before FC that uses compressed weights."""
+    model = _build_simple_fc_model()
+    weights_tensor = model.subgraphs[0].tensor_by_name("weights")
+
+    # Create compression result
+    compression_results = {
+        (0, 0):
+        compressor.CompressionResult(
+            encoded_data=b'\x00\x00',
+            ancillary_data=_make_dummy_ancillary_data(),
```
ddavis-2015: mismatch between `encoded_data` and the `ancillary_data`.  Perhaps `_make_dummy_ancillary_data` should have arguments so that it can generate a DCM that matches the model data?  And `encoded_data` appears to be too short in length.

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+  return dcm.to_bytes() + value_tables
+
+
+class TestDecodeInsertion(unittest.TestCase):
+  """Tests for insert_decode_operators function."""
+
+  def test_insert_single_decode_operator(self):
+    """DECODE operator inserted before FC that uses compressed weights."""
+    model = _build_simple_fc_model()
+    weights_tensor = model.subgraphs[0].tensor_by_name("weights")
```
ddavis-2015: unused variable `weights_tensor`

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+  def test_decode_inputs_structure(self):
+    """DECODE operator has correct inputs: encoded tensor + ancillary."""
+    model = _build_simple_fc_model()
+    weights_tensor = model.subgraphs[0].tensor_by_name("weights")
+
+    compression_results = {
+        (0, 0):
+        compressor.CompressionResult(
+            encoded_data=b'\x00\x00',
+            ancillary_data=_make_dummy_ancillary_data(),
```
ddavis-2015: mismatch between encoded_data and the ancillary_data. Perhaps _make_dummy_ancillary_data should have arguments so that it can generate a DCM that matches the model data? And encoded_data appears to be too short in length.

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,306 @@
...
+  alternate decompression memory resets its allocation offset for each DECODE's
+  Prepare, causing all DECODE outputs to be allocated at the same address. If
+  two consumers share one DECODE and another DECODE runs between them, the
+  intervening DECODE overwrites the shared output, corrupting data for the
+  second consumer.
+
+  Conversely, all compressed tensors read by one consumer are decoded by a
+  single DECODE. Values a consumer needs simultaneously must coexist, and
+  only the outputs of a single DECODE do: the allocation reset happens
+  between DECODE operators, not between the outputs of one.
```
ddavis-2015: I would not provide an explanation of memory management.  Here you can just say that DECODE outputs are tensors with a lifetime limited to the very next operator in the subgraph.  The DECODE operator is designed to trade increased latency for decreased memory usage.

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,306 @@
...
+
+  For each consumer of compressed tensors:
+  1. Create an ancillary data tensor (DCM + type-specific data) for each
+     compressed tensor the consumer reads
+  2. Create an output tensor with the same shape/dtype as each decoded
+     tensor
```
ddavis-2015: how about "as each original consumer tensor"?

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,306 @@
...
+  3. Insert one DECODE operator immediately before the consumer
+  4. Rewire the consumer to use the DECODE outputs
+
+  A subgraph's output list is treated as one more consumer, one which reads
+  its tensors only after the last operator runs: a calling operator (IF,
+  WHILE) copies subgraph outputs when the subgraph returns, and the client
+  reads model outputs after invocation. Compressed tensors in the output
```
ddavis-2015: "and the client reads model outputs after invocation" is confusing.  Best to just remove this fragment.

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,306 @@
...
+    by_subgraph[sg_idx].append(info)
+
+  # Process each subgraph
+  for sg_idx, tensor_infos in by_subgraph.items():
+    subgraph = model.subgraphs[sg_idx]
+
+    # Cache ancillary tensors by original tensor to avoid duplicates. Each
+    # DECODE needs its own output tensor, but ancillary data is identical for
+    # all DECODEs of the same compressed tensor.
+    ancillary_cache: dict[model_editor.Tensor, model_editor.Tensor] = {}
```
ddavis-2015: This cache should be global across all subgraphs with lookup by Buffer, in order to handle the case where the original encoded tensor shares a buffer with another tensor in the same (or a different) subgraph.  Compression of the model cannot be maximized without handling buffer sharing.  Please see the scenario in the PR review comments.

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,306 @@
...
+  name = None
+  if original_tensor.name:
+    name = f"{original_tensor.name}_decoded"
+
+  return model_editor.Tensor(
+      shape=original_tensor.shape,
+      dtype=original_tensor.dtype,
+      quantization=original_tensor.quantization,
+      name=name,
+  )
```
ddavis-2015: Just making a new `Tensor` is insufficient here.  The output must be a complete copy of the original tensor (`TensorT`), with `Tensor.data` and `Tensor.buffer` set to `None`.  Updating the `name` is fine, so it doesnt name conflict with the original.

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,306 @@
...
+  name = None
+  if original_tensor.name:
+    name = f"{original_tensor.name}_ancillary"
+
+  return model_editor.Tensor(
+      shape=(len(ancillary_data), ),
+      dtype=tflite.TensorType.UINT8,
+      data=ancillary_data,
+      name=name,
+  )
```
ddavis-2015, line 71: Because `data` is specified here when creating the ancillary tensor, this prevents sharing of ancillary tensors via buffers.  Please see the general comment for this PR for the scenario.

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,306 @@
...
+  This function updates the tensor in place to reflect its new role.
+
+  Args:
+    tensor: The tensor to rewrite.
+    encoded_data: The compressed/encoded data bytes.
+  """
+  tensor.shape = (len(encoded_data), )
+  tensor.dtype = tflite.TensorType.UINT8
+  tensor.quantization = None
+  tensor.buffer.data = encoded_data
```
ddavis-2015: Encoded tensor buffer sharing works here, across subgraphs, as the original `Buffer` object is retained.  Excellent.  👍

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,306 @@
...
+      subgraph.operators.insert(op_position[consumer], decode_op)
+
+      # Rewire only this consumer to use the decoded outputs
+      for info, decoded in zip(infos, decoded_tensors):
+        _rewire_consumers([consumer], info.tensor, decoded)
+
+    # Decode compressed tensors read from the subgraph's output list, all
+    # with one DECODE appended after the last operator (see docstring).
+    output_infos = [info for info in tensor_infos if info.is_output]
+    if output_infos:
```
ddavis-2015, line 330: Should be `if len(output_infos):`

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+    }
+
+    decode_insert.insert_decode_operators(model, compression_results)
+
+    decode_op = model.subgraphs[0].operators[0]
+    output = decode_op.outputs[0]
+
+    # Output matches original (pre-rewrite) tensor shape and dtype
+    self.assertEqual(output.shape, original_shape)
+    self.assertEqual(output.dtype, original_dtype)
```
ddavis-2015: could also check that output.data == None and output.buffer == None

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+
+    # FC's second input (weights) should now be DECODE's output
+    self.assertIs(fc_op.inputs[1], decode_op.outputs[0])
+    # Original weights tensor should NOT be in FC inputs
+    self.assertNotIn(weights_tensor, fc_op.inputs)
+
+  def test_shared_tensor_decode_per_consumer(self):
+    """Tensor used by multiple ops gets separate DECODE for each consumer."""
+    model = _build_shared_weights_model()
+    weights_tensor = model.subgraphs[0].tensor_by_name("shared_weights")
```
ddavis-2015: unused variable `weights_tensor`

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+                  inputs=[input1, weights],
+                  outputs=[output1],
+              ),
+              model_editor.Operator(
+                  opcode=tflite.BuiltinOperator.FULLY_CONNECTED,
+                  inputs=[input2, weights],
+                  outputs=[output2],
+              ),
+          ],
+      )
```
ddavis-2015, line 161: Doesn't the Subgraph inputs/outputs need to be specified? Otherwise the subgraph compiler will just assign empty vectors to the tflite.SubGraphT.

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+    decode_op2 = sg.operators[2]
+    fc_op2 = sg.operators[3]
+
+    # Each FC should use its own DECODE's output
+    self.assertIs(fc_op1.inputs[1], decode_op1.outputs[0])
+    self.assertIs(fc_op2.inputs[1], decode_op2.outputs[0])
+    # The two DECODEs should have different outputs
+    self.assertIsNot(decode_op1.outputs[0], decode_op2.outputs[0])
+    # The two DECODEs should share the same ancillary tensor
+    self.assertIs(decode_op1.inputs[1], decode_op2.inputs[1])
```
ddavis-2015, line 410: also  check the two DECODES share the same encoded tensor

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+    self.assertEqual(output_decode.opcode, tflite.BuiltinOperator.CUSTOM)
+
+    # Each reader gets its own decoded tensor
+    self.assertIs(fc_op.inputs[1], consumer_decode.outputs[0])
+    self.assertIs(sg.outputs[1], output_decode.outputs[0])
+    self.assertIsNot(consumer_decode.outputs[0], output_decode.outputs[0])
+
+    # Both DECODEs share the ancillary tensor
+    self.assertIs(consumer_decode.inputs[1], output_decode.inputs[1])
+
```
ddavis-2015, line 623: also check both DECODEs share the encoded tensor

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+        name="output",
+    )
+
+    model = model_editor.Model(subgraphs=[
+        model_editor.Subgraph(
+            tensors=[weights1, weights2],
+            operators=[
+                model_editor.Operator(
+                    opcode=tflite.BuiltinOperator.FULLY_CONNECTED,
+                    inputs=[input_t, weights1, weights2],
```
ddavis-2015: CONCATENATION would be a better example here (with output shape [6,4])

tensorflow/lite/micro/compression/decode_insert_test.py
```diff
@@ -0,0 +1,657 @@
...
+
+    decode_op = model.subgraphs[0].operators[0]
+    ancillary_tensor = decode_op.inputs[1]
+
+    # Ancillary tensor data should match what we provided
+    self.assertEqual(bytes(ancillary_tensor.array), ancillary_data)
+
+    # Verify DCM header
+    dcm_bytes = ancillary_tensor.array[:16]
+    self.assertEqual(dcm_bytes[0], 0)  # decode_type = LUT
```
ddavis-2015: use DecodeType.LUT

ddavis-2015 (changes requested):
Looks really good, almost there.  What is here will create a valid and usable model, but not exactly as it should be.

Maximal reduction of the model is not yet achieved due to possible duplication of the ancillary tensor `Buffer`.  Here is the scenario:  Two const-tensors, either in the same or in different subgraphs can contain the exact same data.  The TfLite Converter can (and does) create two tensors in the model, but using a single shared buffer.  For DECODE insertion, the existing code works just fine for encoded tensors, as they are not given a new buffer, just new buffer content.  But this means that the related ancillary tensors, can also share a single buffer.  However, the current insertion code always creates a new `Buffer` object when creating ancillary tensors.

The output tensors of DECODE must be exact copies of the original unencoded tensors, meaning when the output tensors are created they must use the original tensor's `TensorT` object.  Only the name of the output tensors can be modified (as is currently done by the insertion code), and the `Buffer` object set to `None` (buffer in the `TensorT` object is zero).

For the unit tests, maybe a comment could be added stating the models are just for unit test purposes and do not represent fully valid models.  Same for the `encoded_data` and `ancillary_data` of the `CompressionResult` objects.  Something saying these are just dummy payloads for unit testing and do not correspond to the model's tensor data.  Someone looking at these for the first time might mistake these for valid examples and be confused.

rkuester:
Everything is addressed in new commits, kept separate for easy review, relying on the squash when the PR merges. On the two structural points:

DECODE outputs are now full copies of the original tensors, renamed and with no buffer (8330ba39), built by a new Tensor.copy in model_editor (d0fd38b2).

Buffer sharing is preserved by deduplication rather than a cache. Insertion gives every rewritten encoded tensor and every ancillary tensor a fresh buffer, then model_editor.dedupe_buffers merges byte-identical buffers across the whole model and prune_buffers drops unreferenced ones (363a71e1, e1aa8e62, f62df145). That yields what a Buffer-keyed cache would, across subgraphs and for coinciding ancillary data, and covers a case a cache cannot: aliases whose compression results differ. Identical bytes can encode differently, because quantization structure shapes the LUT value table. There the sharing dissolves, rather than one alias corrupting another through a buffer rewritten in place. A compressed tensor sharing a buffer with an uncompressed tensor is skipped with a warning, because the uncompressed alias keeps the original data in the model, so compression cannot shrink it. 37505cd3 tests these cases; 726b26f8 exercises dedup and pruning end to end on a model read from a flatbuffer.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: The design doc update of 2026-07-13 settles this. DECODE may feed IF and WHILE, with the invoking operator's inputs as the consumers. A DECODE output may be a subgraph output, with the DECODE as the subgraph's final operator and the invoking operator's outputs as the consumers. The insertion code conforms to both rules. cbf22b5d adds runtime tests that invoke models with DECODE across subgraph boundaries.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in 16fedda1: consumer lookup moved to model_editor as a public API, Subgraph.consumers_of, with its own tests, and the insertion tests no longer reach into private helpers.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in c2b32ced: the test models now declare their subgraph inputs and outputs.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in c2b32ced: the test models now declare their subgraph inputs and outputs.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in c2b32ced: the test models now declare their subgraph inputs and outputs.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in af0fbc8b: a helper now builds the whole CompressionResult, parameterized by element count, index bitwidth, and value table, so each test passes payloads consistent with the tensor it compresses and the encoded data is sized to match. The module docstring states these are structural fixtures, not decodable data, per your general comment.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in 2ed3e6b3.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in 2ed3e6b3.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Addressed in e7703c56, using your wording for the lifetime rule and the latency-for-memory tradeoff.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Addressed in e7703c56: outputs are now full copies (8330ba39), and the docstring calls them that.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Addressed in e7703c56: the fragment is gone. The passage now says a calling operator copies subgraph outputs when the subgraph returns.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Addressed in f62df145 by deduplication rather than a cache: fresh buffers for every rewritten and ancillary tensor, then dedupe_buffers merges byte-identical buffers across the whole model and prune_buffers drops the orphans. Details in the general comment.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Addressed in 8330ba39: outputs are full copies of the original TensorT, renamed, buffer set to None, built by a new Tensor.copy (d0fd38b2). The test builds its expected value the same way and compares field by field, so every TensorT field participates without being enumerated.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Resolved by the deduplication pass (f62df145): ancillary tensors still start with fresh buffers, and dedupe_buffers merges byte-identical ones afterward, so aliases that compress identically share one ancillary buffer in the built model.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: I kept the implicit form. PEP 8 asks for `if seq:` over `if len(seq):`, and the surrounding code follows that.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in 8330ba39: the test asserts output.buffer and output.array are None, and the field-by-field comparison against a buffer-less copy pins the rest.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in 88cdede4: both tests now assert the DECODEs read the same encoded tensor object.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in 88cdede4: both tests now assert the DECODEs read the same encoded tensor object.

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in d633f59d: the test now uses CONCATENATION with output shape (6, 4).

tensorflow/lite/micro/compression/decode_insert_test.py
rkuester: Addressed in 9e8607f4.

tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -16,6 +16,7 @@
 Provides a clean API for creating, reading, and modifying TFLite models.
 """
 
+import copy
 from dataclasses import dataclass, field
 from typing import Optional, Union, List
```
ddavis-2015: "field" import unused?

tensorflow/lite/micro/compression/model_editor_test.py
```diff
@@ -827,6 +827,284 @@ def test_tensor_by_name_not_found_raises(self):
...
+    self.assertEqual(len(model.buffers), count_before)
+
+  def test_metadata_survives_pruning(self):
+    """Metadata, carried outside tensor buffers, is unaffected."""
+    model = self._read_model_with_two_constants(metadata={"m": b"payload"})
+
+    model_editor.prune_buffers(model)
+
+    roundtrip = model_editor.read(bytes(model.build()))
+    self.assertEqual(roundtrip.metadata["m"], b"payload")
```
ddavis-2015, line 1107: This isn't a proper test for dedup pruning with surviving metadata buffers.  This is really a test of whether "build()" puts metadata buffers into the model.  The model buffers need to change here, before rebuilding and rereading the model.  This test needs to show that after changing the number of buffers in the model, the metadata is still present.

veblush (dismissed):
Thanks for the PR! I'll get this merged once David also approved it.

tensorflow/lite/micro/compression/model_editor.py
ddavis-2015, line 758: operator inputs can contain the value "-1", meaning that an optional tensor is not present.  Fused LSTM is an example, where of the 15 or so tensors it uses for input, only a few are actually present and used.

ddavis-2015, line 911: operator inputs must preserve optional tensors (an index of -1 in the flatbuffer operator input vector)

ddavis-2015, line 945: operator inputs must preserve optional tensors (an index of -1 in the flatbuffer operator input vector)

tensorflow/lite/micro/compression/model_editor.py
```diff
@@ -521,6 +608,79 @@ def build(self) -> bytearray:
...
+  seen = set()
+  for sg in model.subgraphs:
+    sources = [sg.tensors, sg.inputs, sg.outputs]
+    sources.extend(op.inputs for op in sg.operators)
+    sources.extend(op.outputs for op in sg.operators)
+    for source in sources:
+      for tensor in source:
+        if id(tensor) not in seen:
+          seen.add(id(tensor))
+          yield tensor
```
ddavis-2015, line 636: operator inputs must preserve optional tensors (an index of -1 in the flatbuffer operator input vector)

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,345 @@
...
+def _rewire_consumers(
+    consumers: list[model_editor.Operator],
+    old_tensor: model_editor.Tensor,
+    new_tensor: model_editor.Tensor,
+) -> None:
+  """Replace old_tensor with new_tensor in all consumer inputs."""
+  for consumer in consumers:
+    consumer.inputs = [
+        new_tensor if t is old_tensor else t for t in consumer.inputs
+    ]
```
ddavis-2015, line 106: operator inputs must preserve optional tensors (an index of -1 in the flatbuffer operator input vector)

tensorflow/lite/micro/compression/decode_insert.py
```diff
@@ -0,0 +1,345 @@
...
+
+def insert_decode_operators(
+    model: model_editor.Model,
+    compression_results: dict[tuple[int, int], compressor.CompressionResult],
+) -> None:
+  """Insert DECODE operators for all compressed tensors.
+
+  This function modifies the model in-place, inserting a DECODE operator
+  before any operator that uses a compressed tensor as input, and appending
+  a DECODE for compressed tensors listed as subgraph outputs.
```
ddavis-2015: "and appending *to the end of the subgraph* a DECODE for compressed tensors listed as subgraph outputs."

ddavis-2015 (changes requested):
A few remaining issues, but getting real close.

operator inputs can contain the value "-1", meaning that an optional tensor is not present. Fused LSTM is an example, where of the 15 or so tensors it uses for input, only a few are actually present and used.

examples:
```
  Op#0 UNIDIRECTIONAL_SEQUENCE_LSTM(T#0, T#16, T#15, T#14, T#13, T#12, T#11, T#10, T#9, T#-1, T#-1, T#-1, T#6[0, 0, 0, 0, 0, ...], T#5[236545, 236545, 236545, 236545, 236545, ...], T#7[0, 0, 0, 0, 0, ...], T#8[0, 0, 0, 0, 0, ...], T#-1, T#-1, T#17, T#18, T#-1, T#-1, T#-1, T#-1) -> [T#24]

  Op#2 FULLY_CONNECTED(T#25, T#4, T#-1) -> [T#26]
```

Just FYI:
The flatbuffer schema has an "intermediates" vector as part of the operator table.  This is for "hidden" tensors that are not in the operator input or output vector, yet the tensors are used by the operator.  Currently there are no TFLM kernels that use intermediate tensors (the only kernel I know of that used them was the old version of LSTM, which no longer exists in the repo).

rkuester:
> operator inputs can contain the value "-1", meaning that an optional tensor is not present.

Fixed in dfe71853, with the tests added first as expected failures in ce20db84.

tensorflow/lite/micro/compression/model_editor.py
rkuester: Fixed in 667f6548.

tensorflow/lite/micro/compression/model_editor.py
rkuester: Fixed in dfe71853, which also rejects any negative index other than -1. 8a285cb1 extends that to an operator's outputs and a subgraph's inputs and outputs.

tensorflow/lite/micro/compression/model_editor.py
rkuester: Fixed in dfe71853.

tensorflow/lite/micro/compression/model_editor.py
rkuester: Fixed in dfe71853.

rkuester: Fixed in dfe71853.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Yes, but no change needed here. (`test_rewiring_keeps_absent_input` tests this.)

tensorflow/lite/micro/compression/model_editor_test.py
rkuester: Fixed in 0ca2ec68.

tensorflow/lite/micro/compression/decode_insert.py
rkuester: Fixed in 967a2a9f.

ddavis-2015 (approved):
Onwards to @veblush !

veblush (approved):

PR #3633, opened 2026-07-17
WHILE operator input/output copy fix

### LINKED ISSUE #3632: WHILE operator input/output copy fix
@tensorflow/micro 

Remove extraneous tensor copy operation after first invocation of condition subgraph.

Move copy of operator inputs to outputs, such that it occurs before the first invocation of the condition subgraph.  This preserves the operator inputs when one or more of them is the output of DECODE, and alternate decompression memory is in use.  This is because the output of DECODE is for immediate consumption by the next operator in the graph (WHILE), yet it is possible for the WHILE subgraph invocations to share memory with the original DECODE output.

Update the unit test for multiple invocations of the condition and body subgraphs.

When copying tensors between operator inputs/outputs and subgraph inputs/outputs, check if the source and destination tensors share memory.

ddavis-2015:
@tensorflow/micro

Remove extraneous tensor copy operation after first invocation of condition subgraph.

Move copy of operator inputs to outputs, such that it occurs before the first invocation of the condition subgraph. This preserves the operator inputs when one or more of them is the output of DECODE, and alternate decompression memory is in use. This is because the output of DECODE is for immediate consumption by the next operator in the graph (WHILE), yet it is possible for the WHILE subgraph invocations to share memory with the original DECODE output.

Update the unit test for multiple invocations of the condition and body subgraphs.

When copying tensors between operator inputs/outputs and subgraph inputs/outputs, check if the source and destination tensors share memory.

bug=fixes #3632

### DIFF
--- a/tensorflow/lite/micro/kernels/kernel_util.cc
+++ b/tensorflow/lite/micro/kernels/kernel_util.cc
@@ -166,7 +166,10 @@ TfLiteStatus CopyOpInputsToOpOutputs(TfLiteContext* context, TfLiteNode* node) {
     TfLiteEvalTensor* output = tflite::micro::GetEvalOutput(context, node, i);
     int bytes = ValidateAndGetTensorSizes(input, output);
     TF_LITE_ENSURE(context, bytes >= 0);
-    memcpy(output->data.raw, input->data.raw, bytes);
+    // Don't copy if the tensors share memory
+    if (output->data.data != input->data.data) {
+      memcpy(output->data.data, input->data.data, bytes);
+    }
   }
   return kTfLiteOk;
 }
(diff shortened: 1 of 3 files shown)


rkuester:
I verified this fix against a runtime test suite I've been building for DECODE insertion (Python: model_editor plus the LUT compressor, DECODE insertion, and the TFLM interpreter bindings). The suite includes a reproducer for exactly the failure this PR describes: a DECODE output feeding a WHILE input, with a second DECODE in the cond subgraph and alternate decompression memory in use. Without this fix, WHILE re-reads its inputs after invoking the cond subgraph and picks up whatever the cond subgraph's DECODE wrote over the shared alternate memory, and the test fails.

With this PR cherry-picked onto my decode-insert branch, the reproducer passes, along with the rest of the suite and the full kernel test suite, so I see no regressions from the kernel_util changes. The change looks right to me.

The runtime suite is part of #3624, which is awaiting this fix.

veblush (approved):

PR #3533, opened 2026-04-17
Add overflow detection to tensor size calculations

TristanInSec:
Use size_t instead of int for element count accumulation in BytesRequiredForTensor and TfLiteEvalTensorByteLength. Validate each dimension is positive and check for multiplication overflow before computing the final byte count.

This brings the size calculation in line with safe arithmetic practices used elsewhere in the codebase.

BUG=b/general

PR #3637, opened 2026-07-26
Fix integer overflow in BytesRequiredForTensor and TfLiteEvalTensorByteLength

TristanInSec:
## Summary

`BytesRequiredForTensor()` and `TfLiteEvalTensorByteLength()` compute tensor byte size using `int element_count`, which overflows when model-supplied dimensions are large (e.g., 65536 x 65536 wraps int32 to 0). The undersized allocation leads to heap buffer overflow when kernels later write tensor data using the actual dimensions.

## Fix

- Change `element_count` from `int` to `size_t`
- Add overflow detection via division check after each multiplication
- Return `kTfLiteError` on overflow or non-positive dimensions
- Both functions run during Setup (model loading), consistent with the Error Handling Guide's recommendation to validate model-provided parameters early

This replaces the previous PR #3533 which was closed for not following the Error Handling Guide.

## Alignment with Error Handling Guide

Per Section 2 (Phase 1: Setup & Initialization):
> "Model-provided parameters: Check the number of inputs/outputs, tensor types, and tensor shapes."

These functions are called during `InitializeTfLiteTensorFromFlatbuffer` (Setup phase). The fix uses `kTfLiteError` return which callers propagate via `TF_LITE_ENSURE_STATUS`, keeping ROM cost low (no additional string literals).

## Testing

Verified with ASan-enabled build: crafted model with shape [65536, 65536] now returns `kTfLiteError` instead of allocating 0 bytes and overflowing.

Fixes GHSA-7rx3-jmhh-gq5m

BUG=n/a

### DIFF
--- a/tensorflow/lite/micro/memory_helpers.cc
+++ b/tensorflow/lite/micro/memory_helpers.cc
@@ -106,12 +106,18 @@ TfLiteStatus TfLiteTypeSizeOf(TfLiteType type, size_t* size) {
 
 TfLiteStatus BytesRequiredForTensor(const tflite::Tensor& flatbuffer_tensor,
                                     size_t* bytes, size_t* type_size) {
-  int element_count = 1;
-  // If flatbuffer_tensor.shape == nullptr, then flatbuffer_tensor is a scalar
-  // so has 1 element.
+  size_t element_count = 1;
   if (flatbuffer_tensor.shape() != nullptr) {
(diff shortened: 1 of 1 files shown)


ddavis-2015:
This type change will not be implemented prior to the upstream LiteRT adopting such changes in the common/shared code.

PR #3643, opened 2026-07-30
Fix build for cores without XT_NSAU, update in micro_allocations_info.cc

unmeshna017:
- The signal square-root code assumes the Xtensa `nsau` instruction is always available, which breaks the build on HiFi4 cores where the `XT_NSAU` intrinsic is not defined.
- Used GetBuiltinCode() wrapper function instead of opcode->builtin_code() function in micro_allocations_info.cc to ensure opcode read compatibility between the TFLite v3 and v3a schema versions.

Adding @cad-audio and @joshih-cad for review


ddavis-2015:
This PR was replaced by PR #3697 and #3698 for the following reasons:

* The PR did not meet Google code style requirements
* The PR was not focused on a single issue
* There were extraneous files within the PR that should have been marked for removal

PR #3648, opened 2026-08-03
Fix OOB buffer index in GetFlatbufferTensorBuffer

cloudymonstera:
## Summary
- Add a bounds check on attacker-controlled `tensor.buffer()` before indexing the model `buffers` vector in `GetFlatbufferTensorBuffer`.
- Mirrors the existing guard used by the compression-metadata path in the same file (`buffer_index >= buffers->size()`).
- Prevents out-of-bounds reads via `flatbuffers::Vector::operator[]` during `MicroInterpreter::AllocateTensors()` when loading a crafted `.tflite`.

## Test plan
- [ ] Existing TFLM unit / integration tests still pass
- [ ] Crafted model with `buffers.size()==2` and `tensor.buffer()==65535` no longer reaches an OOB index (returns null buffer / fails allocation cleanly instead of reading past the vector)
- [ ] Valid models with in-range `tensor.buffer()` indices still allocate and invoke successfully

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tensorflow/lite/micro/micro_allocator.cc
+++ b/tensorflow/lite/micro/micro_allocator.cc
@@ -203,7 +203,17 @@ void* GetFlatbufferTensorBuffer(
   // First see if there's any buffer information in the serialized tensor.
   // TODO(b/170379532): Add better unit tests to validate flatbuffer values.
   void* out_buffer = nullptr;
-  if (auto* buffer = (*buffers)[flatbuffer_tensor.buffer()]) {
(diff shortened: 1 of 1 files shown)


ddavis-2015:
Please see the [error handling guide](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/error_handling_guide.md) for the appropriate type of validation check PRs to submit.

PR #3649, opened 2026-08-03
Validate OfflineMemoryAllocation metadata buffer size

cloudymonstera:
## Summary
- Validate that the `OfflineMemoryAllocation` metadata `Buffer::data` is large enough for the 3×`uint32` header before reading `metadata_buffer[2]`.
- Validate that the buffer also contains `N`×`int32` offline offsets before returning `&metadata_buffer[3]` to the planner.
- Reject null buffer / empty buffer data early with `kTfLiteError`.

Without these checks, a truncated metadata buffer can OOB-read past `Buffer::data` and supply poisoned offline offsets to `CommitPlan` (`arena_start + offset`).

## Test plan
- [ ] Existing TFLM unit / integration tests still pass (esp. offline-planner cases in `micro_allocator_test.cc`)
- [ ] Undersized OfflineMemoryAllocation buffer (header &lt; 12 bytes, or header OK but offset payload truncated) fails allocation cleanly instead of reading past the buffer
- [ ] Valid OfflineMemoryAllocation models with a full header + offset table still allocate and invoke successfully

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tensorflow/lite/micro/micro_allocation_info.cc
+++ b/tensorflow/lite/micro/micro_allocation_info.cc
@@ -351,13 +351,25 @@ TfLiteStatus AllocationInfoBuilder::GetOfflinePlannedOffsets(
           const flatbuffers::Vector<flatbuffers::Offset<Buffer>>* buffers =
               model_->buffers();
           auto* buffer = (*buffers)[metadata->buffer()];
+          if (buffer == nullptr) {
+            MicroPrintf("OfflineMemoryAllocation: null buffer");
+            return kTfLiteError;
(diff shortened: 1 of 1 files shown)


ddavis-2015:
There is currently no offline memory planning tool.  Thus, OfflineMemoryAllocation metadata is not being used.

Please see the [error handling guide](https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/error_handling_guide.md) for the appropriate type of validation check PRs to submit.

Issue #3258, opened 2025-11-28
Tflite micro python package uploaded to PyPI should be built with O2 optimizations


pmjandreen:
Hi!

First of all I want to say thanks for the excellent tflite micro python package. The only issue I have encountered this far is its inference speed, very impractical to use for evaluations involving thousands of images. I tried building the package myself, editing tflite-**micro/python/tflite_micro/pypi_build.sh**, simply adding **--copt=-O2** to the bazel build call (although I had to suppress a warning also to make it work). With the resulting package my particular model inference was over 4 times faster. Do you think building with optimizations could be made standard for the python packages uploaded to  PyPI?

Best Regards,
Mikael

ddavis-2015:
@rkuester Any ideas on this?

rkuester:
I like it. Thanks, @pmjandreen! I'll work on getting this idea merged. See #3655.

rkuester:
Done. The wheels on PyPI are now built in opt mode (#3657), starting with version 0.dev20260813222316. Nightly publishing had also been stalled since February; that's repaired too, so `pip install --upgrade tflite-micro` gets a current, optimized build.

pmjandreen:
Great to hear! Thanks!

PR #3655, opened 2026-08-07
fix(pypi): repair and optimize the wheel build

rkuester:
This branch repairs the nightly PyPI wheel build, which broke when the bzlmod migration deleted the WORKSPACE file. It builds the wheel in bazel's opt compilation mode instead of the default fastbuild, which leaves C++ optimization off. It pins the checkout action to a commit hash, which the zizmor scan requires. It files a GitHub issue when the nightly build fails; the nightly build failed daily for six months before anyone noticed.

This is a draft PR for running CI, review, and seeing the commits in context. The commits along this branch will be individually submitted for merge.

BUG=addresses #3258

### DIFF
--- a/.github/workflows/pypi_build.yml
+++ b/.github/workflows/pypi_build.yml
@@ -61,3 +61,18 @@ jobs:
         run: |
           python/tflite_micro/pypi_upload.sh \
           bazel-pypi-out/tflite_micro-*.whl
+
+  issue-on-error:
+    needs: [pypi-build]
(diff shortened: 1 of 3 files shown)


rkuester:
All three parts merged: #3653 repaired the wheel build under bzlmod, #3657 built the wheel in opt mode, and #3669 set up the nightly workflow to file an issue when a build fails. Nightly publishing resumed on 2026-08-12 after a six-month gap, and the latest wheels on PyPI are built in opt mode.

PR #3659, opened 2026-08-12
test(compression): add integration tests with TFLM interpreter

rkuester:
Add tests that compress models with LUT compression, run them through
the TFLM Python interpreter, and verify outputs match uncompressed
originals. Cover per-tensor and per-channel quantization, various index
bitwidths, unquantized weights, and alternate decompression memory.

BUG=part of #3256

### DIFF
--- a/tensorflow/lite/micro/compression/BUILD
+++ b/tensorflow/lite/micro/compression/BUILD
@@ -169,6 +169,31 @@ py_test(
     ],
 )
 
+tflm_py_test(
+    name = "compression_integration_test",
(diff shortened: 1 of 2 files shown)


veblush (approved):

rkuester:
@veblush the merge queue run seems to have failed on runner infrastructure. *Makefile x86 Default* died in checkout when the Kubernetes API returned a 500, and *Makefile x86 with Compression* never started because its runner pod stalled in PodInitializing.

PR #3660, opened 2026-08-13
Use SafeCast for float-to-integer conversions in CAST kernel

veblush:
To align TFLM's CAST kernel with LiteRT, which already uses SafeCast for floating-point to integral conversions.

BUG=n/a

### DIFF
--- a/tensorflow/lite/micro/kernels/cast_test.cc
+++ b/tensorflow/lite/micro/kernels/cast_test.cc
@@ -13,6 +13,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
+#include <initializer_list>
+#include <limits>
+#include <memory>
+
 #include "tensorflow/lite/c/builtin_op_data.h"
 #include "tensorflow/lite/c/common.h"
 #include "tensorflow/lite/micro/kernels/kernel_runner.h"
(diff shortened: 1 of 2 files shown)


tensorflow/lite/micro/kernels/cast_test.cc
```diff
@@ -23,18 +27,20 @@ namespace tflite {
...
-  TfLiteIntArray* output_dims = IntArrayFromInts(input_dims_data);
-  const int output_dims_count = ElementCount(*output_dims);
-  constexpr int inputs_size = 1;
-  constexpr int outputs_size = 1;
-  constexpr int tensors_size = inputs_size + outputs_size;
+template <typename InputT, typename OutputT>
+void TestCast(std::initializer_list<InputT> input,
+              std::initializer_list<OutputT> golden) {
+  auto input_data = std::make_unique<InputT[]>(input.size());
+  auto output_data = std::make_unique<OutputT[]>(golden.size());
```
ddavis-2015: heap allocation is verboten

ddavis-2015 (changes requested):

tensorflow/lite/micro/kernels/cast_test.cc
veblush: Good catch. Resolved

ddavis-2015 (approved):
looks good to me

suleshahid (approved):

PR #3680, opened 2026-08-25
Validate RESIZE_NEAREST_NEIGHBOR size tensor against output shape

vee1e:
BUG=n/a

ResizeNearestNeighborPrepare() validates the size tensor's rank, type and constness but never compares its values against the declared output dimensions. Eval() then writes size[0]*size[1]*depth bytes sequentially into the output buffer, which the allocator sized from the declared shape, so a mismatch produces a linear out-of-bounds write (attacker-controlled length and content).

Reject such models at Prepare time and add a regression test that asserts a size/output mismatch fails allocation.

Verified locally: with this change, a model declaring output [1,1,1,1] with a size tensor {1024,1024} is rejected at AllocateTensors instead of overflowing the arena.


vee1e:
@veblush Could I have a review on this when you're free? It's a small fix that validates the size tensor.

PR #3696, opened 2026-09-08
Validate DECODE LUT ancillary, input, and stride bounds in Setup

roopd3v:
BUG=none

`DecodeStateLut::Setup` used `NumElements(output)` as the decompress loop bound without checking `input.bytes`, took the value table as `ancillary + 16` without checking `ancillary.bytes`, and stored the DCM stride byte (0–255) without enforcing `kMaxValueTableChannelStride` (128). A structurally valid FlatBuffer that only mismatches those sizes therefore OOBs in `Eval`.

This change rejects those shapes in `Prepare`/`Setup` with `TF_LITE_ENSURE` (TFLM `docs/error_handling_guide.md`: semantic malformation belongs there). Optimized Width2/3/4 unpackers run only when `stride >= 2^bit_width`; otherwise `WidthAny` with `index < stride`. Official LUT tests use `2^n-1` tables, so this does not require `stride >= 2^bit_width`.

New test `decode_lut_bounds_test` fails before the fix (`Prepare` succeeds) and passes after (`kTfLiteError`). Existing `decode_test` and `decode_state_lut_test` still pass.

```text
make -f tensorflow/lite/micro/tools/make/Makefile test_decode_test
make -f tensorflow/lite/micro/tools/make/Makefile test_decode_state_lut_test
make -f tensorflow/lite/micro/tools/make/Makefile test_decode_lut_bounds_test
```

### DIFF
--- a/tensorflow/lite/micro/kernels/BUILD
+++ b/tensorflow/lite/micro/kernels/BUILD
@@ -720,6 +720,22 @@ tflm_cc_test(
 )
 
 tflm_cc_test(
+    name = "decode_lut_bounds_test",
+    srcs = [
+        "decode_lut_bounds_test.cc",
+    ],
+    deps = [
+        ":decode_test_helpers",
+        ":kernel_runner",
+        "//tensorflow/lite/c:common",
+        "//tensorflow/lite/micro:debug_log",
+        "//tensorflow/lite/micro:op_resolvers",
+        "//tensorflow/lite/micro:test_helpers",
+        "//tensorflow/lite/micro/testing:micro_test",
+    ],
+)
+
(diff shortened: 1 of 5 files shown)


roopd3v:
@veblush could you clarify why this was closed?

This fixes an OOB read in DECODE LUT caused by missing input/ancillary/stride bounds in Setup. I also included regression tests and the existing DECODE tests still pass.

Google OSS VRP has the corresponding memory-corruption report assigned and is waiting for this patch to be merged before they can reassess it under the OT1 rules.

If there is anything wrong with the patch or tests, let me know and I'll fix it. I just need to understand why it was closed without review.

PR #3697, opened 2026-09-08
Fix for cores without XT_NSAU

ddavis-2015:
@tensorflow/micro

The signal square-root code assumes the Xtensa nsau instruction is always available, which breaks the build on HiFi4 cores where the XT_NSAU intrinsic is not defined.

Remove xtensa optimized filter_bank_square_root, as it is slower on HIFI4 and HIFI5 than reference C++ implementation.

This PR is a reimplementation of PR #3643

bug=cadence update


rameshkunasi:
@ddavis-2015 I think the reason for removing **filter_bank_square_root** xtensa optimized kernel is it uses the NSAU intrinsic directly, which will fails on HiFi4 cores. Please correct me if my understanding is wrong.

ddavis-2015:
> @ddavis-2015 I think the reason for removing **filter_bank_square_root** xtensa optimized kernel is it uses the NSAU intrinsic directly, which will fails on HiFi4 cores. Please correct me if my understanding is wrong.

This is also true.  However, in the original PR they had moved this kernel to a `.bak` file, which I found odd (and which we cannot put into the repo).  This caused me to then benchmark the operation using the simulator.  Note, our HIFI4 core has NSAU support.

You can see the code I used to test in https://github.com/tensorflow/tflite-micro/pull/3697/commits/ceb2e7ef2de4ce6c2c62465ca032020aa964938d

rameshkunasi:
Thanks for the clarification. This looks good to me.

veblush (approved):

PR #3698, opened 2026-09-08
Fix CALL_ONCE subgraph tensor lifetimes

ddavis-2015:
@tensorflow/micro

The MarkSubgraphLifetimesIfNecessary method used during arena memory planning, was not handling CALL_ONCE subgraphs.  This could result in all tensors of that subgraph to not be assigned any memory by the memory planner.

Currently, CALL_ONCE is primarily used by Keras for resource tensor initialization.  Because resource tensors do not have their memory assigned through the memory planner, no issue resulted.

This PR is a reimplementation of PR #3643

bug=cadence updates


rameshkunasi (commented):
Looks Good to Me

veblush (approved):

PR #3736, opened 2026-09-17
Remove `tensorflow` and `absl-py` from `tflite-micro` wheel:

veblush:
BUG=n/a


rkuester:
This appears to be mostly obsoleted by #3760, but otherwise looks good to me.

PR #3738, opened 2026-09-18
Restore C++11 compatibility in common.h WrappingAdd/Mul and half.h

zhaoxu-google:
### Problem
Recent commits introduced C++17 language syntax (`if constexpr`, `std::is_integral_v`, `std::common_type_t`, and multi-statement `constexpr` functions) into `tensorflow/lite/kernels/internal/common.h` and `tensorflow/lite/types/half.h`.

This causes build failures on embedded platforms and vendor DSP toolchains that rely on C++11 (such as GCC 6.4, Cadence Xtensa Dinkumware C++11 STL, and targets using `CC_VER11=true`).

### Solution
1. **`tensorflow/lite/kernels/internal/common.h`**:
   - Replaces `if constexpr` with C++11-compatible template specialization (`detail::WrappingArithmetic<T>`).
   - Uses `std::is_integral<T>::value` and `typename std::common_type<...>::type`.
2. **`tensorflow/lite/types/half.h`**:
   - Replaces the multi-statement `constexpr half::from_bits` function with a C++11 single-expression constructor delegation using a private `raw_bits_initializer` tag.

### Compatibility
Maintains 100% full compatibility across C++11, C++14, C++17, and C++20 with zero runtime or binary overhead (template dispatch and constructors resolve entirely at compile time).


veblush:
Thanks for the PR! We cannot accept this change because TFLM has moved to mandating C++17 and no longer maintains C++11 compatibility. Additionally, the modified files (common.h, half.h) are synced from upstream LiteRT and cannot be patched directly in this repository. Closing this PR.

PR #3758, opened 2026-09-19
feat(compression): add LUT mode to spec and DCM

rkuester:
Let a compression spec state whether a tensor gets one value
table or one table per channel along a given axis, instead of
that choice always coming from the tensor's quantization. Honor
the choice in the LUT compressor, let the spec builder state it,
parse it from a spec file when given, and add the mode to the DCM
as a new axis field in the decode header.

Name the mode in a spec file with a per_channel mapping that gives
the axis, or with a bare per_tensor key:

    - lut:
        index_bitwidth: 4
        per_channel:
          axis: 0

    - lut:
        index_bitwidth: 2
        per_tensor:

For an entry that names neither, keep the old behavior and take the
mode from the tensor's quantization.

BUG=part of #3256


tensorflow/lite/micro/compression/spec.py
veblush: Is YAML syntax support for `mode` intentionally deferred to a follow-up PR, or should it be supported here?

rkuester: It was in the next commit, but you're right that there's no reason to defer it. I've pulled it into this PR.

veblush (approved):

PR #3759, opened 2026-09-21
build(compression): clean up test dependencies and tags

veblush:
- Split `view` binary into `view_lib` library and have tests depend on libraries rather than binaries.
- Add missing `micro_log` and `span` deps to `metadata_test_cc`.
- Add `notap` tags to `metadata_saved_test` and `proprietary_integration_smoke_test`.

BUG=n/a


veblush:
cc: @rkuester 

This is mainly for resolving the internal build contraints.

suleshahid (approved):

PR #3733, opened 2026-09-17
Support optional valid output tensor in CIRCULAR_BUFFER custom op

veblush:
Added support for an optional 2nd boolean output tensor in `CIRCULAR_BUFFER` to signal output validity without aborting the graph (`kTfLiteAbort`).

BUG=n/a

PR #3762, opened 2026-09-22
Benchmarking: wait for scalar bool output_valid on streaming models

rameshkunasi:
Streaming models that gate feature extraction and classification subgraphs using Framer/Stacker and IF operators return `kTfLiteOk` with a 0D scalar boolean output_valid tensor set to **false** while internal circular buffers are warming up.

Update `generic_model_benchmark` to check for any 0D boolean output tensor and continue invoking until output_valid becomes true before logging steady-state profiler metrics and exiting the benchmark loop.

BUG=n/a


tensorflow/lite/micro/tools/benchmarking/generic_model_benchmark.cc
ddavis-2015, line 1: @rameshkunasi These changes are not sufficient to support all models reliably.  I suggest adding makefile changes similar to how `GENERIC_BENCHMARK_TENSOR_ARENA_SIZE` is implemented.  Lets call the new makefile var `GENERIC_BENCHMARK_STREAMING_OUTPUT` and the value it is set to is the MicroInterpreter output tensor index (0, 1, 2, ...).  This allows specific builds of the benchmark tool for streaming models.

The `IsOutputValid` method should be adjusted to use the `GENERIC_BENCHMARK_STREAMING_OUTPUT` value, and have `#if defined(GENERIC_BENCHMARK_STREAMING_OUTPUT)` around that method and any use of the method.

ddavis-2015, line 1: @rameshkunasi Could you also make this change (not part of your original PR):
```
    if ((status != kTfLiteOk) && (static_cast<int>(status) != kTfLiteAbort)) {
      MicroPrintf("Model interpreter invocation failed: %d\n", status);
      return -1;
    }
```
Changed to:
```
    if (status != kTfLiteOk) {
      if (static_cast<int>(status) == kTfLiteAbort) {
        continue;
      }
      else {
        MicroPrintf("Model interpreter invocation failed: %d\n", status);
        return -1;
      }
    }
```

ddavis-2015 (changes requested):

ddavis-2015:
This PR is dependent on PR #3733

